PR #3, opened 2019-03-14
Add a `toJson` method for serialization.

### LINKED ISSUE #2: Add a `toJson` method for generated Table class.
Is it possible to add a `toJson` method to the generated table class? Because I sometimes need to serialize the table class into json, thank you very much!

yohom:
Fixes #2.

### DIFF
--- a/moor_generator/lib/src/writer/data_class_writer.dart
+++ b/moor_generator/lib/src/writer/data_class_writer.dart
@@ -1,5 +1,5 @@
-import 'package:recase/recase.dart';
 import 'package:moor_generator/src/model/specified_table.dart';
+import 'package:recase/recase.dart';
 
 class DataClassWriter {
   final SpecifiedTable table;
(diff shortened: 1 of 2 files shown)


moor_generator/lib/src/writer/data_class_writer.dart
```diff
@@ -94,6 +97,17 @@ class DataClassWriter {
     buffer.write(');}\n');
   }
 
+  void _writeToJson(StringBuffer buffer) {
+    buffer.write('Map<String, Object> toJson() {\n return {');
```
simolus3: Might be a personal thing, but I'd prefer this to return `Map<String, dynamic>`, it feels more appropriate as that's also what would be returned by `json.decode`.
Unless you have a better argument in favor of `Object`, of course :)

simolus3:
Thanks again for your contribution :+1:

PR #17, opened 2019-04-23
`insertAll` method should use `orReplace` argument rather than `_orReplace` flag.

yohom:
I removed the `_orReplace` flag, and unified use `orReplace` parameters.
BTW, I think `insertOrReplace` method could be marked as `Deprecated`, because I add `orReplace` parameter in `insert` method too. You can do it later.


simolus3:
Thanks!

PR #177, opened 2019-10-07
Fix version solving fail for extras/encryption

michaelspiss:
Version solving fails with the newest release of moor (2.0.0). This only updates the moor version in pubspec.yaml as the mirrored moor_flutter.dart has not changed except for the removal of functionality that was not supported by encrypted_moor.dart.


simolus3:
Thanks!

PR #181, opened 2019-10-10
Make transaction return a value

minia68:


simolus3:
Thank you for your contribution. Looks good!

PR #238, opened 2019-11-15
Fix typo: now to know

Aarblon:
While reading through the documentation, I noticed this typo where you meant to use the word `know` instead of `now`.


simolus3:
Right - thanks!

PR #264, opened 2019-12-02
add more ignore_for_file rules to the moor_generator

britannio:
The current ignore_for file rules do not counteract the linter rules used in the Flutter SDK [link](https://github.com/flutter/flutter/blob/master/analysis_options.yaml) and subsequently the linter rules in my project. The following change addresses this issue by adding additional linter rules to the moor generator. 

In the future, it may be worth addressing these linter warnings directly by adapting the code generator to generate code that adheres to these new linter rules if possible.

Thanks.

### DIFF
--- a/moor/test/data/tables/custom_tables.g.dart
+++ b/moor/test/data/tables/custom_tables.g.dart
@@ -966,6 +966,13 @@ abstract class _$CustomTablesDb extends GeneratedDatabase {
         readsFrom: {config}).map(_rowToReadRowIdResult);
   }
 
+  Selectable<int> test() {
+    return customSelectQuery(
+        'WITH RECURSIVE\n  cnt(x) AS (\n    SELECT 1\n      UNION ALL\n      SELECT x+1 FROM cnt\n      LIMIT 1000000\n    )\n  SELECT x FROM cnt',
+        variables: [],
+        readsFrom: {}).map((QueryRow row) => row.readInt('x'));
+  }
+
   Future<int> writeConfig(String key, String value) {
     return customInsert(
       'REPLACE INTO config VALUES (:key, :value)',
--- a/moor/test/data/tables/tables.moor
+++ b/moor/test/data/tables/tables.moor
@@ -32,3 +32,12 @@ readMultiple: SELECT * FROM config WHERE config_key IN ? ORDER BY $clause;
 readDynamic: SELECT * FROM config WHERE $predicate;
 
 readRowId: SELECT oid, * FROM config WHERE _rowid_ = $expr;
+
+cfeTest: WITH RECURSIVE
+  cnt(x) AS (
+    SELECT 1
+      UNION ALL
+      SELECT x+1 FROM cnt
+      LIMIT 1000000
+    )
+  SELECT x FROM cnt;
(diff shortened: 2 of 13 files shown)


simolus3:
Sorry - looks like I broke something when changing the base to `develop`. I'll apply the commit manually. Thanks for pointing out the conflicting lint rules.

> may be worth addressing these linter warnings directly by adapting the code generator to generate code that adheres to these new linter rules if possible.

Some of them, like the `lines_longer_than_80_chars` rule, require a fair bit of work to implement that doesn't bring much value IMHO. In general, lints exist for human-written code, and I don't really understand why making it adhere to a certain style is needed. Is ignoring `*.g.dart` files via `analysis_options.yaml` not an option for you?

britannio:
I had a look at https://dart.dev/guides/language/analysis-options#excluding-files but I couldn't get it to work. If I can then this will no longer be an issue for me.
I only added the `lines_longer_than_80_chars` rule to prevent the comment I introduced from giving me a linter warning and I agree with your point about lints existing for human-written code.

simolus3:
This one works for me:
```yaml
analyzer:
  exclude:
    - '**/*.g.dart'
```

You might have to restart the analysis server (the red arrow icon in the "Dart Analysis" tab in IntelliJ/Studio) after making that change.

Mike278:
I’ve never been able to get exclude to work either - it’s the reason I can’t set implicit-casts: false

I think I remember finding a github issue about it

britannio:
> I’ve never been able to get exclude to work either - it’s the reason I can’t set implicit-casts: false
> 
> I think I remember finding a github issue about it

When I run `flutter analyze` with the exclude rule, it correctly ignores the file but VSCode is still highlighting warnings and errors in the generated file. Sounds like an issue with the VSCode tools.

simolus3:
> it’s the reason I can’t set implicit-casts: false

Interesting, I have it set to false in the moor repo and on other projects using moor. I'm not getting any errors on them, even when I have the generated files open. Do you know which generated method is causing this? `implicit-casts` is a language feature that will be gone after NNBD, so this is something worth looking into. 

> Sounds like an issue with the VSCode tools

Sometimes I'm hitting similar problems in IntelliJ, so it's an issue with the underlying analysis server: http://dartbug.com/35475

Mike278:
> When I run flutter analyze with the exclude rule, it correctly ignores the file but VSCode is still highlighting warnings and errors in the generated file.

Same for me in IntelliJ. `flutter analyze` does not find any issues but I still see red underlines in the editor. The normal tricks of restarting analysis server and typing in the file don't work either. 
> Do you know which generated method is causing this?

The map literal returned by `DataClass.toJson()` and the `==(other)` operator both need type arguments.

simolus3:
> The map literal returned by `DataClass.toJson()` and the `==(other)` operator both need type arguments.

Just to clarify - are we still talking about `implicit-cast` here? The behavior you're describing sounds more like `implicit-dynamic: false` or `strict-inference: true`/`strict-raw-types: true`. Being static language options, we should probably support them as well. But to me they aren't as important as `implicit-casts`, which really shouldn't exist.

Mike278:
Sorry sorry sorry yes I meant `implicit-dynamic: false` - must have copypasted the wrong thing. `implicit-casts: false` is working fine.

PR #270, opened 2019-12-06
Fix table not found exception.

Elvis5566:
We'll get table not found exception if there is no data in db because db is not stored after creating tables.

Reproduce steps:
1. select any table and keep tables empty.
2. refresh page.


Elvis5566:
Why test fail?
I got 
+179: All tests passed!

simolus3:
Yeah, the tests broke because of multiple reasons, but its not your fault. The Dart team uploaded a broken `analyzer` package, I introduced a typo and forgot to change some tests after an internal refactoring in the generator. Took some time to figure out, but it looks like the CI is working again.

Thank you for the fix! :+1:

Issue #287, opened 2019-12-20
Option for fully qualified for columns


North101:
I have a complex query that has a predicate

```
listCampaignItemBy:
WITH
    list_campaign AS (
        SELECT campaign.id
        FROM campaign
        LEFT JOIN party
            ON party.campaign_id = campaign.id
        LEFT JOIN lineage
            ON lineage.party_id = party.id
        LEFT JOIN lineage_character
            ON lineage_character.lineage_id = lineage.id
        WHERE $predicate
    )

SELECT item.*
FROM item
WHERE item.item_group_id IN (
    SELECT prosperity.unlock_item_group_id
    FROM campaign
    INNER JOIN prosperity_tick
        ON prosperity_tick.id = campaign.prosperity_tick_id
    INNER JOIN prosperity
        ON prosperity.id <= prosperity_tick.prosperity_id
    WHERE campaign.id IN (SELECT id FROM list_campaign)
)

UNION

SELECT item.*
FROM campaign_item
INNER JOIN item
    ON item.id = campaign_item.item_id
WHERE campaign_item.campaign_id IN (SELECT id FROM list_campaign)

ORDER BY $order;
```
(offtopic: good job on having this actually work!)

When building the $predicate
```
db.listCampaignItemBy(db.campaign.id.equals(widget.id))
```

I get an error saying that `id` is ambiguous. I was able to work around it by using a custom expression
```
CustomExpression<String, StringType>('campaign.id').equals(widget.id)
```
but I being able to use the generated column names would be great e.g.
```
db.listCampaignItemBy(db.campaign.id.withTable.equals(widget.id))
```

simolus3:
This is a duplicate of #269. I appreciate your contribution, I'll do the necessary work in the generator to detect multiple tables being used.

PR #289, opened 2019-12-20
Generate $-variables should generate fully qualified columns 

North101:
Part of a fix for #287 

Last bit is to add `hasMultipleTables: true` to the generated query.
```
  Selectable<Item> <query>(
      Expression<bool, BoolType> predicate, OrderBy order) {
    final generatedpredicate = $write(predicate, hasMultipleTables: true);
    final generatedorder = $write(order, hasMultipleTables: true);
    return customSelectQuery(
       <sql>,
        variables: [
          ...generatedpredicate.introducedVariables,
          ...generatedorder.introducedVariables
        ],
        readsFrom: {
          <tables>
        }).map(_rowToItem);
  }
```

I wasn't able to figure out how to do that

### DIFF
--- a/moor/lib/src/runtime/api/query_engine.dart
+++ b/moor/lib/src/runtime/api/query_engine.dart
@@ -302,8 +302,11 @@ mixin QueryEngine on DatabaseConnectionUser {
 
   /// Will be used by generated code to resolve inline Dart expressions in sql.
   @protected
-  GenerationContext $write(Component component) {
+  GenerationContext $write(Component component, {bool hasMultipleTables: null}) {
     final context = GenerationContext.fromDb(this);
+    if (hasMultipleTables != null) {
+      context.hasMultipleTables = hasMultipleTables;
+    }
 
     // we don't want ORDER BY clauses to write the ORDER BY tokens because those
     // are already declared in sql


simolus3:
Thank you for the contribution. I do most moor work on the `develop` branch and only merge to `master` after a release to pub. In this case, the current state on `develop` already contains a similar fix - we just set `hasMultipleTables = true` unconditionally (4ee228d5a6d47ad333d24b41e80a94d322f58e6a).  Sorry for this annoyance, I'll add an explanation to `CONTRIBUTING.md`.

I like the idea of only enabling it if there are multiple tables, it's better than what I came up with. I'll keep this open and add some logic to detect those cases in the generator.

simolus3:
Merged this to `develop` in c7001f7889486312704cdd003849aae854297d89.

North101:
How do I use the develop branch (without using a local copy)?
I tried setting it in pubspec but it didn't work.
```
dependencies:
  moor_flutter:
    git:
      url: git@github.com:simolus3/moor.git
      ref: develop
      path: moor_flutter

dev_dependencies:
  moor_generator:
    git:
      url: git@github.com:simolus3/moor.git
      ref: develop
      path: moor_generator
```

pubspec.lock
```
  moor:
    dependency: transitive
    description:
      name: moor
      url: "https://pub.dartlang.org"
    source: hosted
    version: "2.1.1"
  moor_flutter:
    dependency: "direct main"
    description:
      path: moor_flutter
      ref: develop
      resolved-ref: "8987da453b21b0364ae40f36b4971d4449f9f812"
      url: "git@github.com:simolus3/moor.git"
    source: git
    version: "2.0.0"
  moor_generator:
    dependency: "direct dev"
    description:
      name: moor_generator
      url: "https://pub.dartlang.org"
    source: hosted
    version: "2.1.1"
```

I then tried putting `moor` manually in pubspec with the develop branch but then I get the error:
```
Because every version of moor_flutter from git depends on moor from hosted and gloomhaven_campaign depends on moor from git, moor_flutter from git is forbidden.
So, because gloomhaven_campaign depends on moor_flutter from git, version solving failed.

pub get failed (1; So, because gloomhaven_campaign depends on moor_flutter from git, version solving failed.)
```

simolus3:
I would recommend to use `dependency_overrides`:
```yaml
dependency_overrides:
  moor:
    git:
      url: git@github.com:simolus3/moor.git
      ref: develop
      path: moor
```
Since this doesn't apply transitive dependencies, you need to have an override entry for `moor`, `moor_generator` and `sqlparser`. Technically `moor_flutter` as well, but I think there aren't any changes on develop.

North101:
Thanks. It's pulling in the develop branch now. Unfortunately $predicate is still generating columns without tables/aliases

simolus3:
I'll take a more detailed look tomorrow, maybe it's because of the CTE or the compound select (although I assume that should still work).

Can you verify that you're really using the latest version from `develop` by looking at the `moor_generator` entry in the `.packages` file? Maybe you also need to run `flutter packages pub run build_runner clean` once so that the build snapshot is rebuilt.

North101:
Yep. I was on the develop branch. When I inspected the code it showed me the `if (hasMultipleTables != null)`

When I changed it back to an unconditional `context.hasMultipleTables = true` it worked again

simolus3:
Thanks for checking. I still can't reproduce this though. I tried to use a CTE on some tables from a unit test file
```sql
CREATE TABLE with_defaults (
    a TEXT DEFAULT 'something',
    b INT UNIQUE
);

CREATE TABLE with_constraints (
    a TEXT,
    b INT NOT NULL,
    c FLOAT(10, 2),

    FOREIGN KEY (a, b) REFERENCES with_defaults (a, b)
);

anotherCfe:
WITH
    data AS (
        SELECT d.a, c.c FROM with_constraints c
        LEFT JOIN with_defaults d ON d.a = c.a AND d.b = c.b
    )
SELECT * FROM config WHERE config_key IN (SELECT a FROM data) AND $predicate
UNION
SELECT 'foo' AS config_key, 'bar' AS config_value;
```

This generates a
```dart
final generatedpredicate = $write(predicate, hasMultipleTables: true);
```
for me... Can you try to find a query on multiple tables that doesn't set `hasMultipleTables` as expected? I can take a look at that then.

PR #298, opened 2019-12-24
[moor_ffi] Implement overrideForAll and overrideFor

negator:
Im not sure if I'm missing something, but these function do not appear to be implemented, eventhough the documentation references them for use.


simolus3:
Ouch, how could I miss that :man_facepalming: Thanks for the contribution :+1:!

simolus3:
Just released moor_ffi 0.3.1 which contains this fix.

PR #309, opened 2020-01-02
Members in the generated DataClass with an annotation

PoloLacoste:
As I was working with this project, I thought it would be interesting to be able to add members to the `DataClass` with a simple annotation `classVariable` (or any better name)

Here is an example below :

```dart
@DataClassName('TodoEntry')
class Todos extends Table {

  TextColumn get content => text()();

  DateTimeColumn get targetDate => dateTime().nullable()();

  IntColumn get category => integer()
      .nullable()
      .customConstraint('NULLABLE REFERENCES categories(id)')();

  @classVariable
  bool withDefaultValue = false;

  @classVariable
  String withoutDefaultValue;
}
```

### DIFF
--- a/moor/lib/src/dsl/table.dart
+++ b/moor/lib/src/dsl/table.dart
@@ -123,3 +123,13 @@ class DataClassName {
   /// {@macro moor:custom_data_class}
   const DataClassName(this.name);
 }
+
+/// A class to to be used as an annotation on members of the [Table] class
+class ClassVariable
+{
+	/// Constructor
+	const ClassVariable();
+}
+
+/// define the annotation
+const classVariable = ClassVariable();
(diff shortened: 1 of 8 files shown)


simolus3:
Thank you for the contribution :+1:  I only took a quick look, but the overall implementation seems reasonable. I will do a proper code review soon.

However, I'm not quite sure if I can find a good use-case for this. The data classes are designed to represent a row in the table. So what benefit do we get if we add another field that's not loaded or stored in the database? I assume you have a fitting scenario where this is useful, would you mind explaining the use-case a bit more? I don't want to reject the idea, I just need some more information.

Another way to add properties to data classes is using extensions and expandos. It arguably requires more boilerplate though:
```dart
final _withoutDefaultValue = Expando<String>();

extension on DataClass {
  String get withoutDefaultValue => _withoutDefaultValue[this];
  set withoutDefaultValue(String value) => _withoutDefaultValue[this] = value;
}
```

PoloLacoste:
Actually I didn't know about `Expando`, which resolve the problem of runtime class members.

The only reason I was implementing this was to solve the problem above.
I was manipulating data objects and wanted to change some property only for runtime.

For example :
- Is my `Todo` object selected ?
- Did I already cache some data from an API in it ?
- etc...

Separating stored data and runtime data combined with `Extension` is a lot cleaner I think.

PR #315, opened 2020-01-03
Missing `.path` in _openConnection

vitaminwater:


simolus3:
Right, thanks again :+1:

PR #321, opened 2020-01-05
Fix code example for `Getting started with sql` doc

DmitriyZaitsev:
Hey there! Thanks for an awesome library!

I was playing with it and found in the documentation a couple of mistakes that don't allow moor_generator to compile the code snippet copied from the doc and pasted to my project.
This PR fixes them.


simolus3:
Thank you for fixing the example :+1:

PR #374, opened 2020-02-02
Make the tables and daos list receive only its supported types (Table/DatabaseAccessor)

pedromassango:
I just made the list of tables receive classes that extends Table class and the daos list receive classes that extends DatabaseAccessor class.

### DIFF
--- a/moor/lib/src/dsl/database.dart
+++ b/moor/lib/src/dsl/database.dart
@@ -16,7 +16,7 @@ part of 'dsl.dart';
 /// ```
 class UseMoor {
   /// The tables to include in the database
-  final List<Type> tables;
+  final List<Table> tables;
 
   /// Optionally, the list of daos to use. A dao can also make queries like a
   /// regular database class, making is suitable to extract parts of your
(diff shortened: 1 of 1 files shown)


simolus3:
Thanks for the contribution!

I'm not sure if this works though. We'd have to change usages from
```dart
@UseMoor(tables: [Users])
```
to
```dart
@UseMoor(tables: [Users()])
```
and convert the type to an actual instance. This is pretty breaking, especially since `Table` doesn't have a const constructor. I've left a note on a related [Dart language issue](https://github.com/dart-lang/language/issues/459#issuecomment-581177058) asking for something like `List<Type<Table>>`.

pedromassango:
> Thanks for the contribution!
> 
> I'm not sure if this works though. We'd have to change usages from
> 
> ```dart
> @UseMoor(tables: [Users])
> ```
> 
> to
> 
> ```dart
> @UseMoor(tables: [Users()])
> ```
> 
> and convert the type to an actual instance. This is pretty breaking, especially since `Table` doesn't have a const constructor. I've left a note on a related [Dart language issue](https://github.com/dart-lang/language/issues/459#issuecomment-581177058) asking for something like `List<Type<Table>>`.

I took some time trying to fix it until I saw you comment here. Thank you.

neiljaywarner:
@pedromassango @simolus3  should this be closed so a surface level evaluation of long-outstanding PRs isn't an orange flag to new people evaluating this package? thanks

simolus3:
Good point - unfortunately the PR doesn't work as is and since `Type` isn't generic it can't be fixed either.

PR #425, opened 2020-03-02
Expose underlying db in encrypted moor

davidmartos96:
Same as #371 but for moor with SqlCipher


simolus3:
The failing CI appears to be something wrong with the build system that's fixed on `develop`.

Looks good, thanks for the contribution!

PR #426, opened 2020-03-03
FIX: onUpgrade callback was not being called when downgrading db in moor_flutter

davidmartos96:
PR splitted in 2 commits. The first one improves a bit the integrations tests with moor_flutter, I was having issues with those changes otherwise.

As the title says, the onUpgrade callback was not being called because sqflite distinguishes between onUpgrade and onDowngrade, so if we are trying to use a db version lower than the version installed, onUpgrade would not fire. That is sometimes undesired if we want to do some checks or custom logic with different db versions. For example trying to import an arbitrary db (with any of the versions supported by the app) into the app with a file picker.

### DIFF
--- a/extras/encryption/lib/encrypted_moor.dart
+++ b/extras/encryption/lib/encrypted_moor.dart
@@ -73,6 +73,9 @@ class _SqfliteDelegate extends DatabaseDelegate with _SqfliteExecutor {
       onUpgrade: (db, from, to) {
         _loadedSchemaVersion = from;
       },
+      onDowngrade: (db, from, to) {
+        _loadedSchemaVersion = from;
+      },
       singleInstance: singleInstance,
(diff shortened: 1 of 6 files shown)


davidmartos96:
@simolus3 Should I use develop as base to fix the CI issue?

simolus3:
Thanks for this and the tests, it looks very good. I think we should add some documentation to explain that downgrades are possible, probably in `Migrator.onUpgrade` or the typedef it uses.

> Should I use develop as base to fix the CI issue?

I'd prefer to merge this into `develop`, yes.

davidmartos96:
@simolus3 Alright, I added some docs and I switched the branch to `develop`.

simolus3:
Looks great, thank you :+1:

PR #432, opened 2020-03-06
Handling exceptional case when executing Prepared Statement

jaggernod:
This Pull Request contains 2 changes:

1. Adds more details to `SqliteException`s. 
`explanation` now contains the [extended result code](https://www.sqlite.org/rescode.html#extrc). 
I have stumbled upon it when started getting generic `(10) SQLITE_IOERR`. 
Now debugging such issues will be much simpler.
2. Does not swallow exceptional cases while iterating over rows while executing a `select` on a `PreparedStatement`.

BTW I was not able to find a way to create an exceptional case to write a test for **#2**. 
I stumbled upon the missing throw when I started getting `(6410) SQLITE_IOERR_GETTEMPPATH`.
But this issue appeared because [Android doesn't have any of the default Unix paths](https://github.com/SRombauts/SQLiteCpp/issues/128#issuecomment-304501634)

Hope the PR is helpful. It is my birthday PR ;)


simolus3:
Thanks for this helpful PR, and happy birthday :tada:

I've found a way to test exceptions in `select` by using an application-defined function. Otherwise this looks very good :+1:

Issue #221, opened 2019-11-01
Proper batch api and updateAll support


JulianBissekkou:
Since `insert` and `insertAll` are available I was wondering if there is an `updateAll` next to `update`.
I couldn't find anything in the documentation. Am I missing something?

Mike278:
`insertAll` exists because moor can generate a single sql `INSERT` statement that inserts multiple rows.
```sql
INSERT INTO users (user_id, first_name) VALUES
(1, 'a'),
(2, 'b'),
(3, 'c')
```

While there are [some fancy sql queries](https://stackoverflow.com/questions/2528181/update-multiple-rows-with-one-query) that can update multiple rows with different values conditionally, I wouldn't expect that as a default provided by moor.

simolus3:
_TL;DR_: We should probably add a proper batch api to moor.

----
To my own surprise, we actually don't generate a single insert statement in `insertAll`, we instead create what's essentially a `Map<String, List<List<Variable>>>` of statements. Then, each key is a sql statement and we have a list of bound variables to run it with. I imagine that this can be faster than a single query for (very) large inserts because you skip duplicate parsing costs. The original motivation for adding `insertAll` was that we get to use the batch api from sqflite, which is definitely faster because we have to send less data over platform channels.

We could introduce something similar for updates, but I think that only makes sense as an `replaceAll`, because then each entity is likely to only affect one row. If we had an `updateAll`, then
```dart
(update(table)..where((row) => predicate))
  .writeAll([firstUpdate, secondUpdate, thirdUpdate])
```
would realistically only apply the `thirdUpdate`, others would be overridden. (That's not technically true if each update only applies one column, but in that case you could just merge them into a single update).

Adding a proper `updateAll` only makes sense if each update has their own where clause. If we add that, we should probably introduce a proper `batch` api to moor that replaces `insertAll`:
```dart
await batched((Batch b) {
   // none of these methods should return a future, a batch should work atomically
   b.insertAll(...);
   b.update(); // can be used in a for loop to achieve update-all behavior that works atomically
   b.replaceAll(...);
});
```
Calling `batched` inside a transaction would probably not be supported, as we don't support nested transactions at the moment.

JulianBissekkou:
We are currently evaluating moor as db solution for a bigger (database heavy) app. 
It looks like this feature is definitely needed since the current performance of updating multiple items is way to slow.
I create a quick performance test that does operations on 10000 items.
`into(myTable).insertAll(items)` took 300ms

[...]

PR #437, opened 2020-03-10
Update testing.md

krabbenprgr:
Added testing example from commit #221  to documentation


simolus3:
Thanks for the contribution. However, I'm not sure if the testing page is the best place to put that example.

Since that test doesn't verify anything, maybe it's better to instead expand the regular documentation on batches? Batches could get a proper subsection [in this article](https://moor.simonbinder.eu/docs/getting-started/writing_queries/#inserts) with an example that also uses updates.

PR #444, opened 2020-03-15
Remove sqflite_sqlcipher git dependency

davidmartos96:
I've made a pub package for the SQLCipher plugin, so the git dependency is not needed anymore. The reference for the plugin is in the branch `publish` in the repository https://github.com/davidmartos96/sqflite_sqlcipher

### DIFF
--- a/docs/content/en/docs/Other engines/encryption.md
+++ b/docs/content/en/docs/Other engines/encryption.md
@@ -3,18 +3,10 @@ title: Encryption
 description: Use moor on encrypted databases
 ---
 
-{{% alert title="Security notice" color="warning" %}}
-> This feature uses an external library for all the encryption work. Importing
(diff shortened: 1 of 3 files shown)


simolus3:
Thanks for this PR and your general work on `sqflite_sqlcipher`!

Can you also update the documentation under [`docs/content/en/docs/Other engines/encryption.md`](https://github.com/simolus3/moor/blob/master/docs/content/en/docs/Other%20engines/encryption.md)? The "security notice" section doesn't have to warn about `sqflite_sqlcipher` being a git dependency anymore. And I think the proguard info from your readme should be added as well, or there should be a link to your pub package explaining the further setup that would have to be done. Thanks!

davidmartos96:
@simolus3 I updated the docs.

Issue #468, opened 2020-04-03
Support for Web Workers


marcgraub:
## Context
Hi! I made an offline web app, it synchronizes data with an API on the background and stores it in a local database with moor.

## Problem
**The problem I have comes when I make inserts to the database, the UI freezes for some ms**, for example, every ten minutes I check for new clients and normally, I make inserts with batches of 50.

## Solution
I've seen you have been working on implementing IndexedDB as a storage engine, it would be great to give the implementation some basic support for Web Workers.

With the current implementation the `window` object isn't accessible from the worker, we can pass a bool to indicate the IndexedDB lives inside a worker and use `WorkerGlobalScope` to get access to IndexedDB.

I gonna do a PR proposal.

Another concern I have is if stream queries are going to be updated when table receives an insert in another web worker. I think not.
If not, exists some method to trigger manually the auto update of a query stream?

I made some tests with the code and web workers seems to solve my problem.

festelo:
Hi @simolus31 Would you like to check https://github.com/festelo/moor/tree/master_worker/moor/lib/src/web/worker ?
Transactions are not implemented currently.
To make it work - [this](https://github.com/festelo/moor/blob/master_worker/moor/lib/src/web/worker/worker.dart) file should be compiled into js and placed somewhere in the app.

Usage:
`
  final storage = w.MoorWebStorageFactory.indexedDb(
    name,
    inWebWorker: true,
    migrateFromLocalStorage: false,
  );
  return w.WebDatabase.withDelegate(
      w.MoorWorkerClient(workerPath, sqlJsPath, storage);
`
inWebWorker *must* be true and migrateFromLocalStorage *must* be false.
workerPath - path to the compiled file.
sqlJsPath  - path to sql-js script.

festelo:
I made this because I had performance issue. But as i found out later, the problem was not in moor... Anyway maybe it will be helpful

kanekv:
How does web storage work? Does it overwrite whole database on every modification?

simolus3:
Unfortunately that's what we need to do, yes.

kanekv:
@simolus3 thanks a lot! Is there a way to manually manage when save happens? E.g. not save it on every modification but once in a while?

simolus3:
Sorry for the slow response. You can wrap the `MoorWebStorage` to avoid storing all the time:

```dart
class _OnlyStoreWhenNeeded implements MoorWebStorage {
  final MoorWebStorage inner;
[...]

PR #469, opened 2020-04-03
Basic support for Web Workers

marcgraub:
With the current implementation the `window` object isn't accessible from the worker, we can pass a bool to indicate the IndexedDB lives inside a worker and use `WorkerGlobalScope` to get access to IndexedDB.

Related: #468

### DIFF
--- a/moor/lib/src/web/storage.dart
+++ b/moor/lib/src/web/storage.dart
@@ -39,20 +39,32 @@ abstract class MoorWebStorage {
   /// However, older browsers might not support IndexedDB.
   @experimental
   factory MoorWebStorage.indexedDb(String name,
-      {bool migrateFromLocalStorage}) = _IndexedDbStorage;
+      {bool migrateFromLocalStorage, bool inWebWorker}) = _IndexedDbStorage;
 
   /// Uses [MoorWebStorage.indexedDb] if the current browser supports it.
   /// Otherwise, falls back to the local storage based implementation.
-  factory MoorWebStorage.indexedDbIfSupported(String name) {
-    return supportsIndexedDb
-        ? MoorWebStorage.indexedDb(name)
+  factory MoorWebStorage.indexedDbIfSupported(String name,
+      {bool inWebWorker = false}) {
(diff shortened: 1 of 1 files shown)


simolus3:
Thanks for your contribution!

To help me test this, do you have a special set up to compile Dart to a web worker js file?

I haven't looked at web workers in detail, so I'm wondering if we can have some rpc-like communication between a worker and the main app. We could then use the existing isolate support code to also synchronize table updates and query streams over that web worker. Then we'd also have better synchronization across multiple tabs using the same database, I assume.
All of that is not in the scope of this PR of course, I'm just wondering if you had some experience with similar approaches you could share.

marcgraub:
It's the first time I use Web workers.

I've created a basic project to test moor with Web Workers, you can see it in [moor-web-worker](https://github.com/marcgraub/moor-web-worker), I'm using the library [dorker](https://pub.dev/packages/dorker).

moor/lib/src/web/storage.dart
```diff
@@ -39,20 +39,32 @@ abstract class MoorWebStorage {
   /// However, older browsers might not support IndexedDB.
   @experimental
   factory MoorWebStorage.indexedDb(String name,
-      {bool migrateFromLocalStorage}) = _IndexedDbStorage;
+      {bool migrateFromLocalStorage, bool inWebWorker}) = _IndexedDbStorage;
```
simolus3, line 42: Let's add some documentation on what `inWebWorker` is doing here. Maybe something like: "When [inWebWorker] is set (defaults to false), the implementation will use [WorkerGlobalScope] instead of [window]".

moor/lib/src/web/storage.dart
```diff
@@ -169,7 +186,7 @@ class _IndexedDbStorage implements MoorWebStorage {
   @override
   Future<Uint8List> restore() async {
     final transaction =
-        _database.transactionStore(_objectStoreName, 'readonly');
+    _database.transactionStore(_objectStoreName, 'readonly');
```
simolus3, line 189: Please run `dartfmt -w` (or `flutter format`) on this file.

PR #492, opened 2020-04-13
doc: update example to include missing imports

abumalick:
`dart:io` is necessary for `File`
`moor.dart` is necessary for `LazyDatabase`

Thanks for the library 👍


simolus3:
Thanks!

PR #494, opened 2020-04-15
Integration tests improvements

davidmartos96:
After coming back to the `moor_ffi` strange behavior on some 32 bit devices and trying it again in my old phone, I noticed that using a memory database wasn't resulting in an API misuse error. 
I took a closer look to the issue and the problem was that the folder database didn't exist in the app folder, so all test were failing with that error.
In the PR I made 3 changes.
1. Make sure that the folder for the target test db exist.
2. Delete the db before running the asset db test, since it was failing from previous runs.
3. Rename the 2 files in the integration tests lib folder
     ffi.dart => moor_ffi.dart
     main.dart => moor_flutter.dart

The third change is not really necessary, but I think it is clearer to see what is being tested.

I will delete my comment in the other issue thread since `moor_ffi` it's working for me. Your `arm-ffi` project is working for me as well.


simolus3:
Thanks for trying it out again! I probably missed this because I always ran the `moor_flutter` integration test first, which created the folder. I also agree with the renaming, as you said it makes testing clearer.

Pretty weird that we got a misuse error from that though, I thought they were only issued for "this is obviously and always wrong" scenarios. I'll investigate some more and probably make `moor_ffi` create the directory if doesn't exist in the next version.

PR #511, opened 2020-04-20
[moor_ffi] Add readOnly support to Database.open 

alextekartik:
This is in response to the following issue: https://github.com/simolus3/moor/issues/510

I choose an opinionated readOnly flag for Database.open but you can decide to change that. I have not added the flag to openFile. Let me know if you want it too.

I added a small unit test too.


simolus3:
Looks good, thanks!

PR #534, opened 2020-05-03
Fix null table prefix

Mike278:


simolus3:
Thanks :+1: I've uploaded version `3.0.1` which contains this fix.

PR #536, opened 2020-05-03
Fix documentation source code issues

mqus:
Just some typos

Btw, is the develop branch the right one? The contributing guidelines say yes but looking at the commits this doesn't seem as true...


simolus3:
Thanks for the fix :+1: 

> The contributing guidelines say yes but looking at the commits this doesn't seem as true...

`develop` is currently behind `master` because I had to make some quick fixes after releasing moor 3.0 and sqlparser 0.8.0. In general it's the right branch.

Issue #537, opened 2020-05-03
SqlParser: derive tables involved in a query


mqus:
Hi! I looked at your sqlparser library and it seems to be very powerful. As you suggested in https://github.com/vitusortner/floor/issues/94 it fits perfectly well to derive the resulting types of a query and I will/want to integrate it into floor. But there is one more thing I would like to do with sqlparser and that is deriving the tables/views that are involved in a query. I would then want to have two groups of tables which are identified:
1. Tables the query directly depends on. Meaning if the data in one of those dependencies changes, the query could then have different results.
2. The tables which are directly altered by the query (especially for delete/update/insert queries)

With the knowledge of the engine (foreign key relations, database views, maybe trigger queries) it would then be possible to 

3. Derive the tables the query indirectly depends on.

This would allow us to update streaming queries more precisely.

I would like to limit the scope a little bit to queries that change the data and not the scheme of the database. So CREATE/DROP statements would be out of scope as well as ALTER TABLE etc.

With this outlined functionality, would you be interested in having it in the sqlparser?
Do you think this is possible? Are there some pieces missing before I could implement this?

If yes, where should I start implementing this?

I would like to try adding this functionality, but I can't promise anything.

simolus3:
To find tables involved in a query, you could use something like [these visitors](https://github.com/simolus3/moor/blob/master/moor_generator/lib/src/analyzer/sql_queries/affected_tables_visitor.dart). Moor uses them to find out which tables a query reads from and which tables it updates. You can use them like this:

```dart
final engine = SqlEngine(); // add tables, etc.
final context = engine.analyze('INSERT INTO foo SELECT * FROM bar');

final tableDetector = UpdatedTablesVisitor()..visit(context.root);
// tableDetector.writtenTables contains foo with UpdateKind.insert
// tableDetector.foundTables bar
```

The visitors aren't part of the `sqlparser` package because I thought this was mainly a moor-specific functionality. I can move those to `sqlparser` if you need them as well.

>  Derive the tables the query indirectly depends on.

This is also something I'm working on for moor. We essentially create a graph to propagate table updates, so that we can say "an insert on table x causes an update on table y and a delete on table z". moor_generator does that [here](https://github.com/simolus3/moor/blob/master/moor_generator/lib/src/services/find_stream_update_rules.dart), but it uses a moor-specific representation of triggers and tables, not the `Table`s you'd use for `sqlparser`. So far we only consider triggers but I want to add support for foreign key constraints as well (#528).
[...]

PR #563, opened 2020-05-13
Sqlparser: Add View support

mqus:
Partially adresses #537 by adding views.

I started working on it (the parser works somewhat) but got stuck because I couldn't really understand how the analyzer works. Could you help me by providing a mid-to-high-level overview and what classes do which parts or pointing me to a good reference?

The things I added to the analyzer are just copy'n'pasted bits from the table but I could not really figure out how to fit it into the rest.

Some Things that are open apart from that:

**Parser**
- [x] Write code
- [x] Write tests
- [x] Check documentation and add if necessary

**Analyzer**
- [x] Write code
- [x] Write tests
- [x] Check documentation and add if necessary


sqlparser/lib/src/analysis/schema/column.dart
```diff
@@ -103,6 +103,73 @@ class TableColumn extends Column {
   }
 }
 
+/// A column that is part of a table.
+class ViewColumn extends Column {
```
simolus3, line 107: I think the best way to represent a `ViewColumn` would be to make this mix in `DelegatedColumn`, where the `innerColumn` is a field holding the column from the select statement used in the create view statement.

We can then remove pretty much everything from this class except from `humanReadableDescription`.

sqlparser/lib/src/analysis/schema/view.dart
```diff
@@ -0,0 +1,104 @@
...
+      resolvedColumns.where((c) => c.includedInResults).toList();
+
+  /// Whether this table was created with an "WITHOUT ROWID" modifier
+  final bool withoutRowId;
+
+  /// Additional constraints set on this table.
+  final List<TableConstraint> tableConstraints;
+
+  /// The ast node that created this table
+  final TableInducingStatement definition;
```
simolus3, line 32: This should be a `CreateViewStatement`, accordingly.

sqlparser/lib/src/ast/visitor.dart
```diff
@@ -122,6 +123,11 @@ class RecursiveVisitor<A, R> implements AstVisitor<A, R> {
     return visitTableInducingStatement(e, arg);
   }
 
+  @override
+  R visitCreateViewStatement(CreateViewStatement e, A arg) {
+    return visitStatement(e, arg);
```
simolus3, line 128: Since `CreateViewStatement` is a `SchemaStatement`, let's use the more accurate

```suggestion
    return visitSchemaStatement(e, arg);
```

mqus:
> Finally, we need some way to add a ResultSet that isn't a table to the SqlEngine. I'd say we can generalize registerTable to registerResultSet.

I assume that we still need the `name` of the view/table, which is not part of the ResultSet. So I would rather add a registerView function, or a registerNamedResultSet function (with an extra NamedResultSet class for tables and views)... or do you have another idea?

sqlparser/lib/src/reader/parser/schema.dart
```diff
@@ -247,6 +249,37 @@ mixin SchemaParser on ParserBase {
...
+  /// Parses a [CreateViewStatement]. The `CREATE` token must have already been
+  /// accepted.
+  CreateViewStatement _createView() {
+    final create = _previous;
+    assert(create.type == TokenType.create);
+
+    if (!_matchOne(TokenType.view)) return null;
+
+    final ifNotExists = _ifNotExists();
+    final name = _consumeIdentifier('Expected a name for this index');
```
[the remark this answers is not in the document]
mqus: whoops, fixed

sqlparser/lib/src/reader/parser/schema.dart
```diff
@@ -247,6 +249,37 @@ mixin SchemaParser on ParserBase {
...
+
+    List<String> columnNames = null;
+    if (_matchOne(TokenType.leftParen)) {
+      columnNames = _columnNames();
+      _consume(TokenType.rightParen, 'Expected closing bracket');
+    }
+
+    _consume(TokenType.as, 'Expected AS SELECT');
+
+    final query = select();
```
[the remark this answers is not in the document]
mqus: I didn't know the difference, is fixed now

sqlparser/lib/src/reader/parser/schema.dart
```diff
@@ -288,6 +321,18 @@ mixin SchemaParser on ParserBase {
       ..setSpan(create, _previous);
   }
 
+  @override
+  List<String> _columnNames() {
```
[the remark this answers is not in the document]
mqus: sure, fixed ;)

sqlparser/lib/src/analysis/schema/view.dart
```diff
@@ -0,0 +1,104 @@
+part of '../analysis.dart';
+
+/// A database view. The information stored here will be used to resolve
+/// references and for type inference.
+class View with HasMetaMixin implements HumanReadable, ResolvesToResultSet {
```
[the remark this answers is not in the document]
mqus: done

sqlparser/lib/src/analysis/schema/column.dart
mqus: thanks, I changed it as you described.

sqlparser/lib/src/analysis/schema/view.dart
mqus: whoops, you're right, fixed :laughing:

sqlparser/lib/src/ast/visitor.dart
mqus: done

sqlparser/lib/src/analysis/schema/view.dart
```diff
@@ -0,0 +1,54 @@
...
+    }
+  }
+
+  @override
+  String humanReadableDescription() {
+    return name;
+  }
+}
+
+List<ViewColumn> _createColumns(
```
simolus3, line 47: Since this is specific to views, I'd make it a private static method in the `View` class.

sqlparser/lib/src/analysis/schema/column.dart
```diff
@@ -103,6 +103,30 @@ class TableColumn extends Column {
   }
 }
 
+/// A column that is part of a table.
```
simolus3, line 106:
```suggestion
/// A column that is part of a view.
```

sqlparser/lib/src/analysis/schema/column.dart
```diff
@@ -103,6 +103,30 @@ class TableColumn extends Column {
...
+class ViewColumn extends Column with DelegatedColumn {
+
+  final String _name;
+
+  @override
+  final Column innerColumn;
+
+  /// The table this column belongs to.
```
simolus3, line 114:
```suggestion
  /// The view this column belongs to.
```

sqlparser/lib/src/analysis/schema/column.dart
```diff
@@ -103,6 +103,30 @@ class TableColumn extends Column {
...
+  View view;
+
+  ViewColumn(this.innerColumn, [this._name]);
```
simolus3, line 117:
```suggestion
  /// Creates a view column wrapping a [Column] from the select statement used to
  /// create the view.
  ///
  /// The optional name parameter can be used to override the name for this column.
  /// By default, the name of the [innerColumn] will be used.
  ViewColumn(this.innerColumn, [this._name]);
```

mqus:
I made View and Table extend NamedResultSet because NamedResultSet is no longer a mixin if it extends anything other than an object

mqus:
Do you have some specific things you want to see tested for the analyzer part? I'm currently a bit at a loss where to put the tests and what to put in it... I want to at least make a simple view and then query something from it, but i'm missing ideas for anything apart from that...

simolus3:
> I want to at least make a simple view and then query something from it

That's a good start to verify that this is working from start to end. You can replicate something like [this test](https://github.com/simolus3/moor/blob/b5cc9b974d34d6227d50a0b1b21c88e91c628736/sqlparser/test/analysis/reference_resolver_test.dart#L9-L42) for views to make sure that column names and types are inferred correctly.

Additionally, you could test `SchemaFromCreateView.readView` directly to make sure the view has the expected name and columns.

Have you seen https://github.com/mqus/moor/pull/1? You can take that as an inspiration if you want to, but it's fine to me if you go in a different direction. One of the things I'd prefer is to report a lint when `columnNames.length != selectColumns.length` instead of making it an assertion. Lints generally have better UX since users can see an explanation of what's wrong. We should have another test to verify that analysis behaves as expected for something like `CREATE VIEW my_view (foo, bar) AS SELECT 1;`

mqus:
Thank you very much for your patience and for the thorough guidance! I have no issues with your suggested changes at all ;)

> We should have another test to verify that analysis behaves as expected for something like CREATE VIEW my_view (foo, bar) AS SELECT 1;

That's still missing, right? I'll take a look and should be finished today.

simolus3:
Great, thank you again for your contribution :+1:

mqus:
I think if you would have done it alone then it would have went a lot faster ;D but thank you for the help!

PR #589, opened 2020-05-22
Fix/min analyzer version missmatch

kuhnroyal:
`languageVersionMajor/Minor` was added in analyzer version 0.39.5 and is used since 4d463dd1453ad9f205df286688d1e6dba906d65e


simolus3:
Thanks :+1: 

In my defense, I searched through its changelog and couldn't find anything, so I assumed that those getters have existed for a long time.

PR #605, opened 2020-05-28
Add toString to companions

friebetill:
This is very useful for unit tests and debugging, because we can see 
directly what values a companion has.

For example in this error message it is difficult to know which fields 
are set incorrectly:
```
ERROR: No matching calls. All calls: MockTestsDao.upsert(Instance of 'TestsCompanion')
```

With `toString` the error message is easier to read:
```
ERROR: No matching calls. All calls: MockTestsDao.upsert(TestsCompanion(id: Value.absent(), testValue: Value(5)))
```

### DIFF
--- a/moor/example/example.g.dart
+++ b/moor/example/example.g.dart
@@ -127,6 +127,15 @@ class CategoriesCompanion extends UpdateCompanion<Category> {
     }
     return map;
   }
+
+  @override
+  String toString() {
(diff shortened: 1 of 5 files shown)


simolus3:
Thanks for the PR :+1: The test failure is due to the analyzer acting up, so I'll merge this.

PR #619, opened 2020-06-08
Add hashcode and equals to value

friebetill:
This simplifies testing, as one can compare Values like this:
```dart
expect(Value(1), Value(1));
```
Or a realistic example:
```dart
final capturedArgument = verify(fooDao.insert(captureAny)).captured.first.createdAt;
expect(capturedArgument, Value(DateTime(0)));
```

A test is still missing which would look something like this:
```dart
test('value support hash and equals', () {
  const first = Value(0);
  final equalToFirst = Value(0);
  const different = Value.absent();

  expect(first.hashCode, equalToFirst.hashCode);
  expect(first, equals(equalToFirst));

  expect(first, isNot(equals(different)));
  expect(first, equals(first));
});
```
I'm not sure where the test is supposed to be.

### DIFF
--- a/moor/lib/src/runtime/data_class.dart
+++ b/moor/lib/src/runtime/data_class.dart
@@ -125,6 +125,17 @@ class Value<T> {
 
   @override
   String toString() => present ? 'Value($value)' : 'Value.absent()';
+
+  @override
+  bool operator ==(Object other) =>
+      identical(this, other) ||
+      other is Value &&
+          runtimeType == other.runtimeType &&
+          present == other.present &&
+          value == other.value;
+
+  @override
+  int get hashCode => present.hashCode ^ value.hashCode;
 }
 
 /// Serializer responsible for mapping atomic types from and to json.
--- a/moor/lib/src/runtime/data_class.dart
+++ b/moor/lib/src/runtime/data_class.dart
@@ -129,10 +129,7 @@ class Value<T> {
   @override
   bool operator ==(Object other) =>
       identical(this, other) ||
-      other is Value &&
-          runtimeType == other.runtimeType &&
-          present == other.present &&
-          value == other.value;
+      other is Value && present == other.present && value == other.value;
 
   @override
   int get hashCode => present.hashCode ^ value.hashCode;
--- a/moor/test/data_class_test.dart
+++ b/moor/test/data_class_test.dart
@@ -107,8 +107,38 @@ void main() {
     expect(first.hashCode, equalToFirst.hashCode);
     expect(first, equals(equalToFirst));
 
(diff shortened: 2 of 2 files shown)


simolus3:
Thanks for this!

> I'm not sure where the test is supposed to be.

You can put it into `moor/test/data_class_test.dart` (or make a new test file, but I'm not sure if that's worth it in this case).

moor/lib/src/runtime/data_class.dart
```diff
@@ -125,6 +125,17 @@ class Value<T> {
 
   @override
   String toString() => present ? 'Value($value)' : 'Value.absent()';
+
+  @override
+  bool operator ==(Object other) =>
+      identical(this, other) ||
+      other is Value &&
+          runtimeType == other.runtimeType &&
```
simolus3, line 133: I'm not sure if we should always consider the `runtimeType` here. For instance, one might expect `Value<bool>.absent()` to be equal to `Value<String>.absent()` since they have identical behavior. I'm not so sure about `Value<String>(null) == Value<bool>(null)` though.

The `runtimeType` comparison is commonly used to guard against unexpected subclasses messing up equality, but I think we don't need that in this case. Maybe we can change this to `(!present || other.runtimeType == runtimeType)`? I don't really have a strong opinion on this, maybe something else makes more sense.

friebetill: Thanks for the suggestions!

I don't know what the most appropriate way is either, but I agree with your argumentation in principle. I still see the problem that the hash code is the same in this case:
```dart
Value<String>(null).hashCode == Value<bool>(null).hashCode`
```
but `equals` in this case is `false`. 
As far as I can see, there are two possibilities: 
1. We remove `(!present || other.runtimeType == runtimeType)` 
2. We adjust the `hashCode` so that the generic `T` matters, e.g. `(present ? T.hashCode : 0) ^ present.hashCode ^ value.hashCode;` (haven't tested it yet)

I decided to go for the first version first, because it's simpler. But I don't have a strong opinion and I'll adjust it if you like.

I also noticed that in the test `companions support hash and equals` the line
```dart 
expect(first, equals(first));
```
doesn't make sense and I adjusted it accordingly.

simolus3: I think it's ok for `hashCode` to be less precise than `==`. But just using `value` and `present` looks good to me as well, I think we can keep it like that.

However, I'd really like to a test case to ensure `Value.absent() != Value(null)`.

friebetill: Done 👍

simolus3:
Looks good now, thanks!

PR #634, opened 2020-06-17
Sqlparser: Allow Backticks(`) for escaping identifiers

mqus:
Hi! I noticed that moor/sqlparser does not accept backticks as delimiters of identifiers. As sqlite *does* accept them, to retain some compatibility with mysql (see https://sqlite.org/lang_keywords.html), I thought it might be good to add them.

If you want me to add tests for this, please tell me what you want to see :)

### DIFF
--- a/sqlparser/lib/src/reader/tokenizer/scanner.dart
+++ b/sqlparser/lib/src/reader/tokenizer/scanner.dart
@@ -183,13 +183,13 @@ class Scanner {
         _string();
         break;
       case $double_quote:
-        _identifier(escapedInQuotes: true);
+        _identifier(escapeChar: $double_quote);
         break;
       case $backquote:
         if (scanMoorTokens) {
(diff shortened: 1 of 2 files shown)


simolus3:
Thanks for this PR!

> please tell me what you want to see :)

One simple test in `test/scanner/scanner_test.dart` should do. Something like

```dart
test('scans identifiers with backticks', () {
  expect(
    Scanner('`SELECT`').scanTokens(),
    contains(isA<IdentifierToken>()
        .having((e) => e.identifier, 'identifier', 'SELECT')),
  );
});
```

mqus:
There you go :)
The failing tests should be unrelated.

simolus3:
Thanks again! Yeah, the CI failure seems to be unrelated. Testing `moor_generator` works locally, so it should be fine.

PR #639, opened 2020-06-19
Docs links move /kripken/sql.js ->/sql-js/sql.js

festelo:
As described at https://github.com/kripken/sql.js the repo has been moved and due to this some links look outdated. 

Next issue:
At https://moor.simonbinder.eu/web/ there's text "You can grab the latest version of sql-wasm.js and sql-wasm.wasm *here* "
Now 'here' leads to 404. If we replace /kripken/ with /sql-js/ the link will look like https://github.com/sql-js/sql.js/tree/master/dist. The page exists, but content is not we want - there's no any *.js files. 
In this PR the link is replaced with https://github.com/sql-js/sql.js/releases


simolus3:
Thanks :+1:

PR #696, opened 2020-07-13
Make DefaultValueSerializer null-safe for blobs

nachoapps:
This fixes an issue where a nullable blob column, e.g. `BlobColumn get logic => blob().nullable()();` will throw an error that looks like `NoSuchMethodError: The method 'cast' was called on null.`


simolus3:
Thanks for this PR! Can you rebase it on `develop` so that I can integrate it with the next moor release?

nachoapps:
Sorry, I should have read your contributing guide more carefully!

Also, I'm apparently too stupid to figure out how to do this properly so instead I've just created a new branch/PR.

PR #697, opened 2020-07-13
Make DefaultValueSerializer null-safe for blobs

nachoapps:
This fixes an issue where a nullable blob column, e.g. `BlobColumn get logic => blob().nullable()();` will throw an error that looks like `NoSuchMethodError: The method 'cast' was called on null.`

(cherry picked from commit 87937ba89b28e956963555d7a456b61b67500043)


simolus3:
Thanks!

nachoapps:
Happy to contribute, I've been using your library for a while now and I love it.

PR #870, opened 2020-10-13
Support table migration with autoindexes

kuhnroyal:
Not sure this is the correct solution but it seems that autoindexes are always derived from the `CREATE TABLE` statement/constraints and should not be migrated by manually.


simolus3:
I agree with your changes, thanks! 

The failures seem to come from the analysis server suddenly complaining about generated files, so I'll merge and take a look at that later.

PR #904, opened 2020-11-04
Fixed typo. j doesn't exist. The variable is p.

Dhi13man:
The path library has been imported as p not j.


simolus3:
Indeed, thanks :+1:

PR #908, opened 2020-11-06
Docs: Update command and sample code folder names to match descriptions

bdlindsay:
I'm looking at trying out the migration test code generation. I got initially caught up on some of the commands listed in the docs because the folder containing the generated schemas didn't match between the schema generation docs and the test code generation docs. I've just made updates so that the naming is consistent between the two sections.

Additional Changes:
+ vs code's auto-fix linting for `.md` files


simolus3:
Thanks :+1:

PR #936, opened 2020-11-25
sqflite_sqlcipher nnbd version

davidmartos96:
I've migrated the sqlcipher package to null safety.
Locally I see a bunch of errors in packages such as the generator or the integration tests. Is that normal? I suppose not all the project is yet migrated.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,5 +8,6 @@ pubspec.lock
 .dart_tool/
 
 benchmark_results.json
+.flutter-plugins
 .flutter-plugins-dependencies
 flutter_export_environment.sh
--- a/extras/encryption/pubspec.yaml
+++ b/extras/encryption/pubspec.yaml
@@ -4,11 +4,11 @@ version: 1.0.0
 author: Simon Binder <oss@simonbinder.eu>
 
 environment:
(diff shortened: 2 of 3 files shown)


simolus3:
Thanks for this PR.

> Locally I see a bunch of errors in packages such as the generator or the integration tests. Is that normal? I suppose not all the project is yet migrated.

The project isn't fully migrated yet, but there shouldn't be any analysis errors. Can you run `pub get` in the affected packages?

davidmartos96:
It may be VSCode having a hard time with the multiproject setup. 
I tried `pub get` inside `extras/integration_tests/tests` and it shows the following though:
```
Resolving dependencies... (2.3s)
Because every version of moor_generator from path depends on build ^1.1.0 which depends on convert ^2.0.0, every version of
  moor_generator from path requires convert ^2.0.0.
And because every version of moor from path depends on convert ^3.0.0-nullsafety, moor_generator from path is incompatible with
  moor from path.
So, because tests depends on both moor from path and moor_generator from path, version solving failed.
```
Also, I'll try to make another PR, because with the VSCode errors I missed that there were some in the encryption delegate file.
List<dynamic> vs List<Object> mostly.

PR #938, opened 2020-11-25
nnbd migration of encrypted_moor and moor_flutter

davidmartos96:
Missing changes from #936 

I'm not sure how you'd like to handle the `List<Object>` casting. The null safety API from sqflite now expects `List<Object>` instead of `List<dynamic>` because null values cannot be bound as arguments.
In the end I was able to run the integration tests correctly. I had to override the `convert` dependency with the null safety version though.

### DIFF
--- a/extras/encryption/lib/encrypted_moor.dart
+++ b/extras/encryption/lib/encrypted_moor.dart
@@ -6,7 +6,6 @@ library encrypted_moor;
 
 import 'dart:async';
 import 'dart:io';
-import 'package:meta/meta.dart';
 import 'package:path/path.dart';
 import 'package:moor/moor.dart';
 import 'package:moor/backends.dart';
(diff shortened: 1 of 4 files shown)


simolus3:
Thanks for the follow-up PR.

> The null safety API from sqflite now expects `List<Object>` instead of `List<dynamic>` because null values cannot be bound as arguments

I don't think that's great (https://github.com/tekartik/sqflite/issues/529#issuecomment-734342275), but it should be fine for moor. The `as List<Object>` seems a bit fragile though, can you use `.cast<Object>()`? Maybe we won't need it at all if sqflite changes their apis to allow nullable arguments.

davidmartos96:
@simolus3   Ok, we can keep it open then. Let's wait if the API can be changed in sqflite. I see that you have already told Alexandre about it.
I wasn't aware that null binding is possible with the moor API. Thanks for letting me know!

simolus3:
FWIW, I changed the `convert` dependency in the core `moor` package in 0259fd58fd17c9ae6c5616d79d8ef1d97e454c29, with that we hopefully don't need that many `dependency_overrides` anymore. 

I also changed the `args` parameter to be a `List<Object?>` instead of dynamic and make `runSelect` return a `Future<List<Map<String, Object?>>>` instead of a map with dynamic values. It'd be great if you could apply that to `moor_flutter` and `encrypted_moor`, but I can do that too after this PR is merged.

davidmartos96:
@simolus3 I rebased with your changes and changed the API to use `Object?`.

simolus3:
Thanks :+1: This looks good to me now.

PR #966, opened 2020-12-12
add warning about the generated compare method

victorcatn:
I think it's important to be aware of that possible problem.


simolus3:
Thanks :+1:

PR #975, opened 2020-12-18
Update testing.md to make Windows testing docs clearer

cervonwong:
I got errors before realizing that I needed to restart my device.

(The diff is weird again :p, lines 113/118 should be identical)


simolus3:
Sounds good, thanks!

PR #985, opened 2020-12-22
Correctly fallback to LocalStorage in private mode on Firefox when using indexedDbIfSupported()

### LINKED ISSUE #982: MoorWebStorage.indexedDbIfSupported() tries to use IndexedDB on Firefox private mode and fails to fallback
Moor on web currently fails to fallback to LocalStorage when IndexedDB is unavailable in private mode on Firefox. This is because window.indexedDB is non-null in private mode on Firefox even though it isn't available for use.

The error handling to fallback to LocalStorage should be more robust to handle this case like [here](https://stackoverflow.com/questions/31159316/detecting-preventing-firefox-indexeddb-invalidstateerror-caused-by-private-brows).

Sidenote: Thank you @/simolus3 for your awesome work on Moor!

CDDelta:
`indexedDbIfSupported()` is no longer a factory as factories cannot be async.

You'll probably want to squash this PR 😅

Fixes #982

Edit: Failed to rebase the previous branch properly, here's a new one.

### DIFF
--- a/docs/content/en/docs/Other engines/web.md
+++ b/docs/content/en/docs/Other engines/web.md
@@ -90,7 +90,7 @@ performant, since we don't have to encode binary blobs as strings.
 To use this implementation on browsers that support it, replace `WebDatabase(name)` with:
 
 ```dart
-WebDatabase.withStorage(MoorWebStorage.indexedDbIfSupported(name))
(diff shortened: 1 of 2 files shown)


simolus3:
This looks good now, thanks!

PR #986, opened 2020-12-27
Update columns.dart to fix typo in doc comments

cervonwong:
Not sure which word is meant to be used, but fixed up the phrasing to make it sound better.


simolus3:
Sounds good, thanks again!

PR #994, opened 2021-01-06
Replace moor_migrations by moor_schemas

lsaudon:
According to the rest of the documentation, the folder is moor_schemas and not moor_migrations.


simolus3:
Thanks!

PR #1004, opened 2021-01-15
Fix converter data types when NNBD is off

kuhnroyal:
This should fix the converter data types having a nullable suffix when NNBD is off.
Discovered/Documented in https://github.com/simolus3/moor/issues/964


simolus3:
Good catch, thanks for the fix :+1:

PR #1036, opened 2021-02-08
Fix typo fom -> from

lsaudon:


simolus3:
Thanks! :+1: 

Merging since the failed analysis appears to be caused by a Dart upgrade.

PR #1043, opened 2021-02-10
added support for CREATE VIEW in .moor files

### LINKED ISSUE #162: Declare views in .moor files
This is not planned for 2.0, but it would be nice to eventually support `CREATE VIEW` and `CREATE INDEX` statements in moor files.

We should probably generate a dataclass for each `CREATE VIEW` statement. Just as we do with tables, queries with the same columns should then return that data class instead of introducing another one.

`CREATE INDEX` statements should be executes as the database gets created. I'm not entirely sure how we would integrate them with the migrations api. Maybe we could have something like
```sql
myIndex: CREATE INDEX ...;
```
and then 
```dart
onUpgrade: (m, from, to) async {
 // ...
 await m.createIndex(myIndex);
}
```

HosseinYousefi:
Hi, I added support for CREATE VIEW as discussed before, obviously this is my first time contributing to this project, so there might be some errors here and there, but I tried to follow the same general style of the project as much as possible.

Now `CREATE VIEW` works like `CREATE TRIGGER` and we can use the views inside the sql queries. I wanted to make this pull request to discuss and later we can work on the documentation and doing the final touches.

Closes #162 

Cheers!

### DIFF
--- a/moor_generator/lib/src/analyzer/moor/entity_handler.dart
+++ b/moor_generator/lib/src/analyzer/moor/entity_handler.dart
@@ -2,9 +2,7 @@ import 'package:moor_generator/moor_generator.dart';
 import 'package:moor_generator/src/analyzer/errors.dart';
 import 'package:moor_generator/src/analyzer/runner/results.dart';
 import 'package:moor_generator/src/analyzer/runner/steps.dart';
-import 'package:moor_generator/src/analyzer/sql_queries/lints/linter.dart';
 import 'package:moor_generator/src/analyzer/sql_queries/query_analyzer.dart';
-import 'package:moor_generator/src/model/view.dart';
 import 'package:sqlparser/sqlparser.dart';
 import 'package:sqlparser/utils/find_referenced_tables.dart';
 
@@ -20,15 +18,16 @@ class EntityHandler extends BaseAnalyzer {
     AnalyzeMoorStep step,
     this.file,
     List<MoorTable> availableTables,
-    List<MoorView> availableViews,
-  ) : super(availableTables, availableViews, step) {
+  ) :
+        // we'll analyze views later, so pass an empty list for now. Otherwise
+        // the incomplete views would be added to the engine.
+        super(availableTables, const [], step) {
     _referenceResolver = _ReferenceResolvingVisitor(this);
   }
 
   final Map<CreateTriggerStatement, MoorTrigger> _triggers = {};
   final Map<TableInducingStatement, MoorTable> _tables = {};
   final Map<CreateIndexStatement, MoorIndex> _indexes = {};
-  final Map<CreateViewStatement, MoorView> _views = {};
 
   _ReferenceResolvingVisitor _referenceResolver;
 
@@ -61,35 +60,19 @@ class EntityHandler extends BaseAnalyzer {
 
         _lint(node, 'special @create table');
         entity.references.addAll(_findTables(node.statement));
-      } else if (entity is MoorView) {
-        final node =
-            _handleMoorDeclaration(entity, _views) as CreateViewStatement;
-        _lint(node, node.viewName);
-        entity.references.addAll(_findTables(node.query));
-        entity.references.addAll(_findViews(node.query));
       }
     }
   }
 
   void _lint(AstNode node, String displayName) {
     final context = engine.analyzeNode(node, file.parseResult.sql);
-    context.errors.forEach(report);
-
-    final linter = Linter(context, mapper);
-    linter.reportLints();
-    reportLints(linter.lints, name: displayName);
+    lintContext(context, displayName);
   }
 
   Iterable<MoorTable> _findTables(AstNode node) {
     final tablesFinder = ReferencedTablesVisitor();
     node.acceptWithoutArg(tablesFinder);
     return tablesFinder.foundTables.map(mapper.tableToMoor);
-  }
-
-  Iterable<MoorView> _findViews(AstNode node) {
-    final tablesFinder = ReferencedTablesVisitor();
-    node.acceptWithoutArg(tablesFinder);
-    return tablesFinder.foundViews.map(mapper.viewToMoor);
   }
 
   Iterable<WrittenMoorTable> _findUpdatedTables(AstNode node) {
--- a/moor_generator/lib/src/analyzer/moor/entity_handler.dart
+++ b/moor_generator/lib/src/analyzer/moor/entity_handler.dart
@@ -4,6 +4,7 @@ import 'package:moor_generator/src/analyzer/runner/results.dart';
 import 'package:moor_generator/src/analyzer/runner/steps.dart';
 import 'package:moor_generator/src/analyzer/sql_queries/lints/linter.dart';
 import 'package:moor_generator/src/analyzer/sql_queries/query_analyzer.dart';
+import 'package:moor_generator/src/model/view.dart';
 import 'package:sqlparser/sqlparser.dart';
 import 'package:sqlparser/utils/find_referenced_tables.dart';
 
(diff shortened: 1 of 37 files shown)


kuhnroyal:
Looks like some great work, maybe I can test this next week.

moor_generator/lib/src/analyzer/sql_queries/query_analyzer.dart
```diff
@@ -28,10 +30,24 @@ abstract class BaseAnalyzer {
...
     }
     return _engine;
   }
 
+  /// Parses the view and adds columns to its resolved columns.
+  @protected
+  void resolveViews() {
+    for (final view in views) {
+      final ctx = _engine.analyzeNode(
+          view.declaration.node, view.declaration.createSql);
```
simolus3, line 44: This is kind of a bad API from me, but the second parameter to `analyzeNode` needs to be the _full source_ containing the view declaration (so the enclosing moor file in this case).

moor_generator/lib/src/analyzer/moor/entity_handler.dart
```diff
@@ -56,6 +61,12 @@ class EntityHandler extends BaseAnalyzer {
 
         _lint(node, 'special @create table');
         entity.references.addAll(_findTables(node.statement));
+      } else if (entity is MoorView) {
+        final node =
+            _handleMoorDeclaration(entity, _views) as CreateViewStatement;
+        _lint(node, node.viewName);
+        entity.references.addAll(_findTables(node.query));
+        entity.references.addAll(_findViews(node.query));
```
simolus3, line 69: References are used to bring entities into the right order (so we ensure that nothing will get created before its dependencies are complete). So essentially, I think that other entity kinds would need to have support for views as well. For instance, if the definition of a trigger contains a select from a view, its `bodyReferences` should contain that view.

We might want to introduce a shared common class for `MoorTable` and `MoorView` (maybe something like `MoorEntityWithResultSet`) and then have a single method for that.

moor_generator/lib/src/analyzer/sql_queries/query_handler.dart
```diff
@@ -121,6 +127,12 @@ class QueryHandler {
       candidatesForSingleTable.clear();
     }
 
+    if (_foundViews.isNotEmpty) {
+      // For now we're not using the single table optimization when selecting
+      // from views.
+      candidatesForSingleTable.clear();
```
simolus3, line 133: Good catch! We could expand the comment to point out that we're not doing this because views don't have data classes yet.

moor_generator/lib/src/analyzer/sql_queries/query_analyzer.dart
```diff
@@ -28,10 +30,24 @@ abstract class BaseAnalyzer {
...
+      views.map(mapper.extractView).forEach(_engine.registerView);
     }
     return _engine;
   }
 
+  /// Parses the view and adds columns to its resolved columns.
+  @protected
+  void resolveViews() {
+    for (final view in views) {
+      final ctx = _engine.analyzeNode(
```
simolus3, line 43: Do we have to do that analysis here? This class is mainly used for analyzing queries and the body of triggers.

Can we analyze and resolve views earlier, say in `MoorParser.parseAndAnalyze`?

moor_generator/lib/src/model/view.dart
```diff
@@ -0,0 +1,47 @@
...
+  /// The associated view to use for the sqlparser package when analyzing
+  /// sql queries. Note that this field is set lazily.
+  View parserView;
+
+  final String name;
+
+  @override
+  List<MoorSchemaEntity> references = [];
+
+  List<ViewColumn> columns;
```
simolus3, line 23: This is probably fine, but we could consider wrapping these with a `MoorColumn`, similar to what `MoorTable` is doing.

simolus3 (commented):
Thank you for this! :tada: 

I have some comments regarding where we resolve views, but otherwise this looks great!

moor_generator/lib/src/analyzer/sql_queries/query_analyzer.dart
HosseinYousefi: Initially I wanted to have it there, but don't we need all views and tables to be registered in the engine prior to analyzing? Since the body of a view is a query I thought that here is a better place, this way if we have some table definitions in another file and we're importing them, we can still create views.

moor_generator/lib/src/analyzer/sql_queries/query_analyzer.dart
HosseinYousefi: My bad, I can change the the second parameter to `view.file` and add a `ParsedMoorFile file;` to `MoorView`.

moor_generator/lib/src/analyzer/sql_queries/query_handler.dart
HosseinYousefi: Added!

moor_generator/lib/src/model/view.dart
HosseinYousefi: We don't need the `columns` right now so I removed it, later we can add it back if we need it.

`MoorColumn` is defined for a table so it has parameters such as `customConstraints` that are not needed for a View. Maybe we can create a custom Column class for views but until then I'll remove it in my next commit.

moor_generator/lib/src/analyzer/sql_queries/query_analyzer.dart
HosseinYousefi: I also added another test that fails the current implementation, making a view from another view, so I need to find a way to resolve views topologically. Queries don't suffer from this because they are created after all tables and import statements.

moor_generator/lib/src/analyzer/moor/entity_handler.dart
HosseinYousefi: Yeah, I was thinking the same.

moor_generator/lib/src/analyzer/sql_queries/query_analyzer.dart
simolus3: > but don't we need all views and tables to be registered in the engine prior to analyzing

Right! Now I remember why I was hesitating to implement views for so long :D

Views referencing other views make the whole situation even harder for sure. If you want to we can ignore that case for a moment (we can add an error message if a view references another view). It's a hard problem and it doesn't map well to the current way moor resolves files.

We currently have a two-step procedure to analyze moor files, and we guarantee that all dependencies have step 1 completed when we enter step 2 of any moor file:

1. Parsing: Parse the moor file and resolve local information (e.g. the table schema) (`ParseMoorStep`)
2. Analysis: Resolve queries and foreign references that need information from other files (`AnalyzeMoorStep`)

We could add another step between them to resolve views, and then ensure that views of dependents have been resolved once we get to step 3. I'm not sure if there's an easier solution that I'm overlooking.

HosseinYousefi: I guess this can be done in the `AnalyzeMoorStep`:

Currently we're using `SqlAnalyzer` prior to the `EntityHandler`, I think if we swap them to first get the references for views, and then have something like `ViewAnalyzer.resolve()` which resolves the views before calling the `SqlAnalyzer.parse()`, that would solve this problem.

HosseinYousefi:
@simolus3 So I added the `ViewAnalyzer` class and also added `visitCreateViewStatement` to `prepare_ast.dart`. The test still doesn't pass which I think is a problem with sqlparser package, I'm committed what I currently have, see if you have any idea why this is happening. I'll continue looking at it tomorrow!

simolus3:
> The test still doesn't pass which I think is a problem with sqlparser package

I think happens because we're essentially analyzing the views twice now:

- `EntityHandler` creates an engine, which registers all moor views. Since views haven't been resolved yet here, the view will be added to the `SqlEngine` without any columns
  - the handler proceeds to `_lint` the view, which causes the error here
- There is a similar problem in `ViewHandler`, since it implicitly adds the views when creating an engine too

I think we can fix this problem by not forwarding views in the constructor of `EntityHandler` and `ViewHandler`. Further, `EntityHandler` needs to stop linting views, this should now happen in `ViewHandler`.

Another problem I just realized is that triggers wouldn't be able to reference views with the current approach. That's okay I guess, we might have to migrate to an approach where we're first analyzing tables only, then views, then everything else in the future.

HosseinYousefi (commented):
Oh, it makes sense, thanks for looking into this.

HosseinYousefi:
I also removed the line saying that views aren't supported in the docs, so hopefully this should be ready to be merged. Let me know if there's anything else that we can work on. Moor is a fantastic package and it has a very well-written code, kodus to you.

simolus3:
Looks good to me now, thanks again for your awesome contribution :tada:

PR #1054, opened 2021-02-17
Exists expression

davidmartos96:
Now that we have subqueries modeled, I think the EXISTS expression would be a good addition.
I'm marking it as a Draft because I'm not sure if you want the function `existsQuery` as top level or if it should be an extension method. `(subquery).exists`.
The docstring is missing as well.
Also, I'm not sure about the `boundVariables` check in the test. I messed around with a very adhoc implementation of subqueries yesterday and I was able to expect it bound to `true` and not 1. Not with your implementation though.


moor/test/expressions/subquery_test.dart
```diff
@@ -0,0 +1,29 @@
...
+    final db = TodoDb();
+    final subquery = db.select(db.users)
+      ..where((tbl) => tbl.isAwesome.equals(true));
+    final existsExpression = existsQuery(subquery);
+
+    final context = GenerationContext.fromDb(db);
+    existsExpression.writeInto(context);
+
+    expect(context.sql, 'EXISTS (SELECT * FROM users WHERE is_awesome = ?)');
+    expect(context.boundVariables, [1]);
```
simolus3, line 17: I have a neat test utility that checks all this at once:

```dart
  expect(existsExpression, generates('EXISTS (SELECT * FROM users WHERE is_awesome = ?)', [1]));
```

You'd have to import `../data/expect_generated.dart`.

moor/lib/src/runtime/query_builder/expressions/exists.dart
```diff
@@ -0,0 +1,35 @@
...
+    if (_not) {
+      context.buffer.write('NOT ');
+    }
+    context.buffer.write('EXISTS ');
+
+    context.buffer.write('(');
+
+    _select.writeInto(context);
+
+    context.buffer.write(')');
```
simolus3, line 23: I don't think we need empty lines between every line here, one after the `EXISTS` should be good enough


```suggestion
    context.buffer.write('(');
    _select.writeInto(context);
    context.buffer.write(')');
```

moor/test/expressions/subquery_test.dart
```diff
@@ -0,0 +1,29 @@
...
+  });
+
+  test('not exists expressions are generated', () {
+    final db = TodoDb();
+    final isInExpression = notExistsQuery(db.select(db.users));
+
+    final context = GenerationContext.fromDb(db);
+    isInExpression.writeInto(context);
+
+    expect(context.sql, 'NOT EXISTS (SELECT * FROM users)');
```
simolus3, line 27: After you added the import, we can also apply it here

```suggestion
    expect(isInExpression, generates('NOT EXISTS (SELECT * FROM users)'));
```

moor/lib/src/runtime/query_builder/expressions/expression.dart
```diff
@@ -119,6 +119,14 @@ abstract class Expression<D> implements FunctionParameter {
   }
 }
 
+Expression<bool> existsQuery(BaseSelectStatement select) {
```
simolus3, line 122: IMO these top-level functions should be declared in `exists.dart`.

In terms of documentation, we might have something like:

```suggestion
/// The `EXISTS` operator checks whether the [select] subquery returns any rows.
Expression<bool> existsQuery(BaseSelectStatement select) {
```

moor/lib/src/runtime/query_builder/expressions/expression.dart
```diff
@@ -119,6 +119,14 @@ abstract class Expression<D> implements FunctionParameter {
...
+  return _ExistsExpression(select, false);
+}
+
+Expression<bool> notExistsQuery(BaseSelectStatement select) {
```
simolus3, line 126:
```suggestion
/// The `NOT EXISTS` operator evaluates to `true` if the [select] subquery does
/// not return any rows.
Expression<bool> notExistsQuery(BaseSelectStatement select) {
```

simolus3 (commented):
Thank you for this! I agree that it's a useful application of subqueries in the Dart api.

>  I'm not sure if you want the function `existsQuery` as top level or if it should be an extension method. `(subquery).exists.`

I don't have strong opinions here either. Extension methods can have better discoverability, but exists kind of feels like a top-level predicate, not a method on queries. You can make them an extension if you want, but I'm fine with this approach too.

> I messed around with a very adhoc implementation of subqueries yesterday and I was able to expect it bound to true and not 1

Variables always map to simple values, so expecting it to be `1` seems correct to me.

davidmartos96:
@simolus3 Thanks for the recommendations! I applied the corrections.

simolus3:
Thanks! :+1:

PR #1056, opened 2021-02-17
Fix join when adding columns to select

davidmartos96:
I've encountered an issue when mixing select with addColumns and join. 

`Unsupported operation: Cannot add to an unmodifiable list`

The problem is the const list that is being created implicitly.


simolus3:
Thanks :+1:

PR #1112, opened 2021-03-21
Update to stable version of  sqflite_sqlcipher

davidmartos96:


extras/encryption/pubspec.yaml
```diff
@@ -7,8 +7,8 @@ environment:
   sdk: '>=2.12.0-0 <3.0.0'
 
 dependencies:
-  moor: ^4.0.0-nullsafety.0
-  sqflite_sqlcipher: ^2.0.0-nullsafety
+  moor: '>=4.0.0 <5.0.0'
+  sqflite_sqlcipher: '>=2.0.0 <3.0.0'
```
simolus3, line 11: I think those are equivalent to the following which IMO are easier to read

```suggestion
  moor: ^4.0.0
  sqflite_sqlcipher: ^2.0.0
```

simolus3 (commented):
Thanks! Let's just keep using the carret syntax here.

extras/encryption/pubspec.yaml
davidmartos96: Ok, sounds good!

simolus3:
Thanks!

Issue #1096, opened 2021-03-13
Support RETURNING in moor files


simolus3:
sqlite3 version 3.35.0, which was released yesterday, added support for `RETURNING` clauses. `RETURNING` clauses can be added to an `UPDATE`, `DELETE` or `INSERT` statement and allow evaluating expressions for affected rows:

```sql
CREATE TABLE t0(
  a INTEGER PRIMARY KEY,
  b DATE DEFAULT CURRENT_TIMESTAMP,
  c INTEGER
);

newT: INSERT INTO t0(c) VALUES(random()) RETURNING *;
```

Here, `newT` would return a single row return the rowid, the the current timestamp and the generated random value.

This is an awesome feature, let's support it in moor! Adding support for this requires a considerable effort across the analyzer, the generator and database implementations:

- [ ] Support this in `sqlparser`
- [ ] Support this in `moor_generator`
- [ ] Support this in sqflite-based implementations
- [ ] Support this in `moor/ffi.dart`
- [ ] Support this in the web implementation

Another interesting approach would be to add a version of `into(table).insert` that automatically returns the full data class by generating a `RETURNING` clause. It probably shouldn't be the default since it requires a super new sqlite3 version, but we could have an extension or a new method for that.

fgblomqvist:
Was there ever a version of `into(table).insert` added that will return the full object that was added?

simolus3:
Yes! You can use `into(table).`[`insertReturning`](https://pub.dev/documentation/moor/latest/moor/InsertStatement/insertReturning.html).

fgblomqvist:
Ah sweet!

kuhnroyal:
@simolus3 Is there a way to get this working on Desktop with `VmDatabase.memory()`?
I started using this but now it is failing in some test cases.

simolus3:
@kuhnroyal Is the error about `RETURNING` in general? sqlite 3.35 is very recent, so the environment where you run your tests on may not have it yet.
On Linux (and probably on macOS too) you could download sqlite3 somewhere and then run the tests with `LD_LIBRARY_PATH=/folder/containing/your/sqlite flutter test`.

kuhnroyal:
Yea the error is about `RETURNING` being unknown it seems. I am on latest MacOS 11.3.1. Will try a custom library path tomorrow. Thanks!

kuhnroyal:
That doesn't seem to work.
The lib here does not exist on my system:
```dart
  if (Platform.isMacOS) {
[...]

PR #1219, opened 2021-05-25
Fix insertReturning stream update

kuhnroyal:
* test case not possible due to sqlite version
* related to #1096

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,25 +9,35 @@ jobs:
   moor:
     name: "moor package"
     runs-on: ubuntu-20.04
-
+    defaults:
+      run:
+        working-directory: moor
     steps:
       # setup
       - uses: actions/checkout@v2
       - uses: cedx/setup-dart@v2
-      - run: sudo apt-get install -y libsqlite3-dev
-        name: Install sqlite3 for tests
+      - name: Install sqlite3 for tests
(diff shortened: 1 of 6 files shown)


simolus3:
Thanks for this 👍 Can we also add a test similar to this one? https://github.com/simolus3/moor/blob/21cf78a361d502bfba9b225aa9359af06be0461b/moor/test/insert_test.dart#L61-L70

kuhnroyal:
See description :)

kuhnroyal:
We could try to add a test suite that only runs in CI and maybe ensure the correct version there.

kuhnroyal:
I found the skips for this, refactored a little bit and added a test.

kuhnroyal:
Hmm, managed to build the sqlite version in CI but it is not being used.

simolus3:
Ah sorry, I should have read more carefully. The test I posted actually uses a mocked database implementation (mainly for historic reasons and so that it can reliably be tested on the web), so we may not even need the special setup.

kuhnroyal:
Yes but the test still fails with a syntax error :(
And the tests don't mean much if they get skipped in CI.

kuhnroyal:
Ok, CI now uses the manually built sqlite version and the test are all 3 running.

simolus3:
Thanks :+1:

PR #1238, opened 2021-05-31
Row class mapper for custom row classes

westito:


westito:
Maybe namings need clarifications

PR #1239, opened 2021-05-31
Custom row classes mapper 2

westito:
Maybe it's a better solution. All fields included in mapping that are used in constructor

### DIFF
--- a/moor/lib/src/dsl/table.dart
+++ b/moor/lib/src/dsl/table.dart
@@ -150,9 +150,16 @@ class UseRowClass {
   /// used to map database rows to the desired row class.
   final String constructor;
 
+  /// Enables reverse mapping generation
+  ///
+  /// When this option is set, moor generates custom row class to companion
+  /// class mapping
+  final bool generateReverseMapping;
+
   /// Customize the class used by moor to hold an instance of an annotated
   /// table.
   ///
   /// For details, see the overall documentation on [UseRowClass].
-  const UseRowClass(this.type, {this.constructor = ''});
+  const UseRowClass(this.type,
+      {this.constructor = '', this.generateReverseMapping = false});
 }
--- a/moor_generator/lib/src/writer/tables/update_companion_writer.dart
+++ b/moor_generator/lib/src/writer/tables/update_companion_writer.dart
@@ -25,7 +25,7 @@ class UpdateCompanionWriter {
 
     _writeCopyWith();
     _writeToColumnsOverride();
-    if (table.hasExistingRowClass) {
+    if (table.generateReverseMapping && table.hasExistingRowClass) {
       _writeFromRowClass();
     }
     _writeToString();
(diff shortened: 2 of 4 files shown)


westito:
Can be a builder option to turn on/off this mapper

simolus3:
Thanks for your contribution!

> Can be a builder option to turn on/off this mapper

Can we make it an optional boolean flag on the `@UseRowClass` constructor? Then users could turn it own on the data classes where they need it.

westito:
Done!

westito:
I fixed CI erros

moor/lib/src/dsl/table.dart
```diff
@@ -150,9 +150,16 @@ class UseRowClass {
   /// used to map database rows to the desired row class.
   final String constructor;
 
+  /// Enables reverse mapping generation
+  ///
+  /// When this option is set, moor generates custom row class to companion
+  /// class mapping
+  final bool generateReverseMapping;
```
simolus3, line 157: Not sure if the line length works out, but I think we should expand the documentation here:

```suggestion
  /// Generate a companion constructor mapping the referenced [type] to a companion.
  ///
  /// When this option is set (it defaults to `false`), moor generates a constructor in the
  /// matching companion class of this table mapping an instance of [type] to a suitable
  /// companion.
  /// This can be useful when a custom data class should be used for inserts or updates.
  /// For this purpose, make it implement the `Insertable` interface, enable 
  /// [generateToCompanion] and use the generated companion in your `toColumns`
  /// implementation.
  /// When using this field, make sure that your existing data class contains a getter 
  /// (or field) for each column in the database.
  final bool generateToCompanion;
```

moor_generator/lib/src/writer/tables/update_companion_writer.dart
```diff
@@ -200,6 +203,64 @@ class UpdateCompanionWriter {
...
+
+    _buffer.writeln(');\n}\n');
+
+    _buffer
+      ..write('static Map<String, Expression> ')
+      ..write('fromRowClassToColumns($dataClassName object,')
+      ..write('{bool nullToAbsent = false}) {\n')
+      ..write('return ${table.getNameForCompanionClass(scope.options)}.')
+      ..write('fromRowClass(object, nullToAbsent: nullToAbsent)')
+      ..write('.toColumns(nullToAbsent);');
```
simolus3, line 259: The overall mapping constructor looks very good, but do we really need this helper? I think users can just do that manually, we don't need to generate a utility function for each companion here.

simolus3 (commented):
I've left some comments, mainly regarding the user-visible api. The generation in the companion looks great!

Also, I'm wondering if we should verify that the class can be used as a companion? For instance, if it's missing a getter or has a getter with the wrong type, we'd generate invalid code which is hard to detect because users like to exclude generated files from analysis.
Maybe we can check that the class contains a getter with a matching type for each expected column (we do something similar when analyzing whether the constructor is eligible).

moor_generator/lib/src/writer/tables/update_companion_writer.dart
westito: It's not really necessary. I made for to set default (false) value to nullToAbsent parameter in toColumns() function. Maybe it's not clear to user, but 'nullToAbsent' have to be false in this case. The row class is probably mutable when user want use this mapping feature. It's an unexcpected operation when user set a nullable field to null (that had a value before), and moor absents and skip update that field after mapping. In fact, maybe we must fix nullToAbsent to false.

Issue #1132, opened 2021-04-09
Option to make all named parameters required?


PiN73:
First, thank you for `named_parameters` option! It makes using queries from .moor files more clear and safe (I mean change of parameters order won't break things). 

## Context

After [6aa022e(diff)](https://github.com/simolus3/moor/commit/6aa022e547b82155bb461fb44122f6764f2ae27b#diff-4df9a31c3706b4d1eda7f1a6a158dc09ca812e464d6089c974ed8f85fdc7dabb) nullable named parameters became non-required.

## Use case

I have a table with nullable column:
```sql
CREATE TABLE product (
    id INT NOT NULL PRIMARY KEY AUTOINCREMENT,
    name TEXT NOT NULL,
    category INT
);
```

and want to query items in any given category, including uncategorized (null):
```sql
getProducts:
    SELECT * FROM product
    WHERE
        CASE WHEN :category IS NULL THEN
            category IS NULL
        ELSE
            category = :category
        END;
```

Then the generated function will have signature:
```dart
Selectable<ProductData> getProducts({int? category})
```
which could potentially lead to mistake in complex queries with many parameters. One could forget to pass all parameters, and nothing will signal about it.

## Solution I'd like

It would be great to have a build option to make all named parameters required.

simolus3:
I see how that query could benefit from `category` being required, but I'm not sure if making all parameters required is the ideal solution here.

Maybe we could add an opt-in hint?

```sql
getProducts (REQUIRED :category):
    SELECT * FROM product
    WHERE
        CASE WHEN :category IS NULL THEN
            category IS NULL
        ELSE
            category = :category
        END;
```

The `REQUIRED` modifier would always make the Dart parameter required.

PiN73:
Thank you for the answer.

This hint may help too if one will declare all query variables in headers like
```
getProducts (REQUIRED :category AS INT OR NULL):
```

Possible limitations I see are

1. The fact that all used variables are declared in header isn't checked. This may lead to inconsistent code IMHO

2. Types will be typed by user and won't be checked, e.g.:

```
getProducts (REQUIRED :category AS TEXT OR NULL):
```
won't give warnings. More on this in https://github.com/simolus3/moor/discussions/1133

PiN73:
And another benefit from global option is that it's easy to implement and maintain IMHO: https://github.com/simolus3/moor/compare/develop...PiN73:named-required

simolus3:
[...]

PR #1242, opened 2021-06-01
Option to make all named parameters required

PiN73:
Use-case is described in #1132


moor_generator/lib/src/analyzer/options.dart
```diff
@@ -88,6 +88,9 @@ class MoorOptions {
   @JsonKey(name: 'named_parameters', defaultValue: false)
   final bool generateNamedParameters;
 
+  @JsonKey(name: 'named_parameters_always_required', defaultValue: false)
+  final bool namedParametersAlwaysRequired;
```
simolus3, line 92: Could you also add a quick description of this parameter to `docs/pages/docs/Advanced Features/builder_options.md`?

PiN73: Done

simolus3:
Thanks!

PR #1251, opened 2021-06-03
Add foreign key support

westito:
Need to write public api docs

### DIFF
--- a/moor/lib/src/dsl/columns.dart
+++ b/moor/lib/src/dsl/columns.dart
@@ -27,6 +27,9 @@ abstract class BlobColumn extends Column<Uint8List?> {}
 /// A column that stores floating point numeric values.
 abstract class RealColumn extends Column<double?> {}
 
+///
+enum KeyAction { noAction, cascade, restrict, setNull, setDefault }
+
 /// A column builder is used to specify which columns should appear in a table.
 /// All of the methods defined in this class and its subclasses are not meant to
 /// be called at runtime. Instead, moor_generator will take a look at your
@@ -164,6 +167,15 @@ class ColumnBuilder<Builder, ResultColumn extends Column<ResultDartType>,
   /// The generated row class will then use a `MyFancyClass` instead of a
   /// `String`, which would usually be used for [Table.text] columns.
   Builder map<T>(TypeConverter<T, ResultDartType> converter) => _isGenerated();
+
+  ///
+  Builder foreignKey(
+    Type references, {
+    KeyAction onUpdate = KeyAction.noAction,
+    KeyAction onDelete = KeyAction.noAction,
+    String column = 'id',
+  }) =>
+      _isGenerated();
 
   /// Turns this column builder into a column. This method won't actually be
   /// called in your code. Instead, moor_generator will take a look at your
--- a/moor_generator/lib/src/model/table.dart
+++ b/moor_generator/lib/src/model/table.dart
@@ -105,7 +105,7 @@ class MoorTable implements MoorEntityWithResultSet {
 
   /// When non-null, the generated table class will override the
   /// `customConstraints` getter in the table class with this value.
-  final List<String> overrideTableConstraints;
+  List<String> overrideTableConstraints;
 
   @override
   final Set<MoorTable> references = {};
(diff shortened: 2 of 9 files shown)


moor_generator/lib/src/model/table.dart
```diff
@@ -105,7 +105,7 @@ class MoorTable implements MoorEntityWithResultSet {
 
   /// When non-null, the generated table class will override the
   /// `customConstraints` getter in the table class with this value.
-  final List<String> overrideTableConstraints;
+  List<String> overrideTableConstraints;
```
simolus3, line 108: Why did this need to become mutable?

moor_generator/lib/src/writer/tables/table_writer.dart
```diff
@@ -365,7 +365,8 @@ class TableWriter extends TableOrViewWriter {
 
       buffer
         ..write('@override\n')
-        ..write('List<String> get customConstraints => const [$value];\n');
+        ..write('List<String> get customConstraints => super.customConstraints'
```
simolus3, line 368: Similar here, it looks like this change is not necessary?

moor/lib/src/dsl/columns.dart
```diff
@@ -165,6 +168,15 @@ class ColumnBuilder<Builder, ResultColumn extends Column<ResultDartType>,
   /// `String`, which would usually be used for [Table.text] columns.
   Builder map<T>(TypeConverter<T, ResultDartType> converter) => _isGenerated();
 
+  ///
+  Builder foreignKey(
+    Type references, {
+    KeyAction onUpdate = KeyAction.noAction,
+    KeyAction onDelete = KeyAction.noAction,
+    String column = 'id',
```
simolus3, line 176: I think this shouldn't have a default value. We could also make it a `Symbol`. The we could let users write something like

```dart
.foreignKey(Categories, #id, onUpdate: ...);
```

moor/lib/src/dsl/columns.dart
```diff
@@ -165,6 +168,15 @@ class ColumnBuilder<Builder, ResultColumn extends Column<ResultDartType>,
   /// `String`, which would usually be used for [Table.text] columns.
   Builder map<T>(TypeConverter<T, ResultDartType> converter) => _isGenerated();
 
+  ///
+  Builder foreignKey(
```
simolus3, line 172: Let's call this method `references` to reflect what it does in SQL.

simolus3 (commented):
Overall this looks pretty nice, thanks! Let me know if you need any help with the documentation.

moor_generator/lib/src/model/table.dart
westito: I didn't want to break the original functionality (where this field can be null if empty). But, foreign key function use this field after it is initialized (to null). Is it better to have a final growable list? That's mutable too :D

moor_generator/lib/src/writer/tables/table_writer.dart
westito: That's necessary because foreign key constraints override the custom constraints that you provide in the definition class. With super the custom constraints is merged with foreign key constraints.

moor/lib/src/dsl/columns.dart
westito: Symbol is a good advice. What parameters should'n be default? I think the referenced (primary key) column is "id" in the cases of 99%. That's why I made it default. Also, - for example - I never provide "onUpdate" in most cases, because i know id fields won't be changed. SQL implementation also not requires ON UPDATE/DELETE clauses in references definition (that means no action). But this is not made just for me :) If you say this fields should be required, I accept.

westito: But yes, it can be a nice declaration with two positional argument and two named. Looks like the resulting sql statement. :)

simolus3: `onUpdate` and `onDelete` have defaults in sqlite3 (`NO ACTION`), so it's reasonable to copy them too.

My problem with `id` being the default is that it might be confusing for users declaring tables with a different primary key. They might assume that, since they didn't have to declare the column in most other cases, moor was somehow smart enough to just pick the right column all of the time :) So far we're not, so it's best to just make it explicit.

westito:
I fix suggestions. Please help with documentation. I'm not so good at writing that, and my english isn't the best too 😄

westito:
Please pick up and continue. Thank you!

PR #1264, opened 2021-06-16
Change predicate from Expression<bool> to Expression<bool?>

North101:
```
listItems($predicate = TRUE):
SELECT *
FROM table
WHERE $predicate
```

will generate
```dart
Future<int> listItems({Expression<bool> predicate = const CustomExpression('(TRUE)')})
```

However all column expressions evaluate to Expression<bool?> which is incompatible.
I'm not sure if this is the correct solution as ExpressionDartPlaceholderType seems to be used elsewhere

### DIFF
--- a/moor_generator/lib/src/model/sql_query.dart
+++ b/moor_generator/lib/src/model/sql_query.dart
@@ -572,7 +572,7 @@ class ExpressionDartPlaceholderType extends DartPlaceholderType {
     if (columnType == null) return 'Expression';
 
     final dartType = dartTypeNames[columnType];
-    return 'Expression<$dartType>';
+    return 'Expression<${options.nullableType(dartType)}>';
   }
 }
 


moor_generator/lib/src/model/sql_query.dart
```diff
@@ -568,7 +568,7 @@ class ExpressionDartPlaceholderType extends DartPlaceholderType {
     if (columnType == null) return 'Expression';
 
     final dartType = dartTypeNames[columnType];
-    return 'Expression<$dartType>';
+    return 'Expression<$dartType?>';
```
simolus3, line 571: Moor must still support projects that haven't migrated to null-safety yet, so we can't unconditionally use null-safe syntax.

```suggestion
    // Use a nullable type because most columns are currently nullable in
    // the Dart api.
    return 'Expression<${options.nullableType(dartType)}>';
```

simolus3 (commented):
I agree that making the expression nullable is the way to go. It's not perfect, but with the current column nullability in Dart there's no other option really.
Also, could you change the base to `develop`? That makes it easier for me to combine this with the current state and release everything together, thanks!

moor_generator/lib/src/model/sql_query.dart
North101: > I agree that making the expression nullable is the way to go. It's not perfect, but with the current column nullability in Dart there's no other option really.

> Also, could you change the base to `develop`? That makes it easier for me to combine this with the current state and release everything together, thanks!
done

> Moor must still support projects that haven't migrated to null-safety yet, so we can't unconditionally use null-safe syntax.
done

simolus3:
Awesome, thanks!

PR #1289, opened 2021-06-26
Define batch callback as typedef (#1288)

### LINKED ISSUE #1288: Replace Function(Batch) with typedef for Batch callback
Currently, `batch` is declared as:

`Future<void> batch(Function(Batch) runInBatch) {}`

If we declare the `Function` via a `typedef`, then it ensures that the analyzer suggestion knows to autofill the `batch` argument rather than inserting an empty function callback, as this can be confusing at first for first-time Moor users.

https://user-images.githubusercontent.com/6836556/123514266-712bb500-d6d5-11eb-893c-91afcbb83bfb.mp4

jamie1192:
Resolves #1288 

https://user-images.githubusercontent.com/6836556/123514493-a553a580-d6d6-11eb-91a9-4cbd6ad001f2.mp4

### DIFF
--- a/moor/lib/src/runtime/api/connection_user.dart
+++ b/moor/lib/src/runtime/api/connection_user.dart
@@ -5,6 +5,8 @@ const _zoneRootUserKey = #DatabaseConnectionUser;
 typedef _CustomWriter<T> = Future<T> Function(
     QueryExecutor e, String sql, List<dynamic> vars);
 
+typedef _BatchRunner = dynamic Function(Batch batch);
+
 /// Manages a [DatabaseConnection] to send queries to the database.
 abstract class DatabaseConnectionUser {
   /// The database connection used by this [DatabaseConnectionUser].
@@ -440,7 +442,7 @@ abstract class DatabaseConnectionUser {
   ///    );
   ///  });
   /// ```
-  Future<void> batch(Function(Batch) runInBatch) {
+  Future<void> batch(_BatchRunner runInBatch) {
     final engine = _resolvedEngine;
 
     final batch = Batch._(engine, engine is! Transaction);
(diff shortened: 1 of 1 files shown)


moor/lib/src/runtime/api/connection_user.dart
```diff
@@ -5,6 +5,8 @@ const _zoneRootUserKey = #DatabaseConnectionUser;
 typedef _CustomWriter<T> = Future<T> Function(
     QueryExecutor e, String sql, List<dynamic> vars);
 
+typedef _BatchRunner = dynamic Function(Batch batch);
```
simolus3, line 8: Thanks for your contribution :+1: 

Since we're not using the result of the function, let's also change the return type to void:

```suggestion
typedef _BatchRunner = void Function(Batch batch);
```

jamie1192: ah cool, shall do!

jamie1192: ah now I remember why I left it as `dynamic`, [the `if` statement](https://github.com/simolus3/moor/blob/519e8328c8f6170a6683bd66b5374f9ecafff219/moor/lib/src/runtime/api/connection_user.dart#L451) here checks the result of `batch`.

Is it fine to remove this check?

davidmartos96: What about returning FutureOr<void> and always do an await instead of the if statement?

simolus3: Ah true, I've missed that. Making it return `FutureOr<void>` should work, with or without the changes to the if statement.

jamie1192: Thanks guys, change has been made.

simolus3:
Thank you :+1:

Issue #1241, opened 2021-06-01
Moor should handle nulls before they get to TypeConverter


North101:
I think Moor should handle nulls before they get to TypeConverter.

```
enum DeckGroup {
  Name,
  Cycle,
  Pack,
  Side,
  Faction,
  Created,
  Updated,
  Synced,
}
```

```
class DeckGroupConverter extends TypeConverter<DeckGroup, String> {
  const DeckGroupConverter();

  @override
  DeckGroup? mapToDart(String? value) {
    if (value == null) return null;

    return DeckGroup.values.firstWhereOrNull((e) => e.toString() == value);
  }

  @override
  String? mapToSql(DeckGroup? value) {
    if (value == null) return null;

    return value.toString();
  }
}
```

```
CREATE TABLE settings(
  deck_group TEXT MAPPED BY `const DeckGroupConverter()` NOT NULL
);
```

Take the above example When I try to access settings.deckGroup (which is not nullable) its actually possible to get a null. Obviously I can easily fix this by providing a default, but I think a better solution is to to never pass nulls to TypeConverter e.g.

```
class DeckGroupConverter extends TypeConverter<DeckGroup, String> {
  const DeckGroupConverter();

  @override
  DeckGroup mapToDart(String value) {
    return DeckGroup.values.firstWhereOrNull((e) => e.toString() == value) ?? DeckGroup.default;
  }

  @override
  String mapToSql(DeckGroup value) {
    return value.toString();
  }
}
```

westito:
+1
I also thoght about it

simolus3:
I agree too! The current design is mainly a leftover from pre-nnbd to avoid making it too breaking. Unfortunately that also means that we need a moor 5 for this.

In the generator it will be a bit more complicated to handle when we have converters mapping non-nullable values to nullable values (or vice versa, or if both are nullable to begin with). The behavior here isn't obvious, I'll think about it some more. The non-nullable to non-nullable case, which is probably the most common, should be easy to support though.

simolus3:
I've fixed this in 06af930941407e16b3a0f2f1bdeb5b7d9038fcb2. Essentially, the change is to always enable `null_aware_type_converters`. Together with `NullAwareTypeConverter`, I think this is quite neat to use:

[...]

PR #1331, opened 2021-07-08
Enforce nnbd for typeconverters

North101:
This just adds a NnbdTypeConverter helper class to prevent non-null columns from returning null (feel free to suggest class or method names).

#1241

### DIFF
--- a/moor/lib/src/runtime/types/custom_type.dart
+++ b/moor/lib/src/runtime/types/custom_type.dart
@@ -39,3 +39,34 @@ class EnumIndexConverter<T> extends TypeConverter<T, int> {
     return (value as dynamic)?.index as int;
   }
 }
+
+/// A type converter automatically mapping `null` values to `null` in both
+/// directions.
+///
+/// Instead of overriding  [mapToDart] and [mapToSql], subclasses of this
+/// converter should implement [requireMapToDart] and [requireMapToSql], which
+/// are used to map non-null values to and from sql values, respectively.
+///
+/// Apart from the implementation changes, subclasses of this converter can be
+/// used just like all other type converters.
+abstract class NullAwareTypeConverter<D, S> extends TypeConverter<D, S> {
+  /// Constant default constructor.
+  const NullAwareTypeConverter();
+
+  @override
+  D? mapToDart(S? fromDb) {
+    return fromDb == null ? null : requireMapToDart(fromDb);
+  }
+
+  /// Map a non-null value from an object in Dart into something that will be
+  /// understood by the database.
+  D requireMapToDart(S fromDb);
+
+  @override
+  S? mapToSql(D? value) {
+    return value == null ? null : requireMapToSql(value);
+  }
+
+  /// Maps a non-null column from the database back to Dart.
+  S requireMapToSql(D value);
+}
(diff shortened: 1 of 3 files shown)


moor/lib/src/runtime/types/custom_type.dart
```diff
@@ -39,3 +39,30 @@ class EnumIndexConverter<T> extends TypeConverter<T, int> {
     return (value as dynamic)?.index as int;
   }
 }
+
+/// Enforces nnbd for TypeConverters
```
simolus3, line 43: I think this is a little generic, how about something like

```suggestion
/// A type converter automatically mapping `null` values to `null` in both directions.
///
/// Instead of overriding  [mapToDart] and [mapToSql], subclasses of this converter
/// should implement [requireMapToDart] and [requireMapToSql], which are used to
/// map non-null values to and from sql values, respectively.
///
/// Apart from the implementation changes, subclasses of this converter can be used
/// just like all other type converters.
```

(you may have to wrap some lines differently)

moor/lib/src/runtime/types/custom_type.dart
```diff
@@ -39,3 +39,30 @@ class EnumIndexConverter<T> extends TypeConverter<T, int> {
...
+abstract class NnbdTypeConverter<D, S> extends TypeConverter<D, S> {
```
simolus3, line 44: I'm not sure about the abbreviated "by default" in the name here, maybe something like `NonNullableTypeConverter`? I don't really like that either, I think the name should point out that nulls are "passed through". I don't have a better idea though :D

simolus3 (commented):
Thanks for this, I think it'll be useful for most type converters in practice :+1: 

Can we have tests for this? We can have an independent test where this class is extended, a test would verify that nulls are passed through when calling `mapToDart` and `mapToSql`. You can put this in `moor/test/types/non_nullable_type_converter_test.dart`

moor/lib/src/runtime/types/custom_type.dart
North101: I also struggled with what to name it (and keep it concise). I forgot that bd meant "by default" so it's probably the wrong name for it. Maybe NullEnforcedTypeConverter? Though that might be confusing to users

simolus3: To me, `NullEnforcedTypeConverter` sounds like `null` values are not allowed to enter the converter at all. Maybe `NullAwareTypeConverter`?

North101: Sounds good to me

North101:
Done I think. I wasn't sure what you wanted from the tests so let me know if you want anything changed

simolus3:
Looks good, thanks!

westito:
```
/// Map a non-null value from an object in Dart into something that will be
/// understood by the database.
D requireMapToDart(S fromDb);

/// Maps a non-null column from the database back to Dart.
S requireMapToSql(D value);
```
I think the two comment is swapped.

simolus3:
Good catch, fixed in d3f757235fb4fcb5f548d0152e32e761c071931a.

PR #1343, opened 2021-07-13
Correctly override Object.== everywhere

kuhnroyal:
Using dynamic results in invalid mocks being generated by mockito since `==(dynamic other)` can not be overridden by `==(Object other)`.


kuhnroyal:
The analyzer errors are now gone even without this change, not sure what my problem was actually.
In any case, this seems to be the correct override in post Dart 2.12

simolus3:
Thanks!

PR #1354, opened 2021-07-17
Mark _ensureOpenCalled before migrations.

Elvis5566:
It looks like we should mark _ensureOpenCalled before migrations, otherwise, it's not possible to run queries during migrations?


simolus3:
You can still run queries by using the higher-level methods on the database class instead of the underlying `executor`. They will delegate to a special executor setup just for migrations.

But I think there's no harm in changing this, so thanks :+1:

PR #1357, opened 2021-07-18
doc: version is a string

boumenot:
Quote the version to convey String instead of double.


simolus3:
Good catch, thanks! The failing CI run is due to a pub problem, we can ignore that for now.

PR #1388, opened 2021-07-29
Prevent NullAwareTypeConverter dart value from being null

North101:
It is currently possible to pass null types to `NullAwareTypeConverter`
```
class MyTypeConverter extends NullAwareTypeConverter<String, String?> {
  String requireMapToDart(String? value);
  String? requireMapToSql(String value);
}
```

Because `NullAwareTypeConverter` handles nulls, a null will never get to `requireMapToDart` and the user may incorrectly try to handle that themselves.

This PR makes the dart value in `NullAwareTypeConverter` extend `Object` preventing it from being nullable and enforcing that behaviour.


simolus3:
This is technically a breaking change but the practical impact is probably very minimal. I'm fine with it, thanks for your contribution :+1:

North101:
I wasn't sure if it was merged into master yet. But thought it was a good change anyway

PR #1431, opened 2021-08-21
Skip default value for required params

### LINKED ISSUE #1430: named_parameters_always_required generates broken code
build.yaml
```yaml
...
named_parameters: true
named_parameters_always_required: true
```

tables.moor
```sql
...
query:
    SELECT *
    FROM todos
    ORDER BY $order;
```

generates
```dart
Selectable<Todo> query({required OrderBy order = const OrderBy.nothing()})
```
which causes compilation error _(Required named parameters can't have a default value)_

Tested on moor 4.5.0, full example: https://github.com/PiN73/test_moor_order

PiN73:
Fixes #1430


simolus3:
Thanks!

PR #1473, opened 2021-10-05
implements where for DoUpdate

### LINKED ISSUE #1472: where statement with onConflict update
i would like to make a query like that
```sql
CREATE TABLE words(word TEXT PRIMARY KEY, occurrences INT);
INSERT INTO words(word,occurrences) VALUES('jovial',1)
  ON CONFLICT(word) DO UPDATE SET occurrences=occurrences+1 where occurrences <6
```
i could almost make it work with 
```dart
     await into(words).insert(
       WordsCompanion.insert(word: word, occurrences: 1),
       onConflict: DoUpdate((old) => WordsCompanion.custom(
         occurrences: old.occurrences + Constant(1),
       )),
     );
```
but i can't add a where statement to it `where occurrences <6`

i suggest to add new parameter to `DoUpdate` class that take a `Where` class or `Expression`
like this
```dart
 onConflict: DoUpdate((old) => WordsCompanion.custom(
         occurrences: old.occurrences + Constant(1),
       ) , where : (old)=>old.occurrences.isSmallerThanValue(5) )
```

i am happy to make a pull request if my approach ok or any other approach you see better

Ali1Ammar:
implements where for `DoUpdate`
also adding test and integration_tests for it
fix #1472

### DIFF
--- a/moor/lib/src/runtime/query_builder/statements/insert.dart
+++ b/moor/lib/src/runtime/query_builder/statements/insert.dart
@@ -197,6 +197,13 @@ class InsertStatement<T extends Table, D> {
 
         first = false;
       }
+
+      if (onConflict._where != null) {
+        ctx.writeWhitespace();
+        final where = onConflict._where!(
+            table.asDslTable, table.createAlias('excluded').asDslTable);
+        where.writeInto(ctx);
+      }
     }
 
     if (onConflict is DoUpdate<T, D>) {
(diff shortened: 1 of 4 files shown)


moor/lib/src/runtime/query_builder/statements/insert.dart
```diff
@@ -317,8 +326,10 @@ class DoUpdate<T extends Table, D> extends UpsertClause<T, D> {
   /// been inserted, use [DoUpdate.withExcluded].
   ///
```
simolus3, line 327: Let's document this new feature, maybe something like "The optional `where` clause can be used to disable the update based on the old value. If a `where` clause is set and it evaluates to `false`, a conflict will keep the old row without applying the update".

simolus3 (commented):
This is looking very good, thanks for adding this feature and the tests :rocket: 

I think we should expand the documentation on the constructor to mention the `where` clause. Also, can you add an entry to the changelog (something like "Add optional `where` clause to `DoUpdate` constructors")?

simolus3:
Thanks again!

PR #1492, opened 2021-10-17
Add 'toCompanion' generator for custom row classes

westito:
Adds 'toCompanion' extension function generator for custom row classes. This avoids writing boilerplate code for mapping custom row classes to companions. You can control generation with 'generate_to_companion_extension' builder option (defaults to false, of course)

### DIFF
--- a/drift/lib/src/dsl/table.dart
+++ b/drift/lib/src/dsl/table.dart
@@ -155,23 +155,17 @@ class UseRowClass {
   /// used to map database rows to the desired row class.
   final String constructor;
 
-  /// Generate a companion constructor mapping the referenced [type] to a
-  /// companion.
+  /// Generate a `toInsertable()` extension function for [type] mapping all
+  /// fields to an insertable object.
   ///
-  /// When this option is set (it defaults to `false`), drift generates a
-  /// constructor in the matching companion class of this table mapping an
-  /// instance of [type] to a suitable companion.
   /// This can be useful when a custom data class should be used for inserts or
   /// updates.
-  /// For this purpose, make it implement the `Insertable` interface, enable
-  /// [generateToCompanion] and use the generated companion in your `toColumns`
-  /// implementation.
-  final bool generateToCompanion;
+  final bool generateInsertable;
 
   /// Customize the class used by drift to hold an instance of an annotated
   /// table.
   ///
   /// For details, see the overall documentation on [UseRowClass].
   const UseRowClass(this.type,
-      {this.constructor = '', this.generateToCompanion = false});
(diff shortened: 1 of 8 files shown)


westito:
Yes, it would be a better solution. I had a similar PR in the past that wasn't merged. There we talked about turn it on/off in builder option. 
I'll change builder option to @UseRowClass then, but the parameter name will be 'generateReverseMapping' as you suggested last time.

westito:
Updated, tests passes!

westito:
Fix analyze errors

westito:
I found a solution: User can use custom getter to match columns. So builder will throw error and give a suggestion for that

westito:
Using 'generateToCompanion' assume user wants to use the row class to all operations, without companions (like ORM). Select object -> modify fields -> save (the the whole object). In this way all fields are included in the update query. When user calls 'toCompanion' there is always a complete data object behind where all fields are set and null means null, never means absent. 
There is two type of user I think: 
1. Who use companions (the regular way). He always writes updates with companion classes
2. Who use row classes only (as I) and never use companions. 
I think there is no user (or very few) who wants to map all fields to companion and modify before insert. In this case, there is a chance to make unexpected errors with  'nullToAbsent'.

I think better solution to replace 'toCompanion' with 'toInsertable'. It can be used without boilerplate 'extends' and 'toColumns' override in custom row class. To achieve 'before save' event there will be an 'OnInsert' interface. If row class implements it, `void onInsert(CompanionClass companion)` function will be called when drift calls 'toColumns' on insert. 
Would be better name 'beforeInsert' or 'beforeSave'. 
Or simply leave 'toCompanion' too, but I think this is "dangerous" a bit. 

I know you don't want to make an ORM from Drift, that's why I'm asking you what you think about this.


Sample:
```
abstract class OnInsert {
  void onInsert(TableWithoutPKCompanion companion);
}

class _$CustomRowClass implements Insertable<CustomRowClass> {
  CustomRowClass object;

  _$CustomRowClass(this.object);

  @override
  Map<String, Expression> toColumns(bool nullToAbsent) {
    final companion = TableWithoutPKCompanion(
      notReallyAnId: Value(object.notReallyAnId),
      someFloat: Value(object.someFloat),
      custom: Value(object.custom),
    );

    if (object is OnInsert) {
      (object as OnInsert).onInsert(companion);
    }

    return companion.toColumns(false);
  }
}

extension CustomRowClassToCompanion on CustomRowClass {
  _$CustomRowClass toInsertable() {
    return _$CustomRowClass(this);
  }
}
```

westito:
Done. I also modified parameter name to 'generateInsertable'

PR #1507, opened 2021-10-22
PostgreSQL support

westito:
PostgresSQL experimental support

All tests pass! (except migration) 🎉

Missing features:
<s>- Migrations, versioning (in progress)</s>

Modifications:
<s>- SQL type names changed for Postgres (SQLite compatible)</s>


drift/lib/src/runtime/query_builder/schema/column_impl.dart
```diff
@@ -93,10 +93,20 @@ class GeneratedColumn<T> extends Column<T> {
   /// [here](https://www.sqlite.org/syntax/column-def.html), into the given
   /// buffer.
   void writeColumnDefinition(GenerationContext into) {
-    into.buffer.write('$escapedName $typeName');
+    final isSerial = into.dialect == SqlDialect.postgres &&
+        type is IntType &&
+        _defaultConstraints == 'PRIMARY KEY AUTOINCREMENT';
```
simolus3, line 98: Checking for the generated default constraints feels a bit hacky to me. Maybe we could add a field like `bool isPrimaryKeyWithAutoIncrement = false` field to the column and have the generator set that to true when needed?

drift/lib/src/runtime/query_builder/statements/insert.dart
```diff
@@ -214,6 +228,9 @@ class InsertStatement<T extends Table, D> {
 
     if (returning) {
       ctx.buffer.write(' RETURNING *');
+    } else if (ctx.dialect == SqlDialect.postgres) {
+      //TODO: handle multiple primary key
+      ctx.buffer.write(' RETURNING ${table.$primaryKey.first.$name}');
```
simolus3, line 233: If you don't want to add support for composite primary keys in this PR (which is fine with me), we should at least throw an exception explaining that they're not currently supported. Silently doing the wrong thing is not a good strategy :D

extras/drift_postgres/lib/src/pg_database.dart
```diff
@@ -0,0 +1,202 @@
...
+
+///
+class PgDatabase extends DelegatedDatabase {
+  ///
+  PgDatabase(PostgreSQLConnection connection)
+      : super(_PgDelegate(connection, connection),
+            isSequential: true, logStatements: true);
+
+  ///
+  factory PgDatabase.open(String host, int port, String databaseName,
```
simolus3, line 12: I'm not sure if we need this constructor. We'll just always have to catch up with the API from `package:postgres` and the convenience for users appears to be small.

extras/drift_postgres/lib/src/pg_database.dart
```diff
@@ -0,0 +1,202 @@
...
+        password: password,
+        timeoutInSeconds: timeoutInSeconds,
+        queryTimeoutInSeconds: queryTimeoutInSeconds,
+        timeZone: timeZone,
+        useSSL: useSSL,
+        isUnixSocket: isUnixSocket));
+  }
+}
+
+///
```
simolus3, line 31: We don't need documentation comments for private classes :D But I can also fill out the docs later.

extras/drift_postgres/lib/src/pg_database.dart
```diff
@@ -0,0 +1,202 @@
...
+  Future _ensureOpen() async {
+    if (_db.isClosed) {
+      await _db.open();
+    }
+  }
+
+  Future<void> _initializeDatabase() async {
+    // TODO: Do we need create these functions?
+    //setup?.call(_db);
+  }
```
simolus3, line 75: I think this method can indeed be removed :D

extras/drift_postgres/lib/src/pg_database.dart
```diff
@@ -0,0 +1,202 @@
...
+
+  String _convertStatement(String statement) {
+    final argMap = {};
+    return statement
+        .replaceAllMapped(_regexIndexed, (match) => '@${match[1]}')
+        .replaceAllMapped(_regexNamed, (match) {
+      final index = argMap.putIfAbsent(match[1], () => argMap.length + 1);
+      return '@$index';
+    });
+  }
```
simolus3, line 88: I don't like that we have to do this mapping, it would break if someone writes something like `':sunglasses:'` in a string literal. We can change how variables are written at runtime, right? Is this as a guard for compiled queries?

extras/integration_tests/tests/lib/database/database.dart
```diff
@@ -69,25 +69,25 @@ class PreferenceConverter extends TypeConverter<Preferences, String> {
...
     'friendshipsOf': ''' SELECT 
           f.really_good_friends, user.**
        FROM friendships f
-         INNER JOIN users user ON user.id IN (f.first_user, f.second_user) AND
-             user.id != :user
+         INNER JOIN users "user" ON "user".id IN (f.first_user, f.second_user) 
+         AND "user".id != :user
        WHERE (f.first_user = :user OR f.second_user = :user)''',
-    'userCount': 'SELECT COUNT(id) FROM users',
+    'userCount': 'SELECT COUNT(id) count FROM users',
```
simolus3, line 87: For these (and similar) changes, I think you can also enable the `new_sql_code_generation` [builder option](https://drift.simonbinder.eu/docs/advanced-features/builder_options/#recommended-options) in the `integration_tests/tests` package. It will make the generator add the explicit column name for you and we don't need to manually change the SQL here.

simolus3 (commented):
Sorry for the slow update, and thanks again for all your work :heart: I have left some comments, but most of them are fairly minor.

drift/lib/src/runtime/query_builder/statements/insert.dart
westito: Returning query isn't really necessary. This emulates SQLite's `last_insert_rowid` what is returns integer primary key. If there is no integer primary key or there is multiple key, SQLite return rowId. If table created `WITHOUT ROWID` then it returns zero. Follow this way, we could return zero if there is multiple keys or the primary key is not integer (non-integer primary key also not handled in this code)

extras/drift_postgres/lib/src/pg_database.dart
westito: It is a method for connect without specify postgres dependency in pubspec. But, I can remove if you think it isn't good

extras/drift_postgres/lib/src/pg_database.dart
westito: It is a way to not to rewrite all queries and related code in Drift and keep backward compatibility. SQLite understand @ and ? too, but Postgres don't accept '?'. Also Postgres not understands named parameters. I rewrited the related code before, but reverted and I made this replace script. Would be a big change to use '@' and remove named parameters. `:sunglasses:` isn't acceptable by SQLite either (https://www.sqlite.org/c3ref/bind_blob.html). Only alphanumeric characters accepted. `:` at the end throw error. Also, the regex not match it, so won't be replaced to `@sunglasses@`. I think this script is the best we have now. Maintains named parameters' order and keeps backward compatibility for current code.

drift/lib/src/runtime/query_builder/schema/column_impl.dart
westito: Drift also use this hacky way ;)

```dart
/// Whether this column has an `AUTOINCREMENT` primary key constraint that was
  /// created by drift.
  bool get hasAutoIncrement =>
      _defaultConstraints?.contains('AUTOINCREMENT') == true;
```

westito:
We have a small problem with 64bit integers. SQLite supports 64bit (In fact, integer size can't be set), therefore I use bigint/bigserial types in Postgres dialect. The problem is, Dart web can't handle integers this size (53bit only). It would be necessary to warn the user in the documentation. Other question is, what if Postgres user want to use 32bit (or 8bit) integers only (for reduce storage size)? May we can add an IntSize enum (default Int64) parameter into integer() column definition function.
I made a research about web 64bit int support. SqlJS supports bigint by a setting, but it is maps Sql.bigint to JS BigInt (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) type what is not supported by Dart web right now. Dart says this JS BigInt not supported all browser right now and anyways, Dart has its own 'BigInt' class that is platform-independent. "Use that if you want real 64bit on web". It is not an option for native JS libs.
https://github.com/dart-lang/sdk/issues/32399
https://github.com/dart-lang/sdk/issues/45092
There is a workaround, maybe that helps. I give it a try later.

extras/drift_postgres/lib/src/pg_database.dart
westito: There is a 'useNativeFunctions()' call in here with native SQLite that creates custom functions. This functions is required? `regexp_moor_ffi, moor_contains, sqrt, etc.`

extras/integration_tests/tests/lib/database/database.dart
westito: Okay, I reverted the database file and enable new sql generator, but I had to change it: `user` is reserved keyword in Postgres, so it is necessary to enclose it in quotation marks. Second problem is, boolean treated as integer right now, and postgres can't accept use as boolean without equal operator
```sql
SELECT COUNT(*) FROM friendships f
WHERE f.really_good_friends
AND (f.first_user = :user OR f.second_user = :user)
```
This sql works with SQLite because it can threat integer as boolean (`f.really_good_friends` can be used without ` = 1`). But Postgres can't handle that, so it throws `argument of AND must be type boolean, not type integer` error. 
Custom query parser don't know that `really_good_friends` is boolean or not. 

1. solution is leave it as is and users have to update existing SQL queries (breaks backward compatible)
2. solution is modify table generation and query builder code to handle booleans differently. Latter is much more work

drift/lib/src/runtime/query_builder/schema/column_impl.dart
simolus3: Touché, putting that into a getter is fine then :)

drift/lib/src/runtime/query_builder/statements/insert.dart
simolus3: Fair enough, let's return an invalid value (say `-1`) if we hit an unexpected case. We should also document this current limitation somewhere, but I can take care of that.

extras/drift_postgres/lib/src/pg_database.dart
simolus3: Yeah I think I would prefer users to import `postgres` too, they are implicitly using its API.

extras/drift_postgres/lib/src/pg_database.dart
simolus3: It can be removed for postgres since these SQL functions are only available with a `NativeDatabase` (something that we explain in the docs too). We can't register custom functions on postgres so it should be fine.

drift/lib/src/runtime/query_builder/statements/insert.dart
westito: Having non-integer or multi key primary doesn't seem unexpected. I should return 0 as SQLite does.

extras/drift_postgres/lib/src/pg_database.dart
simolus3: Sorry, I was being unclear. My point is that the following query does not contain any variables: `SELECT ':sunglasses:'` (it returns a string with the content `:sunglasses:`). After your transformation, we would run `SELECT '?'` which returns a different value!

We can change the `Variable` class to write postgres-compatible parameters. Generated code will be a challenge, of course - I'll take a look and see what can be done about this.

westito: Yes you sure. I didn't thought about it. I already made a full rewrite from '?' to '@' and it's worked like a charm with SQLite and Postgres too. But I reverted for backward compatibility. Can we make a major version change and migrate users to @ (or $)? As I see, SQL parsed in moor files and in Driftdatabase annotation. There It can be automatically replace `?` to `@` in generated code. This change would affect only custom queries.

westito:
- I made a smarter stmt parameter replacer for now. '?' placeholder could be easily changed, but it would break the custom sql queries for current users. If we change '?' to '@' or '$' this placeholder replacer can be removed.
- Fixed primary key returning (returns 0 if primaryKey.count != 1 || not integer as SQLite does)

sqlparser/lib/src/reader/tokenizer/token.dart
```diff
@@ -29,14 +29,20 @@ enum TokenType {
   ampersand,
   analyze,
   and,
+  any,
```
simolus3, line 32: We should probably make these postgres-specific (or at least the `keyword` mapping below). I can take care of adding the necessary logic in the tokenizer since it will probably be a bit ugly. But are we using these tokens anywhere?

extras/drift_postgres/lib/src/pg_database.dart
```diff
@@ -0,0 +1,191 @@
...
+    if (_db.isClosed) {
+      await _db.open();
+    }
+  }
+
+  final _regexIndexed = RegExp(r'\?(\d+)');
+  final _regexNamed = RegExp(r':(\w+)');
+  final _escapeChar = String.fromCharCode(0x0);
+
+  String _convertStatement(String statement) {
```
simolus3, line 56: FWIW, we can also have this depend on `sqlparser` directly and use the tokenizer from that package to identify variable tokens.

Even without adding new token types to that package, you can do `Scanner(statement).scanTokens()` and then replace `QuestionMarkVariableToken`s with `@` tokens. Replacing could work by looking at the token's `span`, which reports the region of the string to replace.

sqlparser/lib/src/reader/tokenizer/token.dart
westito: These only for quoting. Drift use quotes when one of this keywords are used. I don't think it's need to be separated for Postgres.

extras/drift_postgres/lib/src/pg_database.dart
westito: Thank you for suggestion. I can't know and use the parser yet :) This would impact query performance, I think placeholder should be replaced anyways (in long term). May we can use a config for placeholder mark?
In deed, this question mark problem affects only custom queries. Drift could choose mark depending on dialect in generated queries (one of my older code did this). For SQLite nothing change, custom queries will run as before. However, if somebody want to use Postgres, he have to use or change '?' marks to '@'/'$'. SQLite will understand both.

sqlparser/lib/src/reader/tokenizer/token.dart
simolus3: >  I don't think it's need to be separated for Postgres.

The thing is that tests are now failing because the parser is overly strict and always interprets keywords as keywords. So doing `SELECT any FROM table` would be valid in sqlite3 (selects a column called "any"), but the parser now complains that  "any" is an unexpected keyword here.

The list of keywords used for escaping at runtime is available [here](https://github.com/simolus3/moor/blob/develop/drift/lib/sqlite_keywords.dart), that one can be changed without issues :D

extras/drift_postgres/lib/src/pg_database.dart
simolus3: Tokenization only should be reasonably fast (I hope it's comparable to RegExps, otherwise I'll have to take a look at the tokenizer to see where it's going wrong), but indeed there is going to be some overhead.

> However, if somebody want to use Postgres, he have to use or change '?' marks to '@'/'$'. SQLite will understand both.

I think this is fine initially. The SQL parser used to analyze queries is heavily focused on sqlite, so it will take a while to have good support for other dialects there. We have a build option that rewrites queries at compile-time, so we could have another dialect option to rewrite variables too. We don't need to do this in this PR though, let's just rewrite variables at runtime for now (either with the tokenizer or with regexps)

westito: I think tokenizer is faster than regex, I compared overhead to without rewriting. But, I think better to add a runtime option to rewrite or not. For ex, I want to use Postgres and I only use '@' because of that. In this case tokenizer runs for nothing unnecessarily. This parameter mark issue not affects builder, only runtime. I suppose to add a 'compatible mode' option to Postgres connection/PgDatabase. When it is true, mark stay '?' everywhere as currently and runs tokenizer. If false, parameter mark depends on dialect in generated queries.

westito:
I escaped reserved keywords in sqlparser tests. Now it works

simolus3:
> I escaped reserved keywords in sqlparser tests. Now it works

Unfortunately, end users are going to run into the same scenario with their queries. I really think we shouldn't add postgres-specific keywords to the default configuration `sqlparser`. I think the keywords from `sqlparser` are not used in any other part of drift really, so are they really necessary? There's another file (`package:drift/sqlite_keywords.dart`) responsible for escaping at runtime, can we change that one instead?

extras/drift_postgres/lib/src/pg_database.dart
simolus3: I like that idea 👍

westito:
Keywords is necessary for escaping. If user use "user" as column name, Drift won't escape it in generated queries and Postgres queries will fail. If I see well the issue

simolus3:
> Keywords is necessary for escaping. If user use "user" as column name, Drift won't escape it in generated queries and Postgres queries will fail.

Oh, now I get it. I think we could either encourage postgres users to (temporarily) disable `new_sql_code_generation` until we sort this out or we could add a [build option](https://drift.simonbinder.eu/docs/advanced-features/builder_options/) to specify the dialect used. The later will be considerably more complicated, but I still think that we shouldn't add unnecessary keywords for sqlite3 users.

westito:
I am also on the side of maintaining backward compatibility as much as possible. We could add a `postgres_compatible` builder option and later many other thing can be switched with too.  

Oh I see the main problem:
"Inside of named queries, you can use variables just like you would expect with sql. We support regular variables (?), explicitly indexed variables (?123) and colon-named variables (:id). We don't support variables declared with @ or $. "
We can't use @ because it is a special identifier in Drift files.

Then, we can leave then '?' and ':' in Drift files. 'postgres_compatible' mode also enables `new_sql_code_generation`. Sqlparser parses custom SQLs with ':/?', but writes '@' as parameter sign in rewrited queries.

westito:
```dart
/// If [name] is a reserved sql keyword, wraps it in double ticks. Otherwise
/// just returns the [name] directly.
String get escapedName {
    return isKeywordLexeme(name) ? '"$name"' : name;
}
```
Won't be easier to escape all names regardless of it is keyword or not?

westito:
I started to make a compatible mode, but it requires to rewrite drift and drift_dev  in many places. Must update all "escapeIfNeed" call. 
I think adding Postgres keywords only in runtime drift will not cause problem. It is only used for construct runtime queries, not affects custom queries.

westito:
Okay. Current solution:
- There is a "compatible mode" which with Sqlparser looks for postgres keywords too at escaping.
- Drift use '@' in generated queries as parameter mark. It is not affects the custom queries itself. '?' mark must still be used in Drift files 
- I added postgres keywords to runtime drift. It is only affect runtime-generated queries (invisible to use). Custom constraints / queries won't be affected either. Current SQLite queries with postgres keywords will work as before (for ex: `REFERENCES(user)`). On Postgres this will fail obviously, but user have to refactor his queries if he want to use the Postgres extension.

PlugFox:
Hello, when i can try alpha version of postgres support?
I'm at an early stage of my project with backend on the dart, it could be useful for me and you)

westito:
> Hello, when i can try alpha version of postgres support?
> I'm at an early stage of my project with backend on the dart, it could be useful for me and you)

Sure!

Add these lines to pubspec.yaml

```yaml
dependencies:
  drift: any
  drift_postgres: any

dev_dependencies:
  drift_dev: any

dependency_overrides:
  drift:
    git:
      url: https://github.com/westito/drift.git
      ref: pg_support
      path: drift

  drift_postgres:
    git:
      url: https://github.com/westito/drift.git
      ref: pg_support
      path: extras/drift_postgres

  drift_dev:
    git:
      url: https://github.com/westito/drift.git
      ref: pg_support
      path: drift_dev

  sqlparser:
    git:
      url: https://github.com/westito/drift.git
      ref: pg_support
      path: sqlparser
```

PlugFox:
> Sure!

OK, thanks.
If I have any problems, I will let you know)

westito:
Almost forget! Add these builder options too:

```yaml
targets:
  $default:
    builders:
      drift_dev:
        options:
          new_sql_code_generation: true
          compatible_mode_generation: true
```

simolus3:
@westito I like your idea with the build option for compatibility :+1: . Is this ready from your side now? If so I'll take a deeper look tomorrow.

westito:
I thought about move '$expandVar' to executor too 😄 Seemed like a big change.
Is compatible mode is good name or would be changed to for ex. "generic_sql_dialect_generation" or similar?

westito:
I used "compatible mode" naming (could be "generic_sql"), because dialect builder option fixes that dialect in generated code. For ex. in my project the same generated code used on mobile and on backend too. The generated code must be dialect-independent.

simolus3:
> because dialect builder option fixes that dialect in generated code. For ex. in my project the same generated code used on mobile and on backend too

I understand where you're coming from, but that will be a hard trade off to make in the long term. If we want excellent support for both sqlite3 and postgres in drift, we need to know the exact dialect at compile-time for the accurate analysis and code generation.

In your setup (which sounds super awesome btw, I :heart: Dart on the backend), I assume you have separate packages for your app and the backend, right? Because then you could introduce a shared package that stores the `.drift` files. You'd then have a database class in the backend and another one in the app package, they'd both import the shared drift files. However, you can choose different builder options for the two packages to reflect that they target different database engines. Would that work for you?

westito:
I thinking on a solution, but I have a separate [core] package that contains all the data handling (DAO, Repos, stc.) and mobile and backend depend on that. Mobile/web/backend only gives the connection (Pg, Worker or NativeDatabase). If I generate two different .g.dart files, I would need write a lot of interfaces to decide between. This does make no sense. However, SQLite understands Postgres queries (at least we currently use), so database generated for postgres will work for SQLite too. Also, this dialect builder option only affects Drift files and custom queries. If not use these, there is only runtime generated queries that is depends on connection dialect.

simolus3:
> This does make no sense

Yeah I just had the same problem applying that to the integration tests :D

> However, SQLite understands Postgres queries (at least we currently use), so database generated for postgres will work for SQLite too

Yeah, just setting `dialect: postgres` should work for sqlite3 as well so the build option is probably fine. I still think it's the better approach in the long-term, maybe we need support generating code for multiple dialects in the future.

simolus3:
This looks good to me now :tada: . Thanks a lot for all your effort here, I really appreciate it!!

westito:
I'm glad I could help 😊🥳

davidmartos96:
@westito Awesome job! I will give it a go with a shelf server app. I was using some adhoc query helpers until now 😄

PR #1541, opened 2021-11-12
Dsl views

westito:
Only syntax code for discuss

```dart
class TodoCategoryItemCount extends View {
  late $TodoItemsTable todoItems;
  late TodoCategories todoCategories;

  IntColumn get itemCount => integer().generatedAs(todoItems.id.count())();

  @override
  Query as() =>
      select([todoCategories.name, itemCount]).from(todoCategories).join([
        innerJoin(todoItems, todoItems.categoryId.equalsExp(todoCategories.id))
      ]);
}
```


drift/example/main.dart
```diff
@@ -5,13 +5,54 @@ import 'package:drift/native.dart';
...
 class TodoItems extends Table {
   IntColumn get id => integer().autoIncrement()();
   TextColumn get title => text()();
   TextColumn get content => text().nullable()();
+  IntColumn get categoryId => integer().references(TodoCategories, #id)();
+}
+
+class TodoCategoryItemCount extends View {
+  late $TodoItemsTable todoItems;
+  late TodoCategories todoCategories;
```
simolus3, line 22: I'm not entirely sold on the `late` syntax here.

Could we make it so that a user makes a view class abstract, writes `TodoCategories get todoCategories;` and lets drift generate an implementation of the view class (which would then inherit `ViewInfo`) that overrides the getters.

simolus3 (commented):
Thanks for starting this :+1:

drift/example/main.dart
westito: Yeah. That's a good point.

westito:
Dsl view generator beta version :) Tests need 

Generates CREATE VIEW from dsl at runtime. 

```dart
@DriftView(name: 'customViewName')
abstract class TodoItemWithCategoryNameView extends View {
  $TodoItemsTable get todoItems;
  $TodoCategoriesTable get todoCategories;

  TextColumn get title => text().generatedAs(todoItems.title +
      const Constant('(') +
      todoCategories.name +
      const Constant(')'))();

  @override
  Query as() => select([todoItems.id, title]).from(todoItems).join([
        innerJoin(
            todoCategories, todoCategories.id.equalsExp(todoItems.categoryId),
            useColumns: false)
      ]);
}
```

drift_dev/lib/src/analyzer/dart/view_parser.dart
```diff
@@ -0,0 +1,237 @@
...
+  }
+
+  Future<String?> _getStaticReference(
+      FieldElement field, List<MoorTable> tables) async {
+    if (field.getter != null) {
+      try {
+        final node = await base.loadElementDeclaration(field.getter!);
+        if (node is MethodDeclaration && node.body is EmptyFunctionBody) {
+          final type = tables.firstWhereOrNull(
+              (tbl) => tbl.entityInfoName == node.returnType.toString());
```
simolus3, line 150: I think it would be awesome if users didn't have to use the name of the generated class. We should definitely support that still (for tables defined in drift files, for instance), but it would be great if they could use the Dart type too:

```dart
abstract class TodoCategoryItemCount extends View {
  TodoItems get todoItems;
  TodoCategories get todoCategories;
```

You can check if the table was declared in Dart and then use `tbl.fromClass` to check if that matches.

westito: That was my first thought, but you can't use this reference type. `.join([innerJoin(todoItems,..)])` in `as()` shows error because it isn't a `ResultSetImplemetation`. May `innerJoin()` can be modified to accept `Table`

simolus3: Oh, you're right. I think 47f0465795080944c48d135155bba05cafb67cf3 is not a good idea either (it'll break when joining views, for instance).

We could let users do something like this:

```dart
abstract class TodoCategoryItemCount extends View {
  @TodoItems
  TableInfo get todoItems;
```

But I'm not sure if that's more ergonomic. I'll keep pondering.

westito: I think this would be too boilerplate. Join parameter can be 'HasResultSet' that accepts views too, and not breaks current functionality.

westito: Other solution is separate view declaration from query and dsl view don't create a database level view. That was my first idea. Views are projections actually, queries isn't declared in DslView, have to be write like table queries. There is no performance gain of store view in database. Also, not migration friendly.

westito:
Can we make some progress on this? It is done from my side. Can you check the PR?

drift/example/main.dart
```diff
@@ -5,18 +5,67 @@ import 'package:drift/native.dart';
 
 part 'main.g.dart';
 
+class TodoCategories extends Table {
```
simolus3, line 8:
```suggestion
@DataClassName('TodoCategory')
class TodoCategories extends Table {
```

drift/lib/src/runtime/query_builder/schema/view_info.dart
```diff
@@ -0,0 +1,25 @@
...
+///
+/// For more information on views, see the [CREATE VIEW][sqlite-docs]
+/// documentation from sqlite, or the [entry on sqlitetutorial.net][sql-tut].
+///
+/// [sqlite-docs]: https://www.sqlite.org/lang_createview.html
+/// [sql-tut]: https://www.sqlitetutorial.net/sqlite-create-view/
+abstract class ViewInfo<Self extends HasResultSet, Row>
+    implements ResultSetImplementation<Self, Row> {
+  ///
+  String get actualViewName;
```
simolus3, line 15: IMO we can just drop this and use `entityName`. For the table implementation, `actualTableName` only exists for backwards compatibility, we don't need to add a redundant getter for views too.

drift/example/main.dart
```diff
@@ -5,18 +5,67 @@ import 'package:drift/native.dart';
...
   IntColumn get id => integer().autoIncrement()();
   TextColumn get title => text()();
   TextColumn get content => text().nullable()();
+  IntColumn get categoryId => integer().references(TodoCategories, #id)();
+
+  TextColumn get generatedText => text().nullable().generatedAs(
+      title + const Constant(' (') + content + const Constant(')'))();
+}
+
+abstract class TodoCategoryItemCount extends View {
```
simolus3, line 23: Could you add views to `drift/test/data/tables/todos.dart` as well? I would like to see a test ensuring that the `CREATE VIEW` statement matches what we expect (probably in `schema_test.dart`).

drift/example/main.dart
```diff
@@ -5,18 +5,67 @@ import 'package:drift/native.dart';
...
+      const Constant('(') +
+      todoCategories.name +
+      const Constant(')'))();
+
+  @override
+  Query as() => select([todoItems.id, title]).from(todoItems).join([
+        innerJoin(
+            todoCategories, todoCategories.id.equalsExp(todoItems.categoryId),
+            useColumns: false)
+      ]);
```
simolus3, line 54: I don't fully understand the need for having both column getters and a `as()` method defining the query. Of course, we need to know the overall schema of the view at compile-time, so there's some use for column getters. But to me it feels like giving them a `generatedAs` might conflict with the `Query as()` method.

Maybe there's a way to only define columns and their types with getters and then define all of the view in a single method for consistency?
Maybe something like

```dart
abstract class MyView extends View {
  // define injected tables that we reference
  TableA get a;
  TableB get b;

  // define columns generated by this view
  TextColumn get firstColumn;
  TextColumn get secondColumn;

  // Have a special API for defining the query
  ViewQuery as(DatabaseConnectionUser db) {
    return ViewQuery(
       bindColumns: [firstColumn, secondColumn], // for explicit ordering
       query: db.select(...), // The columns returned by this query are bound to the columns defined by this view
    );
  }
}
```

That way,

- we can just use the unmodified `as` method in the migrator, passing the actual database instance
- we don't need to analyze the `as` method at compile time at all, simplifying the generation process
- we still have all the information necessary to extract the view's schema for compile-time analysis
- IMO, we can make it a runtime error if `bindColumns` does not contain all columns, or if `query` returns the wrong amount of columns.

Sorry for delivering feedback this fundamental so late. I hope an approach like this could actually simplify the view parser and generator in `drift_dev` a bit. Knowing how to define views in SQL, defining expressions for view columns in the select statement feels a bit more intuitive to me.

westito: My current implementation use Drift file parser logic and easily integrates to the current Drift logic. The declared views are create DB level views (by CREATE VIEW). The suggested "query only" implementation can't achieve that. Also, this binding solution can cause unexpected behaviours: If the underlying query changes, the view binding can change without the user's knowledge. On the other hand this binding syntax make user lazy. Can query a lot of columns (if use `select()` and not `selectOnly()..addColumns()`) but read only some. Lost all view benefits. 

Migration is not an aspect. On schema change simply drop and recreate all views. 

I can suggest a modify to simplify column declaration:
```dart 
Expression get title => todoItems.title +
      const Constant('(') +
      todoCategories.name +
      const Constant(')'));

   @override
  Query as() => select([todoItems.id, title]).from(todoItems).join([
        innerJoin(
            todoCategories, todoCategories.id.equalsExp(todoItems.categoryId),
            useColumns: false)
      ]);
```

simolus3: > I can suggest a modify to simplify column declaration:

Oh yes, I like that simplified syntax!

westito:
Okay, I modified to expression syntax. Works like before, but not need to declare column.
Issue (I think not mine): Virtual table test failed committed by boyan

simolus3:
Thanks! Could you also add a small "integration" test for the generator? Just add a Dart view declaration to `drift/test/data/tables` somewhere and ensure that the view gets created as it should in `drift/test/schema_test.dart` (let me know if I can help you with this in case anything is unclear). Apart from that this looks ready to me, I can take care of documentation.

westito:
- Fixed a major bug (table reference field names was ignored)
- Add views to schema test

simolus3 (commented):
Thanks for the update! I have started with the documentation and left some (minor) final comments.

drift/lib/src/runtime/query_builder/schema/entities.dart
```diff
@@ -47,28 +47,6 @@ class Index extends DatabaseSchemaEntity {
...
-/// A sqlite view.
-///
-/// In drift, views can only be declared in `.drift` files.
-///
-/// For more information on views, see the [CREATE VIEW][sqlite-docs]
-/// documentation from sqlite, or the [entry on sqlitetutorial.net][sql-tut].
-///
-/// [sqlite-docs]: https://www.sqlite.org/lang_createview.html
-/// [sql-tut]: https://www.sqlitetutorial.net/sqlite-create-view/
-abstract class View<Self, Row> extends ResultSetImplementation<Self, Row>
```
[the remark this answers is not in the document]
westito: This is necessary. It should have been called `ViewInfo` originally as tables do. Because of caret versioning may generated code breaks on current users without regenerate. Can we rename current `View` to (for ex.) `DslView`, but I think the current `View` name would be better to understand and users will regenerate the code if see compiler fails on there. Not a big deal.

westito:
I reverted your ViewInfo modification because compiler fails. `View` can't be a bound of generated view, because don't implement it. It is unnecessary to implement (and shouldn't).
I modified cancellation test (maybe it is not working properly now). It is fails most of the times on my machine. I think it is time depend. Must be reviewed.

simolus3:
> I reverted your ViewInfo modification because compiler fails

You're right, for some reason I didn't see that in the analyzer. I'll also take a look at the flakiness in the stream query cancellation test.

And a big thanks for all your work on this, this looks all good now :tada:

PR #1555, opened 2021-11-21
Convert ByteData before writing into file

munjata:
ByteData can not be written directly into file using writeAsBytes method.
It must be converted to Uint8List before the operation.


simolus3:
Thanks. Btw, please base your changes on `develop` instead of `master` :)

munjata:
Sure I will do that .
Thanks lot for this amazing package.

PR #1557, opened 2021-11-23
Update community_tools.md

NicolaVerbeeck:
Updated viewer and inspector packages to include drift specific versions


simolus3:
Thanks a lot, both for the PR and for renaming your projects!! From my own experience I know how annoying this is, but it's for the best :)

PR #1559, opened 2021-11-24
Add null aware type converter option

westito:
This option let TypeConverter to determinate Dart type nullability.

When you use @UseRowClass without enable 'null_aware_type_converters' this code throws `Expected this parameter to be nullable` error. Using nullable column Drift expects nullable dart type in row class, even if TypeConverter returns non-nullable type.

If you let Drift generates DataClass (default) the generated field will be nullable even if TypeConverter returns non-nullable type. 

This helps wrapping nullable columns in non-nullable type

```dart
TextColumn get customer => text()
    .named('customerId')
    .nullable()
    .references(PartnerTable, #id)
    .map(const RefConv<Partner?>())();

DataClass {
    final Ref<Partner?> customer;
}

class RefConv<T extends BaseModel?> extends TypeConverter<Ref<T>, String> {
  const RefConv();

  @override
  Ref<T>? mapToDart(String? fromDb) {
    return Ref<T>.id(fromDb);
  }

  @override
  String? mapToSql(Ref<T>? value) {
    return value?.id;
  }
}
```

### DIFF
--- a/docs/pages/docs/Advanced Features/builder_options.md
+++ b/docs/pages/docs/Advanced Features/builder_options.md
@@ -80,6 +80,7 @@ At the moment, drift supports these options:
 * `scoped_dart_components`: Generates a function parameter for [Dart placeholders]({{ '../Using SQL/moor_files.md#dart-components-in-sql' | pageUrl }}) in SQL.
   The function has a parameter for each table that is available in the query, making it easier to get aliases right when using
   Dart placeholders.
+* `null_aware_type_converters`: Consider the type of applied type converters to determine nullability of columns in Dart.
 
 ## Assumed SQL environment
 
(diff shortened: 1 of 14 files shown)


simolus3:
I wonder if we should even make this opt-in, as opposed to just enabling it by default. Removing an error case isn't breaking anyone, right?

westito:
It is breaks current functionality. For ex. in test `SyncTypeConverter` has a non-nullable type that's why Drift generates non-nullable fields and breaks current code where syncState can be null.

simolus3:
Of course, my bad :facepalm: This looks good to me then, thanks for the contribution :+1: I have opened #1561 as a (breaking) idea to improve the situation with non-nullable converters generating null values.

Issue #1561, opened 2021-11-25
Null-safety and type converters


simolus3:
For type converters, `mapToSql` and `mapToDart` use the nullable variant of the types used to declare the converter. This can be a bit confusing, especially for the most common case where those are both non-nullable.

In a breaking release, the generator should enforce a reasonable nullability of type converters:

- `mapToSql` and `mapToDart` are changed to only accept the matching type parameters. To declare a nullable type converter, users would have to extend `TypeConverter<String?, String?>`.
- Report a generation error when a type-converter with a nullable SQL type is used on a non-nullable column
- Report a generation error when a type-converter with a nullable Dart type is used  with a non-nullable field on an existing Dart data class.

westito:
I think the current `NullAwareTypeConverter` do this job right. It needs a small modification (`S extends Object` -> `S extends Object?`), but it can handle nullable and non-null types too. Type nullability depends on type provided in the generics.
My opinion is removing `NullAwareTypeConverter` and modify `TypeConverter` to work like the NullAware would solve this issue. Also have to remove `null_aware_type_converters` option and make the underlining logic permanent.

simolus3:
Yes! Considering the example from your PR, there are cases where null in SQL is not mapped to null in Dart. I think we should be able to express things like this by default (so make `null_aware_type_converters` permanent).

However, `NullAwareTypeConverter` is for mapping `null` to `null` and non-null values to non-null values. This is also why we have `S extends Object`, `requireMapToSql` really should return a non-nullable value. I mostly want to be able to reason about how a converter is dealing with nulls at compile time.

- If it's a `TypeConverter<T, S?>` with a nullable `S?`, we know that `mapToSql` might return null, so we can report a warning if the column is not marked as nullable in the database schema
- If it's a `TypeConverter<T?, S>` with a nullable `T`, we can warn if the target is non-nullable in Dart.

We're doing the second thing with `null_aware_type_converters`, but not the first one. Reliably introducing the first warning and making `null_aware_type_converters` the default requires type converters not "lying" about them returning null. So

```dart
// bad: We don't know how this converter deals with nulls internally
abstract class TypeConverter<D, S> {
  S? mapToSql(D? value);
  D? mapToDart(S? fromDb);
}

// better: we can reason about nullability at compile-time, it's going to be declared with
// nullable types if it wants to return null
abstract class TypeConverter2<D, S> {
  S mapToSql(D value);
  D mapToDart(S fromDb);
}
```

[...]

PR #1565, opened 2021-11-26
Rewrite TypeConverter to null aware by default

westito:
DSL table declaration is type safe by default. You can't use nullable (sql) type converter on a not null column, but you can use non-nullable (sql) type converter on null column. With `UseRowClass` the generator will display warning when you use nullable converter on non-nullable field. Generated data class is always match the type converter (dart) type.

For Drift files there is a warning when you want to use nullable TypeConverter on not null column. Vice versa is working. You can use a nullable TypeConverter on a not null column.

```
line 24, column 33: The column declared as not null, but TypeConverter SQL type argument is nullable
   ╷
24 │     sync_state INTEGER not null MAPPED BY `const NullableSyncTypeConverter()`,
   │                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```

### DIFF
--- a/drift/lib/src/runtime/query_builder/schema/column_impl.dart
+++ b/drift/lib/src/runtime/query_builder/schema/column_impl.dart
@@ -266,8 +266,8 @@ class GeneratedColumnWithTypeConverter<D, S> extends GeneratedColumn<S> {
   /// Compares this column against the mapped [dartValue].
   ///
   /// The value will be mapped using the [converter] applied to this column.
-  Expression<bool> equalsValue(D? dartValue) {
-    return equals(converter.mapToSql(dartValue) as S);
+  Expression<bool> equalsValue(D dartValue) {
+    return equals(converter.mapToSql(dartValue));
   }
 }
 
--- a/drift/lib/src/runtime/types/custom_type.dart
+++ b/drift/lib/src/runtime/types/custom_type.dart
@@ -14,15 +14,15 @@ abstract class TypeConverter<D, S> {
 
   /// Map a value from an object in Dart into something that will be understood
   /// by the database.
-  S? mapToSql(D? value);
+  S mapToSql(D value);
 
   /// Maps a column from the database back to Dart.
-  D? mapToDart(S? fromDb);
+  D mapToDart(S fromDb);
 }
 
 /// Implementation for an enum to int converter that uses the index of the enum
 /// as the value stored in the database.
-class EnumIndexConverter<T> extends NullAwareTypeConverter<T, int> {
+class EnumIndexConverter<T> extends TypeConverter<T, int> {
   /// All values of the enum.
   final List<T> values;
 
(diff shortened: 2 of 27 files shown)


westito:
#1561

simolus3:
Thanks for working on this! I should note that this is a breaking change, so we'd have to do a Drift 2.0 for this. I'll try to get the current features on `develop` and DSL views into a new version first, then we can bundle up all breaking change ideas afterwards.

westito:
The current develop branch will be released as 2.0?

simolus3:
I would prefer to release `develop` (after merging your views feature) as drift 1.1.

If there are enough breaking changes bundled up, I'll start merging this and apply the other breaking changes too. Then we can release that as drift 2.0

North101:
Does it make sense to combine JsonTypeConverter into TypeConverter?

simolus3:
Thanks for the contribution! I've applied a similar change inspired by your work in 06af930941407e16b3a0f2f1bdeb5b7d9038fcb2.

westito:
I tried to do a TypeConverter that can handle nullable and non-nullable fields with a single converter.

```dart
class Table {
  IntColumn get defaultLanguage =>
      integer().nullable().map(const EnumConv<Language?>(Language.values))();

  IntColumn get defaultLanguage2 =>
      integer().map(const EnumConv<Language>(Language.values))();
}
```

Version 1:
```dart
class EnumConv<T extends BaseEnum?> extends TypeConverter<T, int?> {
  final List<T> _values;

  const EnumConv(this._values);

  @override
  T fromSql(int? fromDb) {
    return (fromDb == null ? null : _values[fromDb]) as T;
  }

  @override
  int? toSql(T? value) {
    return value?.index;
  }
}
```

Version 2:
```dart
class EnumConv2<T extends BaseEnum?> extends NullAwareTypeConverter<T, int> {
  final List<T> _values;

  const EnumConv2(this._values);

  @override
  T requireFromSql(int? fromDb) {
    return _values[fromDb!];
  }

  @override
  int requireToSql(T value) {
    return value!.index;
  }
}
```

Version 1 works as expected, and there is no syntax errors.
```dart
class GeneratedModel {
  final Language? defaultLanguage;
  final Language defaultLanguage2;
}
```

But I want to simplify and eliminate that null check I thing NullAware should does this. But when I regenerate the code with version 2, the generated class has wrong null type:
```dart
class GeneratedModel {
  final Language? defaultLanguage;
  final Language? defaultLanguage2;
}
```

The problem is NullAware converter always converts `D` type to `D?`
```dart
abstract class NullAwareTypeConverter<D, S extends Object>
    extends TypeConverter<D?, S?>
```
 That's why the generator generates data class property as nullable even it should be non-null. 

The solution is remove `?` mark from `D?` in extends and add a type cast to `D` like in my converter above. The type cast will always be succeed because `requireFromSql` return type matches nullability of `D`

PR #1567, opened 2021-11-28
Add random ordering 

Husssam12:
I just add `OrderingTerm.random()` to `OrderingTerm`  class  to get a number of random rows using sqlite random function.

https://www.sqlitetutorial.net/sqlite-functions/sqlite-random/
https://stackoverflow.com/questions/2279706/select-random-row-from-a-sqlite-table

### DIFF
--- a/drift/lib/src/runtime/query_builder/components/order_by.dart
+++ b/drift/lib/src/runtime/query_builder/components/order_by.dart
@@ -8,7 +8,7 @@ enum OrderingMode {
   /// Descending ordering mode (highest items first)
   desc,
 
-  /// Random ordering mode
+  /// Descending ordering mode (highest items first)
   random,
 }
 
--- a/drift/lib/src/runtime/query_builder/components/order_by.dart
+++ b/drift/lib/src/runtime/query_builder/components/order_by.dart
@@ -54,9 +54,9 @@ class OrderingTerm extends Component {
   void writeInto(GenerationContext context) {
     if (mode != OrderingMode.random) {
       _expression!.writeInto(context);
+      context.writeWhitespace();
     }
     context.buffer.write(_modeToString[mode]);
-    context.writeWhitespace();
   }
 }
 
--- a/drift/lib/src/runtime/query_builder/components/order_by.dart
+++ b/drift/lib/src/runtime/query_builder/components/order_by.dart
@@ -6,26 +6,33 @@ enum OrderingMode {
   asc,
 
   /// Descending ordering mode (highest items first)
-  desc
+  desc,
+
+  /// Random ordering mode
+  random,
 }
 
 const _modeToString = {
   OrderingMode.asc: 'ASC',
   OrderingMode.desc: 'DESC',
+  OrderingMode.random: 'RANDOM()',
 };
 
 /// A single term in a [OrderBy] clause. The priority of this term is determined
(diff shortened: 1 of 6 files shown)


drift/pubspec.yaml
```diff
@@ -11,7 +11,7 @@ environment:
 dependencies:
   async: ^2.5.0
   convert: ^3.0.0
-  collection: ^1.15.0
+  collection: ^1.15.0K
```
westito, line 14: I think there is a misstype

drift/lib/src/runtime/query_builder/components/order_by.dart
```diff
@@ -37,10 +44,18 @@ class OrderingTerm extends Component {
     return OrderingTerm(expression: expression, mode: OrderingMode.desc);
   }
 
+  /// Creates an ordering term  to get a number of random rows
+  /// using sqlite random function.
+  OrderingTerm.random()
+      : _expression = null,
+        mode = OrderingMode.random;
```
westito, line 51: Create factory method instead of custom constructor (other modes is factories too)

drift/lib/src/runtime/query_builder/components/order_by.dart
```diff
@@ -6,26 +6,33 @@ enum OrderingMode {
...
 
   /// The ordering mode (ascending or descending).
   final OrderingMode mode;
 
   /// Creates an ordering term by the [expression] and the [mode] (defaults to
   /// ascending).
-  OrderingTerm({required this.expression, this.mode = OrderingMode.asc});
+  ///
+  /// When [mode] is [OrderingMode.random], the [expression] will be ignored
+  OrderingTerm({required Expression expression, this.mode = OrderingMode.asc})
```
westito, line 34: Change expression to nullable to make factory method work

simolus3:
Thanks for your contribution! I think we don't even have to make the expression nullable. You can just set it to a `FunctionCallExpression('random', [])` to make it work. We also wouldn't have to introduce a new value to the enum, so we simplify things a bit

drift/pubspec.yaml
Husssam12: Sorry 🥲

Husssam12:
> Thanks for your contribution! I think we don't even have to make the expression nullable. You can just set it to a `FunctionCallExpression('random', [])` to make it work. We also wouldn't have to introduce a new value to the enum, so we simplify things a bit 

Thank you for replying 
I should have opened an issue before this pull request 😅

Husssam12:
> Thanks for your contribution! I think we don't even have to make the expression nullable. You can just set it to a `FunctionCallExpression('random', [])` to make it work:
> 
> ```dart
> factory OrderingTerm.random() {
>   return OrderingTerm(expression: FunctionCallExpression('random', []));
> }
> ```
> 
> Then we can also
> 
> - Remove the new case in `OrderingMode`
> - make `expression` public and non-nullable again
> - remove the special handling in `writeInto`
> 
> Quite the simplification win for a simple change :)


Done ✅

simolus3:
>  I should have opened an issue before this pull request :sweat_smile:

No worries, this is fine! And congrats on your first PR, thanks for for your contribution!

simolus3 (approved):

PR #1568, opened 2021-11-28
Check references only on Database

westito:
Using `@DriftAccessor(tables: [Companies])` Drift throws error when tables not included in the list that provided table references. This isn't necessary for Dao
```
 This class has not been added as a table
   ╷
28 │       .references(Addresses, #id)
   │                   ^^^^^^^^^
   ╵

```


simolus3:
Thanks!

PR #1569, opened 2021-11-28
Fix virtual table test

westito:
After boyan's new `query virtual tables with unknown function` test case the `supports virtual tables across moor files` case went broken. Import missing from SQL code. Maybe his import optimiser deleted the import from string literal too.


simolus3:
Thanks. And I really should include `drift_dev` in the CI, so far it has always been to slow but it shouldn't be a problem with the test runner now using incremental compilation.

Issue #1600, opened 2021-12-26
Error message when trying to read blob from database


yanivshaked:
While trying to read a blob field from a database which was previously created, getting the following error message:

```
I/flutter (19432): type 'String' is not a subtype of type 'Uint8List?' in type cast, Stack: #0      BlobType.mapFromDatabaseResponse (package:drift/src/runtime/types/sql_types.dart:171:21)
```

Looking at `sql_types.dart`, the following code looks suspicious:
```
  @override
  Uint8List? mapFromDatabaseResponse(dynamic response) {
    return response as Uint8List?;
  }
```
This is not the correct way to convert a String to Uint8List. The above code should be:
```
  @override
  Uint8List? mapFromDatabaseResponse(dynamic response) {
    var list = response.codeUnits;
    return Uint8List.fromList(list);
  }
```

I will test the above and submit a PR if it works.

simolus3:
Can you share a bit more details on how to reproduce this? The `response` should really be a `Uint8List` already. If it's not, that looks like a bug in the database implementation.

yanivshaked:
Sure. I'm using `DB Browser for SQLite` to create a database. Following is the table creation command:
```
CREATE TABLE "Cocktails" (
	"Id"	INTEGER NOT NULL UNIQUE,
	"Description"	TEXT,
	"Test1"	BLOB,
	"Test2"	BLOB,
	"Test3"	BLOB,
	"Test4"	BLOB,
	PRIMARY KEY("Id" AUTOINCREMENT)
);
```
Error message is received for BLOB fields (All other fields are read correctly without any issue).
Please see my suggestion for a more robust approach in the PR #1601 .

PR #1601, opened 2021-12-26
Fix support for BlobType

yanivshaked:
This PR fixes issue #1600

### DIFF
--- a/drift/lib/src/runtime/types/sql_types.dart
+++ b/drift/lib/src/runtime/types/sql_types.dart
@@ -169,7 +169,7 @@ class BlobType extends SqlType<Uint8List> {
   @override
   Uint8List? mapFromDatabaseResponse(dynamic response) {
     if (response is String) {
-      var list = response.codeUnits;
+      final list = response.codeUnits;
       return Uint8List.fromList(list);
     }
     return response as Uint8List?;
--- a/drift/lib/src/runtime/types/sql_types.dart
+++ b/drift/lib/src/runtime/types/sql_types.dart
@@ -168,8 +168,11 @@ class BlobType extends SqlType<Uint8List> {
 
   @override
   Uint8List? mapFromDatabaseResponse(dynamic response) {
-    var list = response.codeUnits;
-    return Uint8List.fromList(list);
+    if (response is String) {
+      var list = response.codeUnits;
+      return Uint8List.fromList(list);
+    }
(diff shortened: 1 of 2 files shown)


drift/lib/src/runtime/types/sql_types.dart
```diff
@@ -168,7 +168,8 @@ class BlobType extends SqlType<Uint8List> {
 
   @override
   Uint8List? mapFromDatabaseResponse(dynamic response) {
-    return response as Uint8List?;
+    var list = response.codeUnits;
+    return Uint8List.fromList(list);
```
simolus3, line 172: `response` should really be a `Uint8List` already, not a string. Now this method only works when given a string, but we can actually store blobs in the database.

yanivshaked: Correct. To make it more robust, how about:
```
  @override
  Uint8List? mapFromDatabaseResponse(dynamic response) {
    if (response is String) {
      var list = response.codeUnits;
      return Uint8List.fromList(list);
    }
    return response as Uint8List?;
  }
```

drift/lib/src/runtime/types/sql_types.dart
```diff
@@ -168,6 +168,10 @@ class BlobType extends SqlType<Uint8List> {
 
   @override
   Uint8List? mapFromDatabaseResponse(dynamic response) {
+    if (response is String) {
```
simolus3, line 171: Could you also add a test verifying this behavior [here](https://github.com/simolus3/moor/blob/0a2cc40ba5bb0aa78c6192c479ddd1b22bef8ad3/drift/test/types/blob_test.dart#L6)? Just test that `mapFromDatabaseResponse` with a string returns an expected list of bytes.

yanivshaked: Sure, will do so later this evening

yanivshaked: Done, test added

simolus3:
Looks good, thanks!

yanivshaked:
@simolus3 Thanks, how soon can we expect a version release?

simolus3:
I have released this in 1.2.0. Thanks for your contribution!

yanivshaked:
My pleasure, thank you!!

PR #1609, opened 2021-12-31
Reproduction of the Boolean field issue

yanivshaked:
Boolean fields are not read correctly from the database.
I have written a test which reproduces this problem. (Problem is not fixed, but the test shows it)

### DIFF
--- a/drift/lib/src/runtime/types/sql_types.dart
+++ b/drift/lib/src/runtime/types/sql_types.dart
@@ -45,6 +45,9 @@ class BoolType extends SqlType<bool> {
   bool? mapFromDatabaseResponse(dynamic response) {
     // ignore: avoid_returning_null
     if (response == null) return null;
+    if (response is String) {
+      return response.toUpperCase().compareTo('TRUE') == 0;
+    }
     return response != 0;
   }
 
--- a/drift/test/types/bool_test.dart
+++ b/drift/test/types/bool_test.dart
@@ -2,18 +2,17 @@ import 'package:drift/drift.dart';
 import 'package:test/test.dart';
 
 void main() {
-  test('maps without transformation', () {
+  test('bool with underlying string', () {
     const type = BoolType();
-    const data = false;
 
-    expect(type.mapToSqlVariable(data), 0);
-    expect(type.mapFromDatabaseResponse(data), data);
+    expect(type.mapFromDatabaseResponse('FALSE'), false);
+    expect(type.mapFromDatabaseResponse('TRUE'), true);
   });
 
   test('writes bool literals', () {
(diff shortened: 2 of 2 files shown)


simolus3:
I think this is expected - sqlite3 does not distinguish between booleans and ints, so drift is doing that work. We should never be getting a raw `bool` directly from the database implementation.

yanivshaked:
@simolus3 Can you please clarify?
Do you mean that the test doesn't correctly demonstrate the problem we encounter?
We have a database with a Boolean column, reading the Boolean values always return true - even if the stores value is false.

simolus3:
> even if the stored value is false

Can you reproduce this with drift alone or is this also coming from a database created by an external tool?

Drift will store `true` as `1` and `false` as `0`. If you have a database where booleans are stored differently, drift may not interpret them correctly.

yanivshaked:
Yes, this is coming from a database created by an external tool. So if I understand correctly, our best solution is to use integer fields rather than boolean, and check for 0 / 1 values?

yanivshaked:
I looked deeper and found that the binary value stored by our external tool is actually using a `string` field with `TRUE` or `FALSE` values. I assume that such implementation is common, since we are using a simple C# project with `Microsoft.Data.Sqlite` to create the tables.

Changing `mapFromDatabaseResponse` to the following (adding the 2nd `if` statement):

```
  @override
  bool? mapFromDatabaseResponse(dynamic response) {
    // ignore: avoid_returning_null
    if (response == null) return null;
    if (response is String) return response.toUpperCase().compareTo('TRUE') == 0;
    return response != 0;
  }
```
solves the issue, and data is read correctly.

yanivshaked:
@simolus3 Please review the suggest code change.

simolus3:
I don't think drift should have a strong compatibility with other sqlite3 libraries outside of simple datatypes. Sqlite doesn't support dates or booleans out of the box, so IMO it's fine for drift to only chose one translation and stick to that. If you declare a boolean column, drift will actually insert a `CHECK` constraint to only allow `0` or `1`. I'm also thinking about using [STRICT tables](https://www.sqlite.org/stricttables.html) for Dart-defined tables in a future drift version for better type safety. If you then define a boolean column and it's actually a string, things will break! Also, sqlite3 really does use `0` and `1` for boolean values (try for yourself: `SELECT TRUE, FALSE, NOT 'TRUE', NOT 'FALSE'` - strings just don't work as booleans).

If you want to store booleans as strings, I think my recommendation would be to use a [type converter](https://drift.simonbinder.eu/docs/advanced-features/type_converters/) instead.

yanivshaked:
Ok, thank you.

PR #1612, opened 2022-01-01
Add check for existence when creating View

### LINKED ISSUE #1611: In the case of using View, an error occurs when calling migration's createAll
In the case of using View, an error occurs when calling migration's createAll

The code：

```dart
class DemoDatabase extends _$DemoDatabase {
  DemoDatabase() : super(_openConnection());

  @override
  int get schemaVersion => 9;

  @override
  MigrationStrategy get migration => MigrationStrategy(
        beforeOpen: (details) async {},
        onCreate: (m) async {
          await m.createAll();
          into(users).insert(PresetUser.createDefaultUser(active: true).map());
          // await insertLocalUser(this);
        },
        onUpgrade: (m, from, to) async {
          await runMigrations(
            this,
            m,
            from,
            to,
            [
              migration1to2,
            ],
          );
          await m.createAll();
        },
      );
}

```

Output:

```log
[ERROR:flutter/lib/ui/ui_dart_state.cc(209)] Unhandled Exception: SqliteException(1): table record_views already exists
E/flutter ( 3474):   Causing statement: CREATE VIEW record_views AS SELECT tasks.user_id AS "user_id", tasks.sid AS "task_id", tasks.name AS "task_name", tasks.repeat_task_id AS "repeat_task_id", tasks.billable_type AS "task_billable_type", tasks.hourly_rate AS "task_hourly_rate", categories.sid AS "category_id", categories.name AS "category_name", categories.color AS "category_color", categories.billable AS "category_billable", categories.hourly_rate AS "category_hourly_rate", records.sid AS "record_id", records.begin_time AS "begin_time", records.end_time AS "end_time", records.data_status AS "data_status" FROM records LEFT OUTER JOIN tasks ON tasks.sid = records.task_id LEFT OUTER JOIN categories ON categories.sid = tasks.category_id
```

T-Oner:
fix #1611

### DIFF
--- a/drift/lib/src/runtime/query_builder/migration.dart
+++ b/drift/lib/src/runtime/query_builder/migration.dart
@@ -326,7 +326,7 @@ class Migrator {
     } else if (view.query != null) {
       final context = GenerationContext.fromDb(_db);
       context.generatingForView = view.entityName;
-      context.buffer.write('CREATE VIEW ${view.entityName} AS ');
+      context.buffer.write('CREATE VIEW IF NOT EXISTS ${view.entityName} AS ');
       view.query!.writeInto(context);
       await _issueCustomQuery(context.sql, const []);
     }
--- a/drift/test/schema_test.dart
+++ b/drift/test/schema_test.dart
@@ -66,7 +66,7 @@ void main() {
           []));
 
       verify(mockExecutor.runCustom(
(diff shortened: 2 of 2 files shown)


simolus3:
You probably shouldn't call `createAll` outside of creating the database, but we also have `IF EXISTS` for tables so it shouldn't hurt. Thanks!

Issue #1630, opened 2022-01-13
Bug in customInsert and customUpdate


topperspal:
```sql
createUser: INSERT INTO auth(user) VALUES(:user);
updateUser(:id AS INT): UPDATE auth SET user=:user WHERE id=:id;
deleteUser: DELETE FROM auth WHERE id=:id;
```

```dart
// Here is first bug
Future<int> createUser(String user) {
    return customInsert(
      'INSERT INTO auth(user) VALUES(:user)',
      variables: [Variable<String>(user)],
      updates: {},
      updateKind: UpdateKind.delete, // throws error => The named parameter 'updateKind' isn't defined.
    );
  }

  // Second bug
    Future<int> updateUser(String user, int id) {
    return customUpdate(
      'UPDATE auth SET user=:user WHERE id=:id',
      variables: [Variable<String>(user), Variable<int>(id)],
      updates: {},
      updateKind: UpdateKind.delete, // Shouldn't it be UpdateKind.update
    );
  }

  Future<int> deleteUser(String id) {
    return customUpdate(
      'DELETE FROM auth WHERE id=:id',
      variables: [Variable<String>(id)],
      updates: {},
      updateKind: UpdateKind.delete, // Fine
    );
  }
```

topperspal:
First bug can be solved like this

![Untitled](https://user-images.githubusercontent.com/83157610/149280946-164e4ec4-b200-43d5-980b-742f8b126450.png)

simolus3:
Can you please post the versions of `drift_dev` and `sqlparser` from your `pubspec.lock` please. Also, are you using any builder options for drift?

I tried to reproduce the first issue with the following drift file:

```sql
CREATE TABLE auth (
  id INTEGER NOT NULL PRIMARY KEY,
  user INTEGER NOT NULL
);

createUser: INSERT INTO auth(user) VALUES(:user);
updateUser(:id AS INT): UPDATE auth SET user=:user WHERE id=:id;
deleteUser: DELETE FROM auth WHERE id=:id;
```

However, that generates the methods just fine:

```dart
  Future<int> createUser(int user) {
    return customInsert(
      'INSERT INTO auth(user) VALUES(:user)',
      variables: [Variable<int>(user)],
      updates: {auth}, // no extra parameter
    );
  }

  Future<int> updateUser(int user, int id) {
    return customUpdate(
      'UPDATE auth SET user=:user WHERE id=:id',
      variables: [Variable<int>(user), Variable<int>(id)],
[...]

PR #1632, opened 2022-01-13
First bug fixed of #1630

topperspal:


simolus3:
Something's weird here: There's generated code right now where the parameter is absent already: https://github.com/simolus3/moor/blob/d12d8c6f4ac0974980270c7673c148a6c4ace121/drift/test/data/tables/custom_tables.g.dart#L1809-L1815

I haven't looked into this issue yet, but I'm not sure if that's the right fix since its not affecting all queries. I'll take a proper look later today.

simolus3:
Closing since it doesn't appear to be a reproducible issue.

Issue #1634, opened 2022-01-16
Aggregate function that creates a dart list


LeFrosch:
**Is your feature request related to a problem? Please describe.**
I’m currently working on a project where I have to organise profiles with multiple profile pictures, thus I have created the following tables:
```sql
CREATE TABLE profile (
    uuid TEXT NOT NULL PRIMARY KEY,
    ...
);

CREATE TABLE image (
    uuid TEXT NOT NULL,
    id TEXT NOT NULL,
    ...

    FOREIGN KEY (uuid) REFERENCES profile (uuid) ON DELETE CASCADE,
    PRIMARY KEY (id)
);
```
Now I want to retrieve one profile with a specific uuid and all it’s images. At first the query seams straight forward:
```sql
getProfile(REQUIRED :uuid AS TEXT): 
SELECT p.**, i.**
FROM profile as p
JOIN image as i USING (uuid)
WHERE p.uuid = :uuid;
```
But this query would return an array of [(the profile, image1),  (the profile, image2), …], with the following generated result type:
```dart
class GetProfileResult {
  final ProfileData p;
  final ImageData i;
  GetProfileResult({
    required this.p,
    required this.i,
  });
}
```

**Describe the solution you'd like**
The perfect return for this query would be (the profile, [image1, image2, …]). Therefore, my feature request would be an aggregate function that creates a dart list. Like: 
```sql
getProfile(REQUIRED :uuid AS TEXT): 
SELECT p.**, TO_LIST(i.**)
FROM profile as p
JOIN image as i USING (uuid)
WHERE p.uuid = :uuid
GROUP BY p.uuid;
```
And a result type like:
```dart
class GetProfileResult {
  final ProfileData p;
  final List<ImageData> i;
  GetProfileResult({
    required this.p,
    required this.i,
  });
}
```
In my opinion this is quite a common use case and an aggregate function like this could be very useful. I would create a PR that tries to implement this feature, but I don’t know where to start and if this would be possible?

simolus3:
While this certainly is a cool feature, it will require a lot of work to get right :) Internally, we'd have to split the queries into two steps:

1. First, load the relevant profiles and store them in an intermediate Dart list
2. Then, go through each profile and load matching images in a separate query.

Rewriting the queries will be quite hard for general SQL queries. It might be a bit easier if we add a syntax where the nested SQL needs to be written down explicitly, maybe:

```sql
getProfile(REQUIRED :uuid AS TEXT): 
[...]

PR #1638, opened 2022-01-17
Support for nested queries like (#1634)

LeFrosch:
I had to do some major changes to the architecture to implement this feature, therefore I created this draft to show them to you. The changes are:
- Selectable.map now accepts async functions
- The QueryWriter now passes the required data around (except the scope), therefore I removed the query field 
- The QueryHandler now passes the required data around (except the context etc.), therefore I removed the source field
- SqlQuery has now a root field, because when generating subqueries I had to change the root AstNode but also wanted to keep the AnalysisContext. I was not able to remove the root field from the analysis context, because some legacy methods evely relied on this field

The generation for the nested query works roughly, but accessing data from the main query is not yet supported. But the example now works: 
```sql
CREATE TABLE profile (
    uuid        TEXT    NOT NULL,
    name        TEXT    NOT NULL,

    PRIMARY KEY (uuid)
);

CREATE TABLE image (
    uuid        TEXT    NOT NULL,
    id          TEXT    NOT NULL,

    FOREIGN KEY (uuid) REFERENCES profile (uuid) ON DELETE CASCADE,
    PRIMARY KEY (id)
);

getProfile(REQUIRED :uuid AS TEXT):
SELECT 
    p.** AS profile,
    LIST(SELECT * FROM image AS i WHERE i.uuid = :uuid) AS images 
FROM profile AS p
WHERE p.uuid = :uuid;
```
Generates: 
```dart
Selectable<GetProfileResult> getProfile(String uuid) {
  return customSelect(
      'SELECT"p"."uuid" AS "nested_0.uuid", "p"."name" AS "nested_0.name", 0 FROM profile AS p WHERE p.uuid = ?1',
      variables: [
        Variable<String>(uuid)
      ],
      readsFrom: {
        image,
        profile,
      }).map((QueryRow row) async {
    return GetProfileResult(
        profile: profile.mapFromRow(row, tablePrefix: 'nested_0'),
        images: await customSelect(
            'SELECT * FROM image AS i WHERE i.uuid = ?1',
            variables: [
              Variable<String>(uuid)
            ],
            readsFrom: {
              image,
            }).map(image.mapFromRow).get());
  });
}

class GetProfileResult {
  final ProfileData profile;
  final List<ImageData> images;
  GetProfileResult({
    required this.profile,
    required this.images,
  });
}
```
Things left todo:
- [x] Remove 0 workaround to fix trailing comma error
- [x] Pass data form the main query 
- [ ] Write some tests :)
- [ ] Investigate how much it takes to support nested nested queries
- [ ] Consider running independent subqueries concurrently if possible.


drift/lib/src/runtime/query_builder/statements/query.dart
```diff
@@ -250,28 +250,29 @@ abstract class Selectable<T>
...
 class _MappedSelectable<S, T> extends Selectable<T> {
   final Selectable<S> _source;
-  final T Function(S) _mapper;
+  final FutureOr<T> Function(S) _mapper;
 
   _MappedSelectable(this._source, this._mapper);
 
   @override
-  Future<List<T>> get() {
+  Future<List<T>> get() async {
```
simolus3, line 265: It looks like this doesn't need to be `async`?

drift/lib/src/runtime/query_builder/statements/query.dart
```diff
@@ -250,28 +250,29 @@ abstract class Selectable<T>
...
     return _source.get().then(_mapResults);
   }
 
   @override
   Stream<List<T>> watch() {
-    return _source.watch().map(_mapResults);
+    return _source.watch().asyncMap(_mapResults);
```
simolus3, line 271: I think I prefer a distinction between the `_mapper` returning a future or not here. `asyncMap` pauses the upstream subscription while the async mapper runs, which seems unnecessary if we know we have a synchronous function.

drift_dev/lib/src/model/sql_query.dart
```diff
@@ -443,19 +468,67 @@ class NestedResultTable {
...
+
+  /// [hashCode] that matches [isCompatibleTo] instead of `==`.
+  @override
+  int get compatibilityHashCode => hashCode;
+
+  /// Checks whether this is compatible to the [other] nested result, which is
+  /// the case iff they have the same and read from the same table.
+  @override
+  bool isCompatibleTo(NestedResult other) => this == other;
+}
```
simolus3, line 530: Maybe the first comment should also point out that we don't currently support using nested queries with reusable [result class name](https://drift.simonbinder.eu/docs/using-sql/drift_files/#result-class-names), which is the reason those methods exist.

IMO the documentation comments don't add much value on this subclass.

drift_dev/lib/src/writer/database_writer.dart
```diff
@@ -117,9 +117,7 @@ class DatabaseWriter {
     }
 
     // Write implementation for query methods
-    for (final query in db.queries ?? const <Never>[]) {
-      QueryWriter(query, dbScope.child()).write();
-    }
+    db.queries?.forEach((query) => QueryWriter(dbScope.child()).write(query));
```
simolus3, line 120: I mostly prefer `for` loops when the body is not a single function call (which [Effective Dart](https://dart.dev/guides/language/effective-dart/usage#avoid-using-iterableforeach-with-a-function-literal) recommends as well).

With your restructuring, we could now do

```dart
final queryWriter = QueryWriter(dbScope.child());
db.queries?.forEach(queryWriter.write);
```

simolus3 (commented):
Thanks for this awesome work :rocket: I had a quick look, I'll take another deeper look at the generated code later.

drift_dev/lib/src/writer/database_writer.dart
LeFrosch: The QueryWriter still has the `final StringBuffer _buffer;` field as long as it is fine to use the same buffer for all queries that should work

drift/lib/src/runtime/query_builder/statements/query.dart
LeFrosch: That's right, my bad :)

drift/lib/src/runtime/query_builder/statements/query.dart
LeFrosch: I’ll just add a map and asyncMap function, like the Stream class has? :)

drift_dev/lib/src/model/sql_query.dart
LeFrosch: Yes I agree. I probably just left them in because I copied these methods from the super class.

LeFrosch:
Just finished the implementation for passing data from the main query to the nested queries :) I think it should be possible to support nested queries in nested queries (multiple nested queries is perfectly fine), but this is currently UB. The example now can be written like this:
```sql
getProfile(REQUIRED :uuid AS TEXT):
SELECT 
    p.** AS profile,
    LIST(SELECT * FROM image AS i WHERE i.uuid = p.name) AS images
FROM profile AS p
WHERE p.uuid = :uuid;
```
And generates the following code:
```dart
Selectable<GetProfileResult> getProfile(String uuid) {
  return customSelect(
      'SELECT"p"."uuid" AS "nested_0.uuid", "p"."name" AS "nested_0.name",0, p.name AS nested_query_0_p_name FROM profile AS p WHERE p.uuid = ?1',
      variables: [
        Variable<String>(uuid)
      ],
      readsFrom: {
        image,
        profile,
      }).asyncMap((QueryRow row) async {
    return GetProfileResult(
      profile: profile.mapFromRow(row, tablePrefix: 'nested_0'),
      images: await customSelect('SELECT * FROM image AS i WHERE i.uuid = ?1',
          variables: [
            Variable<String>(row.read('nested_query_0_p_name'))
          ],
          readsFrom: {
            image,
          }).map(image.mapFromRow).get(),
    );
  });
}
```
The generation for the nested variables works as follows:
1. All References in subqueries are checked if they point to something inside of the nested query or rely on something from the parent query. If they rely on the parent, these references are replaced by variables (`NestedQueryVariable` new subclass of `ColonNamedVariable`). The replacement is done in the sqlpaser in the `NestedQueryResolver`.
2. In the `QueryHandler` every nested request receives a unique name. I tried to use the `InferredResultSet` nested prefix system, but couldn’t get it working. Because when I needed the names the result set was not yet completely built.
3. Before writing the query in the `QueryWriter` the `NestedQueryTransformer` adds all the required result rows to the select statement.

drift_dev/lib/src/analyzer/sql_queries/type_mapping.dart
```diff
@@ -120,14 +120,17 @@ class TypeMapper {
...
     // contain two entries, both referring to the same variable. To do that,
     // we use the fact that each variable has a unique index.
-    final variables = ctx.root.allDescendants.whereType<Variable>().toList();
+    final collector = _VariableCollector();
+    root.accept(collector, null);
+
+    final variables = collector.result;
     final placeholders =
-        ctx.root.allDescendants.whereType<DartPlaceholder>().toList();
+        root.allDescendants.whereType<DartPlaceholder>().toList();
```
LeFrosch, line 133: Because nested queries could contain variables that are not required by the main query, it is no longer possible to just use `root.allDescendants` to find all variables. Also all variables in subqueries are indexed separately .

LeFrosch: But I think for placeholder this should still be fine, idk how they work exactly.

sqlparser/lib/src/analysis/steps/prepare_ast.dart
```diff
@@ -249,4 +255,18 @@ class AstPreparingVisitor extends RecursiveVisitor<void, void> {
...
+    if (e is NestedQueryColumn) {
+      // create a new scope for the nested query to differentiate between
+      // references that can be resolved in the nested query and references
+      // which require data from the parent query
+      e.select.scope = e.scope.createChild();
+      AstPreparingVisitor(context: context).start(e.select);
+    } else {
+      super.visitMoorSpecificNode(e, arg);
+    }
+  }
```
LeFrosch, line 271: This is also the reason why a new `AstPreparingVisitor` is created for every subquery.

drift_dev/lib/src/analyzer/sql_queries/type_mapping.dart
```diff
@@ -340,3 +355,27 @@ class TypeMapper {
     }
   }
 }
+
+/// Because nested variables should not be included when extracting all
+/// FoundElements allDescendants.whereType<Variable>() no longer works.
```
simolus3, line 360: The "no longer works" comment probably isn't too helpful to someone first reading this in the future :) Maybe we should add a comment to the `if (e is NestedQueryColumn)` explaining that we stop here to avoid reading variables from subqieries.

drift_dev/lib/src/analyzer/sql_queries/type_mapping.dart
simolus3: Placeholders are things like `ORDER BY $order`, which generates an `OrderBy` as a Dart parameter.

Thinking about queries like `SELECT ..., LIST (SELECT * FROM ... ORDER BY $order)`, wouldn't that be a problem then?

sqlparser/lib/src/analysis/steps/nested_query_resolver.dart
```diff
@@ -0,0 +1,68 @@
...
+
+  _NestedQueryTransformer(this.context);
+
+  @override
+  AstNode? visitReference(Reference e, void arg) {
+    // if the scope of the nested query cannot resolve the reference, the
+    // reference needs to be retrieved from the parent query
+    if (e.entityName != null &&
+        e.scope.resolve(e.entityName!, includeParents: false) == null) {
+      final result = e.scope.resolve(e.entityName!);
```
simolus3, line 39: Instead of checking scopes manually, have you seen `Reference.resultEntity`? You can use it to get access to the AST node responsible for adding something to a scope, which might be a bit easier to check?

I haven't tried it, but does something like `e.resultEntity.origin.scope != e.scope` work (modulo null handling for unresolved references)?

sqlparser/lib/src/analysis/steps/nested_query_resolver.dart
```diff
@@ -0,0 +1,68 @@
...
+        context.reportError(AnalysisError(
+          type: AnalysisErrorType.referencedUnknownTable,
+          message: 'Unknown table or view in nested query: ${e.entityName}',
+          relevantNode: e,
+        ));
+      } else {
+        return NestedQueryVariable(
+          entityName: e.entityName,
+          columnName: e.columnName,
+        )..setSpan(e.first!, e.last!);
```
simolus3, line 51: I'm not sure if I like entirely synthetic AST nodes introduced this late during analysis.

Is it possible to move parts of this analysis into the `drift_dev` package so that `sqlparser` isn't concerned with all of it? We could somehow mark references that need capturing so that the drift-specific rewriter can pick those up later?

sqlparser/lib/src/analysis/steps/prepare_ast.dart
```diff
@@ -170,16 +170,22 @@ class AstPreparingVisitor extends RecursiveVisitor<void, void> {
     visitChildren(e, arg);
   }
 
-  void _resolveIndexOfVariables() {
+  @override
+  void visitNestedQueryVariable(NestedQueryVariable e, void arg) {
+    _foundVariables.add(e);
+    visitChildren(e, arg);
```
simolus3, line 176: If we end up keeping the `NestedQueryVariable` in sqlparser, it sounds like this is essentially unreachable since we only add them in a later analysis phase? Or am I misreading something?

sqlparser/lib/src/analysis/steps/nested_query_resolver.dart
LeFrosch: The problem I ran into and why I left them inside of the sqlparser is, that after introducing these syntactic variables the indices of all variables have to be calculated again (after the transform call in the _transfrom method) with: 
```dart
AstPreparingVisitor.resolveIndexOfVariables(
  e.allDescendants.whereType<Variable>().toList(),
);
```
And if we move this step to drift_dev some kind of `resolveIndexOfVariables` method needs to be exposed.

sqlparser/lib/src/analysis/steps/prepare_ast.dart
LeFrosch: No, I think that’s right. I probably only added this for completeness sake, but it could be removed.

sqlparser/lib/src/analysis/steps/nested_query_resolver.dart
LeFrosch: I tried it and I think your method also works. Should I remove the `includeParents` flag then?

drift_dev/lib/src/model/sql_query.dart
```diff
@@ -157,6 +158,28 @@ abstract class SqlQuery {
...
+      for (final subElement in subQuery.query.elementsWithNestedQueries()) {
+        if (elements
+            .none((e) => e.dartParameterName == subElement.dartParameterName)) {
+          elements.add(subElement);
+        }
+      }
+    }
+
+    return elements;
+  }
```
LeFrosch, line 182: I added support for the dart placeholder :) I mainly handled them the same way like variables. But I think this method needs some more attention, because this is used to generate code that is required for the main query and all nested queries (for example method parameters).

North101:
So I'm not sure how feasible this is for this feature (and maybe you're already doing it) but in my current projects I do this to be try and be more efficient:
```dart
return CombineLatestStream.combine2<List<Table1>, Map<String, List<Table2>>>(
  db.getTable1(...).watch(),
  db.getTable2(...).watch().map((event) {
    return groupBy<Table2Result, String>(event, (e) => e.table1Id);
  }),
  (table1ResultList, table2ResultListMap) {
    return table1Result.map((table1Result) {
      return FinalResult(
        table1Result: table1Result,
        table2ResultList: table2ResultListMap[table1Result.id] ?? const [],
      );
    });
  },
);
```

The idea being that each query only updates and builds the result when their own table changes. And then the final result gets returned using by combining those results. e.g if Table2 gets updated then table1ResultList doesn't need to be reprocessed.

It could be that my use case is too specific or im wrong about it being more efficient but I thought it was maybe worth mentioning. (sorry for the poor explaination)

LeFrosch:
@North101 If I understand your example, query two does not depend on data retrieved in the first one? But I get your point. In the current implementation, even if the result of the nested query does not change, it will be run again when something from the main query changes.

North101:
Yes, each query is only run when the respective tables are updated. I should have lead with that.

I also meant to mention that this obviously only works if you don't depend on the results from the other table (so if you were passing e.g. an `id` value to be filtered by then i'd pass that to both tables rather than reference the "outer" table)

LeFrosch:
Personally I see three main problems when trying to integrate this optimization into the code generation:
- It would be necessary to check if the nested query does not require any data from the main query. And only if this is the case we could use this optimization. (this should not be to difficult)
- And idk if it is possible to create a generic approach, where it is not necessary to differentiate between watching and getting the result of a query. Is it possible to return a `Selectable` from your method?
- Also the code two join those queries seems quite complex to generate for generic types and an arbitrary amount of nested queries.

sqlparser/lib/src/analysis/steps/nested_query_resolver.dart
LeFrosch: Okey nvm it is not working, somehow resultEntity is always null and I can’t find where it should be set to some value?

LeFrosch: If I set resultEntity in the `ReferneceResolver` like this: 
```dart
void _resolveReferenceInTable(Reference ref, ResultSet resultSet,
    {ResultSetAvailableInStatement? source}) {
  var column = resultSet.findColumn(ref.columnName);
  if (column == null) {
    _reportUnknownColumnError(ref, columns: resultSet.resolvedColumns);
  } else {
    if (source != null) {
      column = AvailableColumn(column, source);
    }
    ref.resolved = column;
    ref.resultEntity = source;
  }
}
```
Then it works :)

simolus3: Oh yeah, it should absolutely be set there. Good catch!

sqlparser/lib/src/analysis/steps/prepare_ast.dart
simolus3: The `assert(false)` is a very good idea for that, thanks!

drift_dev/lib/src/model/sql_query.dart
```diff
@@ -157,6 +158,28 @@ abstract class SqlQuery {
...
+  /// Returns all found elements, from this query an all nested queries. The
+  /// elements returned by this method are in no particular order, thus they
+  /// can only be used to determine the method parameters.
+  ///
+  /// This method makes some effort to remove duplicated parameters. But only
+  /// by comparing the dart name.
+  List<FoundElement> elementsWithNestedQueries() {
+    final elements = List.of(this.elements);
+
+    final subQueries = resultSet?.nestedResults.whereType<NestedResultQuery>();
```
simolus3, line 171: This (and the early bail-out in the collecting visitor when seeing nested result queries) wouldn't supported nested nested queries, like `LIST (SELECT ..., LIST(...))`. For our own sanity, I don't think we have to support that, but then we should add a linter flagging invalid usages of `LIST` somewhere.

LeFrosch: Already did something like this earlier today, just pushed it :)

simolus3:
> Is it possible to return a `Selectable` from your method?

We can always add a `Selectable.combineLatest` to drift :D

> Also the code two join those queries seems quite complex to generate for generic types and an arbitrary amount of nested queries.

True. It could be interesting as a follow-up work (I will take a look at it) for performance though.

LeFrosch:
Btw, what’s the difference between `SqlSelectQuery.requestedResultClass` and `InferredResultSet.resultClassName`?

simolus3:
> Btw, what’s the difference between `SqlSelectQuery.requestedResultClass` and `InferredResultSet.resultClassName`?

`requestedResultClass` is set for queries like `myQuery AS RequestedResultClass: SELECT * FROM ...`. It tells the generator to use `RequestedResultClass` as a name for the generated class and allows to use the same result class for different queries (if they return compatible columns).

`resultClassName` is the actual class name to use. If `requestedResultClass` is non-null, it will be the same as that one. Otherwise it's a new result class or the name of the data class if the query only selects a full table.

LeFrosch:
Thanks for having a lock on this :) I could compose a list of code snippets where I’m not that sure if this is the right approach?

LeFrosch:
In my opinion the following snippets could be improved. I don’t think they are particularly bad, but could cause some problems down the line.

1. `elementsWithNestedQueries` in drift_dev/lib/src/model/sql_query.dart:168
The main problem I have with this method is, that to filter duplicates the elements are compared by their `dartParameterName` which should be unique but I think there could be a better way to filter duplicates.
2. Line 947 in sqlparser/lib/utils/node_to_text.dart
It is necessary to filter out NestedQueryColumns, because otherwise the comma from the join can not be removed. But I think this is somewhat in conflict with the pattern of the NodeSqlBuild and could be confusing when trying to understand the control flow. 
3. In the QueryHandler in line 69 the `requestedResultClass` is passed to the `_QueryHandlerContext.requestedResultClass`. This is later passed to the `SqlSelectQuery.requestedResultClass` and the `InferredResultSet.resultClassName`. I’m not sure if this is correct, because I think before my changes the `requestedResultClass` was only passed to the select query.

simolus3:
I've played around with the implementation and the results a bit, this is very cool! Thanks again for your amazing work here :rocket: 

I managed to move the analysis entirely to the `drift_dev` package by treating outgoing references as variables before they're being replaced.

>  The main problem I have with this method is, that to filter duplicates the elements are compared by their `dartParameterName` which should be unique but I think there could be a better way to filter duplicates.

I think we should probably disallow duplicate placeholders altogether in SQL.

>  It is necessary to filter out NestedQueryColumns, because otherwise the comma from the join can not be removed. But I think this is somewhat in conflict with the pattern of the NodeSqlBuild and could be confusing when trying to understand the control flow.

It looks like we're never generating a query with nested result columns in the end? Isn't the transformer replacing those in an earlier step? I've removed the filter and nothing broke in the generated code.

> In the QueryHandler in line 69 the `requestedResultClass` is passed to the `_QueryHandlerContext.requestedResultClass`. This is later passed to the `SqlSelectQuery.requestedResultClass` and the `InferredResultSet.resultClassName`. I’m not sure if this is correct, because I think before my changes the `requestedResultClass` was only passed to the select query. 

I'll take a look at revisiting the logic used to chose the result class name. We can probably infer the name from the nested result set used and don't need to pass it down with `requestedResultClass`.

LeFrosch:
I took a look at your changes and your approach seems way cleaner than my attempt and also the analysis makes much more sense in the drift package than in the sql parser. 
> It looks like we're never generating a query with nested result columns in the end? Isn't the transformer replacing those in an earlier step? I've removed the filter and nothing broke in the generated code.

Oh yes you're right, in the beginning I kept the columns and just ignored them during code generation. Only later I changed the transformer to remove them completely.

Thanks for all your help regarding this feature :)

simolus3:
No worries, and thank you for starting all the work on this. It's a really cool feature :)

I'll take a look at concurrency with multiple nested queries after merging this to `develop`, the feature is done already and the rest is mostly a matter of optimization.

PR #1642, opened 2022-01-19
Move rxdart tests into separate files

bryanoltman:
Moves rxdart-specific tests into separate files. This makes it easier to exclude these tests from the test suite in environments that don't support rxdart.


bryanoltman:
@simolus3 I'm not especially familiar with Github's CI – do you have any idea why the drift analyze check (`dart analyze --fatal-infos --fatal-warnings`) is failing? This PR doesn't change the files it mentions, and the check passes locally.

simolus3 (approved):
`develop` had an unclean state, but it should all be fine now.

PR #1655, opened 2022-01-25
Update index.md

mit-mit:
Simplify generation commands (there was a bug several releases ago that required the longer style, but should not be needed anymore)


simolus3:
Thanks! I have applied your PR to `develop` as cdde1e57f37868c5ad1506781c96b172a2448bb2. Pushing to `master` would deploy the website right away, but at the moment the build is a bit broken outside of `develop`... The changes will go live with the next release!

PR #1656, opened 2022-01-25
Sqlparser accepts AS in delete statement

LeFrosch:
The Sqlparser accepts AS in delete statements, like the following: 
```sql
DELETE FROM tbl AS t;
```

### DIFF
--- a/sqlparser/lib/src/reader/parser.dart
+++ b/sqlparser/lib/src/reader/parser.dart
@@ -1565,7 +1565,7 @@ class Parser {
 
     _consume(TokenType.from, 'Expected a FROM here');
 
-    final table = _tableReference();
+    final table = _tableReference(allowAlias: false);
 
     final where = _whereOrNull();
     final returning = _returningOrNull();


simolus3:
Is this really supposed to be an error? Looking at the syntax diagram [here](https://www.sqlite.org/lang_delete.html) (see `qualified-table-name`) I would think that this is allowed?

I can also run this through `sqlite3`'s CLI without errors:

```
sqlite> CREATE TABLE foo(bar);
sqlite> DELETE FROM foo AS t;
```

LeFrosch:
You are right, I should have checked with the specs first sorry :L

But I had the trigger like this:
```sql
CREATE TRIGGER AFTER INSERT ON tbl
BEGIN
	DELETE FROM tbl AS t;
END;
```
And when executing this the `AS` is a syntax error. But when I look at the trigger (syntax)[https://www.sqlite.org/lang_createtrigger.html] the `AS` should be fine here too. So I just assumed that `AS` is not allowed with delete statements.

simolus3:
Huh, it looks like this is actually documented. There are a bunch of restrictions on delete and update statements [in triggers](https://www.sqlite.org/lang_update.html#restrictions_on_update_statements_within_create_trigger):

1. The table-name specified as part of an UPDATE statement within a trigger body must be unqualified.
2. The INDEXED BY and NOT INDEXED clauses are not allowed on UPDATE statements within triggers. (fine, we don't parse this at all at the moment)
3. The LIMIT and ORDER BY clauses for UPDATE are unsupported within triggers, regardless of the compilation options used to build SQLite.  (same here, nothing to worry about)
4. The RETURNING clause is not supported for triggers (this one is only mentioned for deletes but it also appears to happen for updates).

If you want to add lints for 1. and 4. that'd be great! There's a `_isTopLevelStatement` in `LintingVisitor`, we can report an error if that's false and we hit a return / qualified table reference.

simolus3:
I will take care of the lints. Thanks for starting the discussion here :)

LeFrosch:
Thanks alot for the fix :) I don’t have that much free time currently, elsewise I would have tried to do it myself :L

PR #1674, opened 2022-02-09
Improve migration guidelines

MathieuPierfitte:
The migration docs using `if (from == 1)` mislead me to think skipping versions would still run intermediate migrations. However, if a user goes straight from v1 to v3, the migration process won't be `1 -> 2` + `2 -> 3` as I thought, but `1 -> 3` directly, thus the code in `if (from == 2)` would never run, skipping a necessary migration.

I think the proposed change might prevent others from making the same mistake as I did.

PS: Thanks heaps for this lib, it's awesome.

### DIFF
--- a/docs/pages/docs/Advanced Features/migrations.md
+++ b/docs/pages/docs/Advanced Features/migrations.md
@@ -18,7 +18,8 @@ class Todos extends Table {
   TextColumn get title => text().withLength(min: 6, max: 10)();
   TextColumn get content => text().named('body')();
   IntColumn get category => integer().nullable()();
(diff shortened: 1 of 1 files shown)


simolus3:
Good point, thanks!

simolus3:
> What do you think about including that in the docs?

Good idea, I'll add that information to the docs. In general, one needs to be a bit careful since some pragmas (in particular, `PRAGMA foreign_keys`) don't work in transactions. So wrapping the whole migration block in a transaction is not a silver bullet, but it definitely can make sense.

kuhnroyal:
And write migration tests :D

PR #1675, opened 2022-02-11
change name of the update method to avoid confusion

saliouseck2009:
The name of the method update in line 140 Cause exceptions for the methods `moveImportantTasksIntoCategory` and `update` method itself .
The problem is when these two methods try to call the native update method on `package:drift/src/runtime/api/runtime_api.dart` the program use the update method define in this page and trigger exceptions . It takes me lots of time .


simolus3:
Good catch, thanks!

PR #1678, opened 2022-02-13
Add custom parent class option to generated data classes

westito:
With this feature you can define a parent class for all generated data classes. I use it for a base model class with ID that can handle foreign key objects with a wrapper, add simple - ID only - equals and use a generic save function in DAOs with generated UUIDs. You can use `dynamic` type argument to add the data class as type argument to parent class (not necessary).

Missing: Check parent class extends 'DataClass'

```dart
/// BaseModel
abstract class BaseModel<T> extends DataClass implements Insertable<T> {
  abstract String id;

  @override
  bool operator ==(Object other) {
    return other is BaseModel &&
        id == other.id &&
        runtimeType == other.runtimeType;
  }

  @override
  int get hashCode => runtimeType.hashCode ^ id.hashCode;
}

/// Declaration
abstract class BaseTable extends Table {

  @override
  Set<Column<dynamic>> get primaryKey => {id};
}

@DataClassName('MyModel', extending: BaseModel<dynamic>)
class MyModels extends BaseTable {
  @override
  String? get tableName => 'my_models';

  TextColumn get name => text().named('name')();
}

/// Generated class
class MyModel extends BaseModel<MyModel>
    implements Insertable<MyModel> {
  String id;
[...]
}

/// Foreign key wrapper
class Ref<T extends BaseModel> {
  late final T value;
  final String id;

  Ref.id(this.id);
  Ref.value(this.value) : id = value.id;

  @override
  String toString() {
    return value.toString();
  }
}

/// Generic save function in DAO
Future<void> save(M entity) async {
  if (entity.id == '') {
    entity.id = idSource.nextId<M>();
    await into(table).insert(entity);
  } else {
    await update(table).replace(entity);
  }
}
```


simolus3:
I like the idea, but I'm not sure about the literal replacement of `dynamic` in the code generation. Maybe we could enforce the following things?

- The parent class must either have zero or one type parameters
- The extended type must be a subtype of `DataClass<Object?>`
- When the parent class has a type parameter, we instantiate it to the actual type of the data class in the generator (as you already did, but I think it's cleaner to explicitly write `$parentClassName<$ownClassName>` instead of replacing `dynamic` in the string)

westito:
Okay, I added checks and modify to your advice. I recommend to rename `@DataClassName` to `@DataClass(name: '')`. With this new option the current naming is not so convenient.

westito:
I modified type check to allow dynamic as type parameter so parent class can be provided without type parameter in annotation. Code will be less verbose.
```dart
@DataClassName('Company', extending: BaseModel)
//vs
@DataClassName('Company', extending: BaseModel<Object>)
```

drift_dev/lib/src/analyzer/data_class.dart
```diff
@@ -0,0 +1,85 @@
...
+          return null;
+        }
+      }
+
+      return className;
+    } else {
+      base.step.reportError(
+        ErrorInDartCode(
+          message: 'Parameter `extending` in @DataClassName must be used with a'
+              ' class',
```
simolus3, line 78: Nit: Let's put the space on the end of the line

```suggestion
          message: 'Parameter `extending` in @DataClassName must be used with '
              'a class',
```

simolus3 (commented):
Thanks! The implementation looks good to me now, but I think we should add some tests, especially around the error cases.

You can take a look at [this file](https://github.com/simolus3/moor/blob/develop/drift_dev/test/analyzer/dart/custom_row_classes_test.dart) for example on how we can test analyzing Dart code. We could have a bunch of tests with data classes misusing this feature (e.g. doing `extends: void Function()`) and ensuring that we get the error message we expect.

westito:
I added the tests

simolus3 (approved):
Thanks for your contribution!

PR #1682, opened 2022-02-14
TypeConverter not existent type argument class workaround

westito:
When a generated data class referenced in `TypeConverter<*>`'s type argument, the analyser can't process returned type, so the field type will be `dynamic` in generated code. This workaround parse type right from source code when no type information available. You must provide type argument for map to use this feature. `.map<GeneratedDataClass>(const MyTypeConverter<GeneratedDataClass>())();`

### DIFF
--- a/drift_dev/lib/src/analyzer/custom_row_class.dart
+++ b/drift_dev/lib/src/analyzer/custom_row_class.dart
@@ -129,12 +129,12 @@ void _checkType(ParameterElement element, MoorColumn column, Step step) {
     return;
   }
 
-  DartType expectedDartType;
+  DriftDartType expectedDartType;
 
   if (column.typeConverter != null) {
     expectedDartType = column.typeConverter!.mappedType;
   } else {
-    expectedDartType = provider.typeFor(column.type);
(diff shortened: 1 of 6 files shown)


simolus3:
I'm not entirely opposed to this, but I'm not sure if I like adding these workarounds. Users should probably configure their build to ensure that drift sees generated code by other builders if needed. The docs explain the other way around (how to [use drift-generated classes in other builders](https://drift.simonbinder.eu/docs/advanced-features/builder_options/#using-drift-classes-in-other-builders)), a similar technique could be used to fix this problem as well.

If we ever implement a mode where drift does not generate part files but rather files with their own imports, we need to know where a type comes from. There are a bunch of places where we don't know that already, but ideally I don't want to encourage workarounds for what effectively sounds like a broken build configuration.

westito:
At generated code I mean the code generated by Drift. I try to reference generated data classes in type converters, but I can't. The only solution without workaround is use custom row classes.

simolus3:
Oh I see, there probably isn't much that could be done about that then (at least not at the moment). The workaround seems reasonable then.

PR #1726, opened 2022-03-04
Fix table in README

creativecreatorormaybenot:
The table was not properly formatted and therefore not typeset by Markdown.

image


simolus3:
Thanks!

PR #1734, opened 2022-03-07
chore: integrate melos (without changing structure)

Ehesp:
Introduces Melos without changing the directory structure.


simolus3:
I've revisited this in https://github.com/simolus3/drift/pull/2060, drift is now using melos. Using it with `usePubspecOverrides: true` for the bootstrap command works really well together with `dart run` and `build_runner`. Thanks for the idea and your support!

Issue #1742, opened 2022-03-09
Don't need to set Schema version every time


YeungKC:
When writing #1741, I got Database is locked, because the reads set schema version again, and I check the source code, I found no way to determine if the version number is updated.

PR #1743, opened 2022-03-09
Determine schema version.

YeungKC:
closes: #1742


simolus3:
Thanks for this! It looks like one test expecting the call to `schemaVersion` needs to be updated.

It would be great to have tests for this: One where the second call to `setSchemaVersion` should happen, one where it shouldn't.

simolus3:
Thanks!

Issue #1761, opened 2022-03-19
View parser issue


marhi:
When running: `flutter clean ; flutter pub get ; flutter pub run moor_generator` I get an error:

```
Failed to build moor_generator:moor_generator:
../../../dev/flutter/.pub-cache/hosted/pub.dartlang.org/drift_dev-1.1.0/lib/src/analyzer/dart/view_parser.dart:101:25: Error: The getter 'snakeCase' isn't defined for the class 'String'.
Try correcting the name to the name of an existing getter, or defining a getter or field named 'snakeCase'.
    return element.name.snakeCase;
                        ^^^^^^^^^
pub finished with exit code 1
```

Environment:
```
Flutter 2.10.3 • channel stable • https://github.com/flutter/flutter.git
Framework • revision 7e9793dee1 (2 weeks ago) • 2022-03-02 11:23:12 -0600
Engine • revision bd539267b4
Tools • Dart 2.16.1 • DevTools 2.9.2
```

dev_dependencies:
```
 ...
  moor_generator: ^4.6.0+1
 ...
```

PR #1762, opened 2022-03-19
fix: wrap element name inside ReCase

marhi:
Fixes an issue mentioned in #1761.

### DIFF
--- a/drift_dev/lib/src/analyzer/dart/view_parser.dart
+++ b/drift_dev/lib/src/analyzer/dart/view_parser.dart
@@ -111,7 +111,7 @@ class ViewParser {
       }
     }
 
-    return element.name.snakeCase;
+    return ReCase(element.name).snakeCase;
   }
 
   Future<Iterable<MoorColumn>> _parseColumns(ClassElement element) async {


simolus3:
While this would probably fix the issue, I don't understand how it could have happened - it should have picked [this extension](https://github.com/techniboogie-dart/recase/blob/master/lib/recase.dart#L114) by default.

I wonder if I got the dependencies wrong and you got an old version of recase that doesn't have the extension? Could you take a look at your `pubspec.lock` and see which exact version of `recase` you got?

marhi:
```  recase:
    dependency: transitive
    description:
      name: recase
      url: "https://pub.dartlang.org"
    source: hosted
    version: "4.0.0"
```

In any case.... might be good that this is explicitly wrapped in ReCase, since the two other occurencies are as well.

simolus3:
> In any case.... might be good that this is explicitly wrapped in ReCase, since the two other occurencies are as well.

Yeah it definitely can't hurt, it's just a bit confusing...

Either way, thanks for your contribution!

PR #1774, opened 2022-03-27
Unique constraint DSL

westito:
- Adds `unique()` column builder option to add UNIQUE constraint without custom constraint declaration
- Adds `uniqueKey` to table definition for multi-column unique key


drift/lib/src/dsl/table.dart
```diff
@@ -62,6 +62,23 @@ abstract class Table extends HasResultSet {
...
+  ///  IntColumn get recipe => integer()();
+  ///  IntColumn get ingredient => integer()();
+  ///
+  ///  IntColumn get amountInGrams => integer().named('amount')();
+  ///}
+  /// ```
+  /// The getter must return a set literal using the `=>` syntax so that the
+  /// drift generator can understand the code.
+  @visibleForOverriding
+  Set<Column>? get uniqueKey => null;
```
simolus3, line 80: I'm not sure if a single getter returning a set is the best way to expose this. After all, a table can have multiple different unique constraints.

The unique constraint on the column builder is a great addition for sure, but for this I think it would be better to have a generic `List<TableConstraint>` getter? Users could then do something like

```dart
List<TableConstraint> get tableConstraints => [
  unique(recipe, ingredient)
];
```

But of course this would be much harder to parse in the generator... Maybe there's a simpler solution that could still support multiple independent unique constraints?

drift/lib/src/runtime/query_builder/schema/table_info.dart
```diff
@@ -28,6 +28,16 @@ mixin TableInfo<TableDsl extends Table, D> on Table
...
 
+  /// The unique key of this table. Can be empty if no custom primary key has
+  /// been specified.
+  ///
+  /// Additional to the [Table.primaryKey] columns declared by an user, this
+  /// also contains auto-increment integers, which are primary key by default.
+  Set<GeneratedColumn> get $uniqueKey => const {};
+
+  @override
+  Set<Column> get uniqueKey => $uniqueKey;
```
simolus3, line 39: I think the `$primaryKey` indirection is a left-over from early days and doesn't need to be replicated here :D

I think we can just do `@override Set<GeneratedColumn> get uniqueKey;` here to make the return type more specific, but it looks like there doesn't have to be a `$uniqueKey` getter.

simolus3 (commented):
Thanks! The column-level unique handling looks good, but I think we need to look at an API supporting multiple `UNIQUE` constraints on tables.

drift/lib/src/dsl/table.dart
westito: I think the best solution is a `List<Set<GeneratedColumn>> uniqueKeys => [{column1, column2}, {column1, column3}]` for multiple keys. `UNIQUE` is the most used table level constraint that's why I think it deserves DSL option. For FKs there is the column level `references()` already. I think we not need to cover all constraint option and add for ex. `ON CONFLICT` clause as DSL. There is the table and column level  `customConstraints` for special needs.

simolus3: Good point, I agree on the `List<Set<GeneratedColumn>>` type which should be able to express this well.

westito: `GeneratedColumn` is not works, because you have to self-reference columns that `Column`

westito:
Done! Drift can handle multiple unique keys now :)

drift/lib/src/runtime/query_builder/schema/table_info.dart
```diff
@@ -28,6 +28,14 @@ mixin TableInfo<TableDsl extends Table, D> on Table
   @override
   Set<Column> get primaryKey => $primaryKey;
 
+  /// The unique key of this table. Can be empty if no custom primary key has
+  /// been specified.
+  ///
+  /// Additional to the [Table.primaryKey] columns declared by an user, this
+  /// also contains auto-increment integers, which are primary key by default.
+  @override
+  List<Set<Column>> get uniqueKeys => const [];
```
simolus3, line 37: It could be a `List<Set<GeneratedColumn>>` in this override though, right?

drift_dev/lib/src/analyzer/dart/column_parser.dart
```diff
@@ -355,6 +359,17 @@ class ColumnParser {
       );
     }
 
+    if (foundFeatures.contains(const UniqueKey()) &&
```
simolus3, line 362: Could you add tests for the error cases similar to the one in say `test/analyzer/dart/column_references_test.dart`?

drift/lib/src/runtime/query_builder/migration.dart
```diff
@@ -288,6 +288,21 @@ class Migrator {
       context.buffer.write(')');
     }
 
+    if (table.uniqueKeys.isNotEmpty) {
```
simolus3, line 291: Can you also add a test to `drift` by adding a unique key to some table in `test/data/tables/todos.dart` and make sure that the unique constraint is generated in the migrator?

simolus3 (commented):
Looking good now, thanks! There should be some tests for this in this generator and in drift to make sure everything is working as intended though :)

drift/lib/src/runtime/query_builder/schema/table_info.dart
westito: Really

drift/lib/src/runtime/query_builder/migration.dart
westito: Sure!

westito:
I regenerated all .g files, so there is a lot of change according to it. I had to fix an analzye issue (maybe caused a change in the new Dart version). Aaaand, I had to update SQLite version from 1.5.1 to 1.6.1 in drift after merge last development branch, because code compile failed.

drift/lib/src/ffi/database.dart
```diff
@@ -1,7 +1,6 @@
 import 'dart:io';
 
 import 'package:meta/meta.dart';
-import 'package:sqlite3/common.dart';
```
simolus3: Are you on an older Dart or Flutter version by any chance? In the latest `sqlite3` version, the `dispose` method was moved from `Database` to `CommonDatabase`  and I was getting lints about an unresolved comment reference. 

I've forgotten to update the constraint in the pubspec though, I'll fix that.

simolus3 (commented):
Thanks for the integration test!

drift/lib/src/ffi/database.dart
simolus3: Ah, I just saw your comment. But with `sqlite3` version 1.6.1 importing `common.dart` should work though.

westito: I don't know what's the problem actually. Test simply not compiled. Can you fix this and resolve conflicts?

simolus3: The CI works now so I think it should be fine? For me, the tests in `drift` and `drift_dev` work locally too (I'm on Dart 2.16.2). Let me know if you still have problems with them.

simolus3:
I have added a section in the web documentation for this and some generator tests, everything looks good to me now.

Thanks again for your contribution! :heart:  :rocket:

PR #1775, opened 2022-03-27
Fix null aware type converter nullability suffix

westito:
When using null aware type converters, drift generator doubles the nullability operator in generated data class. In this case not need to append the question mark because type converter's type already have it and field nullability depends on that. 

Képernyőfotó 2022-03-27 - 15 26 43


simolus3:
Thanks!

PR #1795, opened 2022-04-06
Some minor bug fixes :)

LeFrosch:
While using your library I found two bugs and since I am a little bit familiar with your code base, I tried to fix them myself. 

The first bug was related to `RETURNING` insert statements, the following custom select statement was generated from this example:
```sql  
insertTest:
INSERT OR IGNORE INTO profile (uuid, name) VALUES ('test', 'test') RETURNING uuid;
```
Generated:
```dart
return customWriteReturning(
    'INSERT OR IGNORE RETURNING uuid INTO profile (uuid, name) VALUES (\'test\', \'test\') RETURNING uuid',
    variables: [],
    updates: {
      profile
    }).then((rows) =>
    rows.map((QueryRow row) => row.read<String>('uuid')).toList());
```

The second bug was a little bit tricky, because somehow the 'ColonNamedVariable'  could overwrite captured variables in nested queries. I think the following example makes it clear:
```sql  
getProfile(REQUIRED :uuid AS TEXT):
SELECT 
    p.** AS profile,
    LIST(SELECT i.uuid, * FROM image AS i WHERE i.uuid = :uuid OR  i.uuid = p.uuid )
FROM profile AS p;
```
Generated:
```dart
...

nestedQuery0: await customSelect(
       'SELECT i.uuid, * FROM image AS i WHERE i.uuid = ?1 OR i.uuid = ?1',
        variables: [
          Variable<String>(uuid)
        ],

...
```

The fixes I found where quite simple and I hope they can resolve these issues :)

### DIFF
--- a/sqlparser/lib/utils/node_to_text.dart
+++ b/sqlparser/lib/utils/node_to_text.dart
@@ -727,7 +727,6 @@ class NodeSqlBuilder extends AstVisitor<void, void> {
         InsertMode.insertOrFail: TokenType.fail,
         InsertMode.insertOrIgnore: TokenType.ignore,
       }[mode]!);
-      visitNullable(e.returning, arg);
     }
 
     _keyword(TokenType.into);
--- a/drift_dev/lib/src/analyzer/sql_queries/nested_queries.dart
+++ b/drift_dev/lib/src/analyzer/sql_queries/nested_queries.dart
@@ -84,6 +84,13 @@ class NestedQueryAnalyzer extends RecursiveVisitor<_AnalyzerState, void> {
   }
 
   @override
+  void visitVariable(Variable e, _AnalyzerState arg) {
+    arg.actualAndAddedVariables.add(e);
+
+    super.visitVariable(e, arg);
+  }
+
+  @override
   void visitReference(Reference e, _AnalyzerState arg) {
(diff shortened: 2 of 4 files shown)


sqlparser/lib/utils/node_to_text.dart
```diff
@@ -727,7 +727,6 @@ class NodeSqlBuilder extends AstVisitor<void, void> {
         InsertMode.insertOrFail: TokenType.fail,
         InsertMode.insertOrIgnore: TokenType.ignore,
       }[mode]!);
-      visitNullable(e.returning, arg);
```
simolus3, line 730: Well this is embarrassing :sweat_smile: Thanks for the fix! Could you also add a test case similar to this one? https://github.com/simolus3/drift/blob/9e4e177b97d9804aa7bc903b42ad07d4dca2b8fb/sqlparser/test/utils/node_to_text_test.dart#L303-L305

simolus3 (commented):
Thanks for the fixes!

sqlparser/lib/utils/node_to_text.dart
LeFrosch: I added the test case you ask for and also edited the tests for nested queries to include variables that refer to the enclosing query :)

PR #1815, opened 2022-04-23
Fix Headings for database comparison in FAQ

JosiahMendes:
Floor and Firebase were H2 headings when I think they should have been H3 to fall under "How does drift compare to X?"

Thanks for all your hard work on drift! Very useful and much appreciated! :)


simolus3:
Thanks!

PR #1820, opened 2022-04-25
Fix not null when using JsonTypeConverter

North101:
When using JsonTypeConverter, append `!` if the column is not null.


simolus3:
Good catch, thanks for the fix!

PR #1835, opened 2022-05-11
Add BigInt support

westito:
Add support for dedicated BigInt column type to end-to-end int64 support on web. Currently only works with `WasmDatabase`.


simolus3:
I think we should definitely support BigInts for better web support, but I'm not so sure about introducing a separate column type for it. To me, having a `BigInt`-specialized column indicates that this column would actually be able to store arbitrarily-large integers. That's not true though, it's still a 64-bit integer in the database. I wonder if we could instead have something like:

```dart
class MyTable extends Table {
  IntColumn get someInt => integer()(); // reported as an int in Dart
  IntColumn get probablyLargeIntUsedOnTheWeb => integer(bigInt: true)(); // reported as a BigInt in Dart
}
```

Then:

- native-only database implementations would continue to expose ints as `int` internally
- on the web, the wasm database implementation would expose sqlite ints as `BigInt` internally
- a translation unit in drift takes care of exposing them as an `int` or as a `BigInt` to the user, depending on the column type chosen.

Since a variable doesn't expose its value, there could also be a `Variable.withBigInt` returning a `Variable<int>` while storing a `BigInt` internally since it's the same type in sqlite3.

westito:
Yeah, BigInt name isn't a good choice. May we can implement a custom Int64 type with bound check. I didn't want a new column type, but I thought it is the only way. IntColumn only accepts `int`, so for ex. mapping won't work for int64.
What if, there is a separate `Int64Column get => int64()` column type, but under the hood no new `ColumnType` only generated code changes.

simolus3:
I think that would still be kind of confusing (if we add an `int64` column now, users might be confused what an `int` column was all along). If there's no new `ColumnType` associated with it, couldn't we go with the parameter on the `integer()` column builder to indicate whether this column should generate as a `BigInt`?

simolus3:
Oh of course I've overlooked that part sorry. Your approach sounds fine to me then. For a name, let's use something like `intAsBigInt()` until we come up with something better, at least that clearly describes what this is doing :D

westito:
I can't find a better solution. All types based on `ColumnType`. It is much more overhead to keep only integer column type and hack `BigInt` into to it somehow. I renamed to `Int64Column`, so it clarifies this is not a 'real' `BigInt`.
- I fixed tests that broken because of Dart 2.17.

drift/lib/src/web/web_db.dart
```diff
@@ -166,7 +175,7 @@ class _WebDelegate extends DatabaseDelegate {
 
     while (stmt.step()) {
       columnNames ??= stmt.columnNames();
-      rows.add(stmt.currentRow());
+      rows.add(stmt.currentRow(useBigInt));
```
simolus3, line 178: I assume this applies to all integer values then? There's no way to only expose certain columns as `BigInt`? If so, should we adapt the `IntType` implementation to map `BigInt`s from SQL into Dart ints to handle this case?

Also, for the custom wasm implementation in `package:sqlite3`, there might be a case where we _want_ an `int` in Dart (because the user didn't write `int64()`), but actually get a `BigInt` (because that implementation always reports large ints as `BigInt`). For that we also need a conversion in `IntType` I believe?

drift/lib/src/web/sql_js.dart
```diff
@@ -70,7 +70,7 @@ class _QueryExecResult {
 class _SqlJsStatement {
   external void bind(List<Object?> values);
   external bool step();
-  external List<Object?> get();
+  external List<Object?> get(Object? params, dynamic config);
```
simolus3, line 73: `jsify` has mediocre codegen and performance, we can write this like this:

```dart
@JS
@anonymous
class _SqlJsStatementGetOptions {
  external _SqlJsStatementGetOptions({bool useBigInt = false});
}

// And then use
external List<Object?> get(Object? params, _SqlJsStatementGetOptions config);
get(null, _SqlJsStatementGetOptions(useBigInt: useBigInt))
```

dart2js will emit a direct JS object literal when creating anonymous JS classes which is what we want.

drift/lib/src/web/sql_js.dart
```diff
@@ -134,7 +137,15 @@ class SqlJsDatabase {
...
-      _obj.run(sql, args);
+      _obj.run(
+          sql,
+          args.map((e) {
+            if (e != null && e is BigInt) {
+              return _bigInt(e.checkRange.toString());
+            } else {
+              return e;
+            }
+          }).toList());
```
simolus3, line 148: Let's use a collection for here, something like

```dart
[for (final arg in args) if (e is BigInt) _bigInt(e.checkRange.toString()) else e]
```

(Also no need for the `e != null`, it's included in the `is` even without null-safety).

drift/lib/src/runtime/query_builder/expressions/aggregate.dart
```diff
@@ -112,6 +112,49 @@ extension ArithmeticAggregates<DT extends num> on Expression<DT?> {
       _AggregateExpression('TOTAL', [this], filter: filter);
 }
 
+/// Provides aggregate functions that are available for BigInt expressions.
+extension BigIntAggregates<DT extends BigInt> on Expression<DT?> {
```
simolus3, line 116: I'm not entirely sure whether we should expose these numeric expressions for `BigInt` as well since it's the same thing as a normal integer in SQL. I think the main convenience is  that drift automatically reports those values as `BigInt`s without any loss of precision along the way, but that's it.

So if users want to do arithmetic on BigInts, IMO they can do `(foo.dartCast<int>() + bar.dartCast<int>()).dartCast<BigInt>()`. Sure, having that in drift is way more convenient, but

- using `BigInt`s is not something that most users will want to deal with in the end (it's only useful for columns known to store values larger than 2⁵², so typically large IDs on which one doesn't do arithmetic a lot)
- we realistically need to maintain those expressions separately, with integration tests and everything

I see how this isn't ideal for few users, but I think we can make our lives easier here by forcing users to go the indirect way through `dartCast`.

simolus3 (commented):
Thanks for the update! I think we should also do an implicit conversion from `BigInt` to `int` in `IntType`, but apart from that this looks pretty good.

I'll also release a new version of the sqlite3 package now so that we can actually make use of this :D

drift/lib/src/web/web_db.dart
westito: I didn't fully understand. When `useBigInt` enabled (it's only for WebDatabase sql.js implementation) all `int`s read as `BigInt`. At the end doesn't matter it is `int` or `BigInt` because it's always converted to String and mapped back to `int`/`BigInt` depending on column type (this is default Drift behaviour). WASM is returns `BigInt` only if the number > safeInt. As whole, it doesn't matter Drift runs on WebDatabase (with or without `useBigInt`) or on WASM that returns `int` or `BigInt`.

drift/lib/src/web/sql_js.dart
westito: Thanks. I thought it is not the best solution, but the docs for jsinterop is almost zero.

drift/lib/src/runtime/query_builder/expressions/aggregate.dart
westito: I do not agree. Casting to int is against the sense of BigInt. I understand using BigInt rather edge case and affect only few users. But skipping BigInt aggregations and arithmetics will make BigInt support partial and users won't understand why operations do not work on some columns. I think these two class doesn't make things more complicated compared to the whole project.

drift/lib/src/web/web_db.dart
simolus3: Ah fair point. Yeah the implementation could probably be optimized by doing `if (response is BigInt) return response.toInt()` but it should work already.

drift/lib/src/runtime/query_builder/expressions/aggregate.dart
simolus3: > Casting to int is against the sense of BigInt

Is it though? In SQL, a `Expression<BigInt>` and `Expression<int>` is the same thing. The only difference is how that value is mapped from and to Dart. So casting intermediate values to `Expression<int>` will not change the query in any way.

Perhaps we can keep the new API but use casts around the existing methods internally (it will generate the same SQL), I'll think about it some more. If it turns out to be easy to support, I agree that it's better to have these methods. I'm mostly worrying about duplicate code that's easy to forget about when making changes to the original APIs for `int`.

westito: Yeah. Sorry. I forget it is just a DSL for SQLite. And this internal cast would be nice. The BigInt aggreg is casted to regular int.

westito:
Okay. I modified what you asked for. Left an issue: DriftClient is don't know that the server side database supports BigInt or not. No communication for it. My workaround is put a parameter to DriftClient (supportsBigInt). This is only matters and throw exception if user use int64 column.

drift/lib/src/runtime/query_builder/expressions/aggregate.dart
simolus3: I think it should be fine for postgres too since the generated SQL is the same and since those aggregates also work on `bigint`s in postgres?

simolus3:
> Left an issue: DriftClient is don't know that the server side database supports BigInt or not.

Thinking about it some more, it may be saner to assume that a `BigInt` is supported if a user is using them. That way we avoid all this mess with the new compatibility getter. I can remove that logic if you don't see a downside. My idea is to essentially:

- Support `BigInt` in `drift_sqflite` as well by mapping it to an int (direction from the database to `BigInt` is not necessary since `drit_sqflite` only works on native platforms and since `BigIntType` supports mapping from `int`s implicitly)
- Very prominently warn that `BigInt` supports need to be enabled on a `WebDatabase` to be supported. We could add this information to the `int64()` column builder, the changelog, and the website's section on `BigInt` support. That way users now that they need to set that flag to use `BigInt`s on the web.
- When a `BigInt` variable is used on a `WebDatabase` without bigint support, we can throw an exception asking users to enable the flag. I think it's unlikely that users will only read `BigInt`s, so that way get can point them towards the right thing.

With that done, we can hopefully remove the `supportsBigInt` getter on `QueryExecutor` and all the parameters needed for the remote. I can take care of that, I'm just wondering if you see a downside that I might be missing?

westito:
There is an exception when user try to create table with BigInt without support. I was thinking on this, but underlying database does not know what type the column is. It gets a query + raw parameter list. But you're right. It can be simply check that WebDatabase gets BigInt parameter type or not. Feel free to modify

simolus3 (approved):
Looks good to me now, thanks for your work on this :+1:

westito:
There is a big issue with DriftServer/Client connection on SharedWorker. Client tries to send BigInt through the channel, but it is fail to convert to JS:

```
[   +4 ms] Error: UnimplementedError: structured clone of other type
               at Object.throw_ [as throw] (http://localhost:60183/dart_sdk.js:5080:11)
               at html_common._StructuredCloneDart2Js.new.walk (http://localhost:60183/dart_sdk.js:112639:17)
               at html_common._StructuredCloneDart2Js.new.copyList (http://localhost:60183/dart_sdk.js:112647:29)
               at html_common._StructuredCloneDart2Js.new.walk (http://localhost:60183/dart_sdk.js:112625:21)
               at html_common._StructuredCloneDart2Js.new.copyList (http://localhost:60183/dart_sdk.js:112647:29)
               at html_common._StructuredCloneDart2Js.new.walk (http://localhost:60183/dart_sdk.js:112625:21)
               at html_common._StructuredCloneDart2Js.new.copyList (http://localhost:60183/dart_sdk.js:112647:29)
               at html_common._StructuredCloneDart2Js.new.walk (http://localhost:60183/dart_sdk.js:112625:21)
               at html_common._StructuredCloneDart2Js.new.convertDartToNative_PrepareForStructuredClone (http://localhost:60183/dart_sdk.js:112652:23)
               at Object.convertDartToNative_PrepareForStructuredClone (http://localhost:60183/dart_sdk.js:113325:58)
               at Object.convertDartToNative_SerializedScriptValue (http://localhost:60183/dart_sdk.js:113238:24)
               at MessagePort.[dartx.postMessage] (http://localhost:60183/dart_sdk.js:91087:35)
```

simolus3:
Good catch! This should be fixed in 21170b7ce850a74c44201b0df8f28a2bfc243bc1.

PR #1840, opened 2022-05-17
Dart view aliases

simolus3:
- Generate `CREATE VIEW` statements with explicit column names matching the getter names in Dart.
- In the `SELECT` statement for views, give each table a short alias. This allows writing view queries joining the same table more than once (https://github.com/simolus3/drift/discussions/1839).
- For view columns that reference existing columns from a joined table by writing them in the `select()` query, don't include column metadata (it doesn't appear to be used).

@westito If you have some time to take a look at this I'd love to hear your thoughts on this since you wrote the original implementation. Do you believe this might break something?


westito:
LGTM 💪🏻

westito (approved):

PR #1845, opened 2022-05-19
Add BigInt support to PgDatabase

westito:
I added BigInt value converter for postgres

### DIFF
--- a/extras/drift_postgres/lib/src/pg_database.dart
+++ b/extras/drift_postgres/lib/src/pg_database.dart
@@ -59,9 +59,8 @@ class _PgDelegate extends DatabaseDelegate {
       final args = row.arguments;
 
       await _ec.execute(stmt,
-          substitutionValues: args
-              .asMap()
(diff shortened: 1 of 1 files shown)


extras/drift_postgres/lib/src/pg_database.dart
```diff
@@ -93,13 +93,19 @@ class _PgDelegate extends DatabaseDelegate {
       result = await _ec.query(statement);
     } else {
       result = await _ec.query(statement,
-          substitutionValues: args
-              .asMap()
-              .map((key, value) => MapEntry((key + 1).toString(), value)));
+          substitutionValues: args.asMap().map((key, value) =>
```
simolus3, line 96: This looks good to me, but don't `runBatched` and `_runWithArgs` need the same transformation?

westito: You sure, I missed.

simolus3 (approved):

PR #1846, opened 2022-05-20
Fix column writer nullability check

westito:
I found problem (?) with column generation regarding to nullability + converters. 
The generated data class is not nullable, but other function treat type as nullable. I made a possible fix. Am I right? 
 
Képernyőfotó 2022-05-20 - 14 41 56
Képernyőfotó 2022-05-20 - 14 42 27
Képernyőfotó 2022-05-20 - 14 41 49


drift_dev/lib/src/model/types.dart
```diff
@@ -75,7 +75,7 @@ class DriftDartType {
 extension OperationOnTypes on HasType {
   /// Whether this type is nullable in Dart
   bool get nullableInDart {
-    return (nullable && !isArray) || typeConverter?.hasNullableDartType == true;
```
simolus3, line 78: This may have to change in the future, but I believe your changes are correct now.

Once we get nullability with type converters right in the next breaking version, there could be a way to apply non-nullable type converters onto a nullable column, in which case the old method is correct because the result would still be nullable.

simolus3 (approved):
Looks good to me, good catch!

simolus3:
Huh, looks like this causes us to sometimes generate a duplicate nullability suffix (`PureDefault copyWith({MyCustomObject?? txt})` in the failing build step).

drift_dev/lib/src/model/types.dart
westito: It can be nullable. But this is the responsibility of user. If column is not nullable, converter can still return null in any case. It was true before this dartType depend nullability.

westito:
I found the problem. Fixed!

drift_dev/lib/src/model/types.dart
simolus3: I think this should be something like the following now:

```dart
if (isArray) return false; // Returns a non-nullable `List`

final converter = typeConverter;
if (converter != null) {
  // Nullable if the type converter is nullable in Dart or if nulls side-step the converter.
  return converter.hasNullableDartType || (converter.canBeSkippedForNulls && nullable);
}
return nullable;
```

simolus3 (approved):
I the logic to determine whether a type is nullable in Dart needs a slight update with the changed to converters now. Apart from that this looks good to me.

Also, sorry for the slow response on this. If you want to I can fix the conflicts and merge it now.

westito:
The issue came back. The field is nullable, but the converter isn't.

Képernyőfotó 2022-08-19 - 9 23 00

PR #1847, opened 2022-05-21
Add synchronised persistence to WasmDatabase

westito:
I made my synchronised persistence idea. Modification made to sqlite3 is required to this feature. I committed my sqlite3 repo in the pubspec.yaml temporarily. I think IndexeddbFileSystem may need an optimisation for a better performance. You said an "asyncWrite: false" option. But a delayed persist must be exists. It is possible sqlite3 writes to file besides a query.


drift/lib/src/sqlite3/database.dart
```diff
@@ -17,15 +17,18 @@ abstract class Sqlite3Delegate<DB extends CommonDatabase>
   bool _hasCreatedDatabase = false;
   bool _isOpen = false;
 
+  final bool _syncPersistence;
```
simolus3, line 20: I'm not sure if this even needs to be optional? I think users expect changes to be in the database after awaiting the future.

drift/lib/src/sqlite3/database.dart
```diff
@@ -132,6 +143,9 @@ abstract class Sqlite3Delegate<DB extends CommonDatabase>
     if (_closeUnderlyingWhenClosed) {
       beforeClose(_db);
       _db.dispose();
+      if (_syncPersistence) {
+        await _db.flush();
```
simolus3, line 147: To avoid the new API on the database (see my comments on the sqlite3 PR), we could export the environment through `WasmSqlite` instead of the database. Then we could call `flush()` directly on the file system instead of adding new methods to the database that only make sense on some platforms.

westito:
You sure. I thought this would useful for libs other than Drift. That's why I put some interface to sqlite3.dart. This commit wire filesystem directly into Drift. The sqlite3.dart PR I made not need anymore.

simolus3 (approved):
LGTM :+1:

PR #1883, opened 2022-06-08
Fix remove null value not work.

sososdk:


drift/lib/src/runtime/query_builder/statements/update.dart
```diff
@@ -65,7 +65,7 @@ class UpdateStatement<T extends Table, D> extends Query<T, D>
     _sourceTable.validateIntegrity(entity).throwIfInvalid(entity);
 
     _updatedFields = entity.toColumns(true)
-      ..remove((_, value) => value == null);
+      ..removeWhere((_, value) => value == null);
```
simolus3, line 68: Thanks for your contribution! It's true that this `remove` call is obviously wrong and was probably meant to be a `removeWhere`, but now I'm wondering if it's necessary at all:

- When using updates with a data class, null values aren't part of the map since we pass `nullToAbsent: true` in the `toColumns` method.
- When using updates with a companion and there are null values, I think we should definitely write those instead of dropping them.

So I think the correct behavior may be to just drop the `remove` entirely. Do you have an example where null values are written when you didn't expect that to happen?

simolus3:
I've removed the offending call entirely in 721b894141aa096964adb55663f1354803d19990. Thanks for pointing this out and for your PR!

PR #1896, opened 2022-06-19
Using read<int> instead readInt

cnmade:
by the document and api document, readInt may replace with read<int>?

`
Use read<int>(key) directly
`


simolus3 (approved):
Nice catch, thanks for the contribution!

PR #1899, opened 2022-06-22
TypeConverter mapToDart -> fromSql, mapToSql -> toSql

North101:
So this is just a change to make TypeConverter and JsonTypeConverter more similar in syntax.

Since 2.0 is a breaking change I thought I'd try to slip this in :P
If you don't want it you can just decline / close it.

I was also considering changing `TypeConverter` to `SqlTypeConverter`


simolus3:
`mapToDart` never felt like a very good name for that method given that it's still returning a Dart object. I like the name change :+1:

PR #1912, opened 2022-06-28
Fix tests

westito:
- I updated my generator, local test runner script. 
- Fixed some compiler warnings
- Update DataClass/Companion writer according to new TypeConverter

<u>Two issues:</u>
1.
Current (NullAware)TypeConverter not fits well when you want use custom queries. 
Because of modified NullAwareTypeConverter (`06af930941407e16b3a0f2f1bdeb5b7d9038fcb2`) I had to update `extras/app` after code generation. The `color` property was a non-null type before, now it is generated as nullable. Okay, then I can use plain TypeConverter instead of NullAware. But! There is a custom query that can return nullable value for `color` (in a generated result class), so I have to use nullable converter and get this warning 🤷🏻‍♂️
```
 drift_dev:drift_dev on lib/database/database.dart:line 30, column 40 of package:app/database/tables.dart: This column is non-nullable in the database, but has a type converter with a nullable SQL type, meaning that it may potentially map to `null` which can't be stored in the database.
   ╷
30 │   IntColumn get color => integer().map(const ColorConverter())();
   │                                        ^^^^^^^^^^^^^^^^^^^^^^
```

2.
I ran `sqlparser` tests on chrome accidentally and I got this error. I suspect sqlparser will never run in a browser, but it might be worth a look.
```
00:06 +53 -1: test/analysis/types2/misc_cases_test.dart: miscellaneous type inference cases for SELECT (3 / 4.) = ? [E]                                                                                        
  Expected: ResolvedType:<ResolvedType(BasicType.real, hint: null, nullable: false, array: false)>
    Actual: ResolvedType:<ResolvedType(BasicType.int, hint: null, nullable: false, array: false)>
```


examples/web_worker_example/build.yaml
```diff
@@ -23,7 +10,8 @@ targets:
...
           new_sql_code_generation: true
       build_web_compilers|entrypoint:
+        enabled: true
         generate_for:
-          # This one is compiled in the other target
-          exclude:
-            - "web/worker.dart"
+          - web/worker.dart
+        options:
+          compiler: dart2js
```
simolus3, line 17: Was the previous build configuration breaking anything?

For debugging purposes, I prefer the main entrypoint to be compiled with dartdevc. The worker needs to be dart2js and that's fine, but if possible I want to use both compilers in a single build here.

extras/integration_tests/drift_testcases/build.yaml
```diff
@@ -0,0 +1,14 @@
+targets:
+  $default:
+    builders:
+      drift_dev:
+        options:
+          override_hash_and_equals_in_result_sets: true
+          generate_connect_constructor: true
```
simolus3, line 7: While we're at it, can we enable `new_sql_code_generation` as well?

tool/generate_all.sh
```diff
@@ -1,20 +1,24 @@
...
-echo "- Generate migrations_example"
-(cd ../extras/migrations_example && rm -rf .dart_tool && dart pub get && dart run build_runner build --delete-conflicting-outputs)
-echo "- Generate web_worker_example"
-(cd ../extras/web_worker_example &&  rm -rf .dart_tool && dart pub get && dart run build_runner build --delete-conflicting-outputs)
-echo "- Generate with_built_value"
-(cd ../extras/with_built_value &&  rm -rf .dart_tool && dart pub get && dart run build_runner build --delete-conflicting-outputs)
-echo "- Generate flutter_web_worker_example"
-(cd ../extras/flutter_web_worker_example && rm -rf .dart_tool && dart pub get && dart run build_runner build --delete-conflicting-outputs)
+
+function generate() {
```
simolus3, line 4: Nice, thanks!!

drift_dev/lib/src/writer/tables/update_companion_writer.dart
```diff
@@ -194,7 +194,9 @@ class UpdateCompanionWriter {
           ..write(mapSetter)
           ..write('(converter.toSql($getterName.value)');
 
-        if (!column.nullable && scope.generationOptions.nnbd) {
+        if (!column.nullable &&
+            converter.sqlTypeIsNullable &&
+            scope.generationOptions.nnbd) {
```
simolus3, line 199: I'm not sure if we need this check at all now. My understanding is that this only happens if we have a non-null value in Dart that may get mapped to SQL `NULL` by the type converter?

In that case, I think we should include the `null` value instead of asserting that it doesn't exist. We may have to change `variableTypeCodeWithoutArray` so that it returns a nullable type if either the column is nullable or if there's a converter with a nullable SQL type (currently it only checks the former).

drift_dev/lib/src/writer/tables/data_class_writer.dart
```diff
@@ -269,8 +269,9 @@ class DataClassWriter {
         // apply type converter before writing the variable
         final converter = column.typeConverter;
         final fieldName = converter!.tableAndField;
-        final assertNotNull = !column.nullable && scope.generationOptions.nnbd;
-
+        final assertNotNull = !column.nullable &&
```
simolus3, line 272: Same here, maybe we don't need this check anymore.

examples/web_worker_example/build.yaml
simolus3: Ah I see what's going wrong now. I'll try to fix it with advice from https://github.com/dart-lang/build/issues/3275#issuecomment-1087697299.

simolus3 (commented):
Thanks! This should make it a lot easier to quickly re-generate all examples.

I have some questions about the null handling from Dart to SQL with converters and build options in the web worker example. But maybe I'm overlooking some complexity.

westito:
Postgres tests: As I see you import the common test files into PG lib. It won't run because custom queries in SQLite tests contains '?' as placeholder mark and  ':xy' named parameters that postgres not supports. In the first case '?' have to be replaced by '@1,2,3,..' that SQLite supports too (pg not supports anonymous placeholders anyways). In the latter case the ':' argument mark has to be replaced by '@'.

examples/web_worker_example/build.yaml
simolus3: Fixed in 4fae017d364da6097dee7bdd50a620e541e0f3ff.

westito: Yes, this was the problem. Sorry, I didn't notice that this build file should generates two separate files

drift_dev/lib/src/writer/tables/update_companion_writer.dart
westito: I don't fully understand the second paragraph. This is only eliminates an unnecessary assertion. I think this is not modify the logic at all. Before my change the generator always wrote a `!` at the end of `toSql()` even the TypeConverter has non-nullable type argument (lint: unnecessary_non_null_assertion).

drift_dev/lib/src/writer/tables/data_class_writer.dart
simolus3: > This prevent user to create a single TypeConverter for nullable and non-nullable versions of the same type.

My idea is that drift would do that for users. They just write the non-nullable type converter. If this converter is applied to a nullable column, drift wraps that in a nullable converter behind the scenes.

> We assume user know what he do.

I think I understand the reasoning, but I don't fully agree with it. One of the reasons null safety is a popular Dart feature is that, without deep integration into the type system, dealing with nulls is really hard and error-prone. If we add non-null assertions everywhere and effectively trust the user to know what they're doing, aren't we thwarting null-safety with that?

I'll take a guess and assume that almost all type converters are written in a way that they are declared with non-nullable types. And yet, users currently have to guard for `null` (or use a `NullAwareTypeConverter`) even though their converters were never meant to map `null` values in the first place. I think the current behavior fixes that:

- if you apply a non-nullable type converter to a non-nullable column, it will always be used.
- if you apply a non-nullable type converter to a nullable column, it will be skipped for `null`s (which aren't mapped); it will be used for other values.

With some changes I made in bfad77788e4451633a8999271a48d613991b5493, this now works for compiled queries too (where a non-nullable column could be nullable because it's coming from an `OUTER JOIN` for instance). In my opinion, getting this case right is the most important thing.

westito: That was a good idea :) Perfect! Thanks

PR #1913, opened 2022-06-28
Logging long debug messages

igbenic:
print method does not allow for very long messages, proposed fix by splitting debug messages into 300 character lines


simolus3 (approved):
Thanks!

PR #1915, opened 2022-06-29
Fix non-null JsonTypeConverter

North101:
With the latest fixes to TypeConverters I was getting errors and warnings after running build_runner related to JsonTypeConverter. This PR fixes those errors and removes those warnings

### DIFF
--- a/drift_dev/lib/src/model/column.dart
+++ b/drift_dev/lib/src/model/column.dart
@@ -159,7 +159,8 @@ class MoorColumn implements HasDeclaration, HasType {
     }
 
     // We currently use nullable columns everywhere because it's not clear how
-    // to express nullability in joins otherwise.
+    // to express nullability in joins otherwise. When serializing data with
+    // JsonTypeConverter it needs to match the nullability of the column
     return options.nnbd && (!checkNullable || nullable) ? '$code?' : code;
   }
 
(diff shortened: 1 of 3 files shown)


drift_dev/lib/src/model/column.dart
```diff
@@ -158,7 +160,7 @@ class MoorColumn implements HasDeclaration, HasType {
 
     // We currently use nullable columns everywhere because it's not clear how
     // to express nullability in joins otherwise.
-    return options.nnbd ? '$code?' : code;
+    return options.nnbd && (!checkNullable || nullable) ? '$code?' : code;
```
simolus3, line 163: The comment above is kind of confusing now, let's add a quick note about how it needs needs to be non-null to pass the right type to the serializer.

simolus3 (commented):
Thanks! I just have a minor nit about updating a comment, apart from that this looks good to me.

drift_dev/lib/src/model/column.dart
North101: Updated to:
```
    // We currently use nullable columns everywhere because it's not clear how
    // to express nullability in joins otherwise. When serializing data with
    // JsonTypeConverter it needs to match the nullability of the column
```

simolus3 (approved):

PR #1916, opened 2022-06-30
Change integration tests dialect

westito:
I changed integration tests' dialect to postgres to make compatible with it. It is compatible with sqlite too.


simolus3 (commented):
Thanks for fixing the tests (and the `logStatements` parameter, nice call!).

Still, I think that the examples intended to be used by Flutter developers should still be having `sqlite` as their selected dialect. First, the detailed options like the sqlite3 version and enabled modules only really make sense with that dialect. Also, Flutter users are almost certainly using sqlite as their DBMS, so keeping that in the examples seems like the best call.

westito:
Yes, you're right. I was too diligent 😄

tool/test_all.sh
```diff
@@ -21,6 +22,7 @@ function run_test() {
 function run_test_flutter() {
     echo "------------- Running flutter test: $1 -------------"
     pushd $1 > /dev/null
+    rm -rf .dart_tool
```
simolus3, line 25: Are those necessary? Doing `pub upgrade` and a rebuild should have the same effects as clearing that folder, right? Does this fix anything that I'm overlooking?

westito: Maybe not. It is just my "just for sure" habbit. I can remove this line if you think so

simolus3: Let's remove them, that way rebuilds may be a bit faster. If it turns out to be a problem we can bring them back later.

westito: Okay. I removed

simolus3 (approved):
Thanks!

PR #1921, opened 2022-07-04
More DateTime functions and changes

North101:
Add unixepoch, julianday and strftime DateTime functions.
Allow passing all supported modifiers to date / datetime / time (breaking).

### DIFF
--- a/drift/lib/src/runtime/query_builder/expressions/datetimes.dart
+++ b/drift/lib/src/runtime/query_builder/expressions/datetimes.dart
@@ -10,137 +10,6 @@ const Expression<DateTime> currentDate =
 const Expression<DateTime> currentDateAndTime =
     _CustomDateTimeExpression("strftime('%s', CURRENT_TIMESTAMP)");
 
-/// The "n days" modifiers simply add the specified amount of time to the date
-/// and time specified by the arguments to the left. Note that "±NNN months"
-/// works by rendering the original date into the YYYY-MM-DD format, adding the
-/// ±NNN to the MM month value, then normalizing the result. Thus, for example,
-/// the date 2001-03-31 modified by '+1 month' initially yields 2001-04-31, but
-/// April only has 30 days so the date is normalized to 2001-05-01. A similar
-/// effect occurs when the original date is February 29 of a leapyear and the
-/// modifier is ±N years where N is not a multiple of four.
-DateTimeModifier nDaysModifier(int days) => DateTimeModifier._('$days days');
-
-/// The "n hours" modifiers simply add the specified amount of time to the date
-/// and time specified by the arguments to the left. Note that "±NNN months"
-/// works by rendering the original date into the YYYY-MM-DD format, adding the
-/// ±NNN to the MM month value, then normalizing the result. Thus, for example,
-/// the date 2001-03-31 modified by '+1 month' initially yields 2001-04-31, but
-/// April only has 30 days so the date is normalized to 2001-05-01. A similar
-/// effect occurs when the original date is February 29 of a leapyear and the
-/// modifier is ±N years where N is not a multiple of four.
-DateTimeModifier nHoursModifier(int hours) =>
-    DateTimeModifier._('$hours hours');
-
-/// The "n minutes" modifiers simply add the specified amount of time to the
-/// date and time specified by the arguments to the left. Note that
-/// "±NNN months" works by rendering the original date into the YYYY-MM-DD
-/// format, adding the ±NNN to the MM month value, then normalizing the result.
-/// Thus, for example, the date 2001-03-31 modified by '+1 month' initially
-/// yields 2001-04-31, but April only has 30 days so the date is normalized to
-/// 2001-05-01. A similar effect occurs when the original date is February 29 of
-/// a leapyear and the modifier is ±N years where N is not a multiple of four.
-DateTimeModifier nMinutesModifier(int minutes) =>
-    DateTimeModifier._('$minutes minutes');
-
-/// The "n seconds" modifiers simply add the specified amount of time to the
-/// date and time specified by the arguments to the left. Note that
-/// "±NNN months" works by rendering the original date into the YYYY-MM-DD
-/// format, adding the ±NNN to the MM month value, then normalizing the result.
-/// Thus, for example, the date 2001-03-31 modified by '+1 month' initially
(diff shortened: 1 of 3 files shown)


North101:
Kind of expected this might fail. Its a timezone issue

North101:
Not sure how to fix it

North101:
Done unless you'd like some more tests

simolus3 (commented):
Thanks, I really like this functionality being available in the Dart API.

I'm a bit unsure about exposing modifiers as parameters on the formatting functions though. If one wants to use modifiers, it looks like the unix modifier always needs to come first? The assumption in most parts of drift is that "DateTime in SQL= unix timestamp" and since these methods are extensions on `Expression<DateTime>`, it's always going to have that format.

We might just make the `unixEpochModifier` prefix implicit so that it can't be overridden, but that's not super intuitive either.
I wonder if something like this would work to:

```dart
extension on Expression<DateTime?> {
  Expression<DateTime?> modify(DateTimeModifier modifier) {
    // internally does strftime(%s, this, unixepoch, modifier)
  }
  Expression<DateTime?> modifyThrough(Iterable<DateTimeModifier> modifiers) {
    // internally does strftime(%s, this, unixepoch, ...modifiers)
  }
}
```

Then, if users wanted to print the formatted date three days in the future, they could do something like:

```dart
myDateTime.modify(DateTimeModifier.nDays(3)).datetime
```

Some advantages are:

- getters can stay getters (not a big deal though, we're about to do a breaking release either way)
- it allows the modifiers to be used quite easily when you want the result of modifiers to still be a datetime expression, the current API only works when formatting to strings.
- conceptually, having a transformation like "plus x days" be a function from and to datetimes feels right. There might be cases where you want to apply this transformation in a query without formatting right away (something like `row.expires.isLessThan(currentDateTime.modify(DateTimeModifier.nMonths(1))`)

One disadvantage may be that we're generating less efficient SQL due to nesting those `strftime` calls? I'm not sure if that will be an actual problem.

Either way the API I'm proposing here is just an idea, I'm happy to hear your thoughts about this as well. Maybe it's too verbose like that?

North101:
I also considered doing what you suggested so I'm happy to make those changes.

> I'm a bit unsure about exposing modifiers as parameters on the formatting functions though. If one wants to use modifiers, it looks like the unix modifier always needs to come first? The assumption in most parts of drift is that "DateTime in SQL= unix timestamp" and since these methods are extensions on Expression<DateTime>, it's always going to have that format.

I wasn't sure if having a custom datetime converter with e.g a ISO-8601 string would work with the datetime columns

North101:
I think the API is a lot cleaner / easier to understand with your suggestions though maybe a little less flexible. Before *I think* you could have a column with a custom `DateTime` `TypeConverter` and you could apply the modifiers to make it work with `DateTimeExpressions`. It wasn't very nice, but it was possible. But as you said, `Drift` generally assumes you're using the `unixepoch` format so I'm happy as it is.

Should I comment out / remove `DateTimeModifier.julianDay()` and `const DateTimeModifier.auto()`? I can't see a use for them. Just double checking if i've missed a use case for them.

drift/lib/src/runtime/query_builder/expressions/datetimes.dart
```diff
@@ -107,3 +148,164 @@ class _StrftimeSingleFieldExpression extends Expression<int?> {
...
+  /// The "n seconds" modifiers simply add the specified amount of time to the
+  /// date and time specified by the arguments to the left. Note that "±NNN
+  /// months" works by rendering the original date into the YYYY-MM-DD format,
+  /// adding the ±NNN to the MM month value, then normalizing the result. Thus,
+  /// for example, the date 2001-03-31 modified by '+1 month' initially yields
+  /// 2001-04-31, but April only has 30 days so the date is normalized to
+  /// 2001-05-01. A similar effect occurs when the original date is February 29
+  /// of a leapyear and the modifier is ±N years where N is not a multiple of
+  /// four.
+  DateTimeModifier.seconds(double seconds) : this._('$seconds seconds');
```
simolus3, line 198: Let's use `num` here so that int expressions can be used too.

```suggestion
  DateTimeModifier.seconds(num seconds) : this._('$seconds seconds');
```

simolus3 (commented):
> Should I comment out / remove `DateTimeModifier.julianDay()` and `const DateTimeModifier.auto()`? I can't see a use for them. Just double checking if i've missed a use case for them.

I think we can remove them (and perhaps make `unixepoch` private). They could be useful if we have something like a SQL `parse` function that goes from `Expression<String>` to `Expression<DateTime?>`. Here, being able to specify the format as auto/unixepoch/julianDay kind of makes sense.

> Before I _think_ you could have a column with a custom `DateTime` `TypeConverter` and you could apply the modifiers to make it work with `DateTimeExpressions`

Doing `dartCast<DateTime>()` on something that isn't a unix timestamp has always been pretty dangerous because all the date time expressions expect that format and happily add a `unixepoch` modifier everywhere.

Having a proper parse function would be interesting for users storing datetimes in a format not natively understood by drift.

North101:
> Doing dartCast<DateTime>() on something that isn't a unix timestamp has always been pretty dangerous because all the date time expressions expect that format and happily add a unixepoch modifier everywhere.

One option would be to add `date`, `time`, `datetime`, `unixepoch`, `julianday` to `Expression<int?>` without the implied `unixepoch` modifier.

North101:
Something like:

```dart
/// 
extension IntDateTimeExpressions on Expression<int?> {
  /// Applies modifiers that alters the date and/or time. Each modifier is a
  /// transformation that is applied to the time value to its left. Modifiers
  /// are applied from left to right; order is important.
  Expression<DateTime?> toDateTime(Iterable<DateTimeModifier> modifiers) =>
      FunctionCallExpression('strftime', [
        const Constant<String>('%s'),
        this,
        ...modifiers
      ]);
}

/// 
extension StringDateTimeExpressions on Expression<String?> {
  /// Applies modifiers that alters the date and/or time. Each modifier is a
  /// transformation that is applied to the time value to its left. Modifiers
  /// are applied from left to right; order is important.
  Expression<DateTime?> toDateTime(Iterable<DateTimeModifier> modifiers) =>
      FunctionCallExpression('strftime', [
        const Constant<String>('%s'),
        this,
        ...modifiers
      ]);
}
```

Though I'm happy to consider this done.

PR #1926, opened 2022-07-12
Make `includeJoinedTableColumns` default to false for `selectOnly`

North101:
This always catches me off guard and I feel like it should default to false when using `selectOnly`. Since 2.0 is breaking I think this is a good time to change it, although it could cause bugs as its a change in behaviour.

### DIFF
--- a/drift/CHANGELOG.md
+++ b/drift/CHANGELOG.md
@@ -5,6 +5,8 @@
   It is an error to use a non-nullable type converter on a column that is nullable in
   SQL and vice-versa.
 - __Breaking__: Mapping methods on type converters are now called `toSql` and `fromSql`.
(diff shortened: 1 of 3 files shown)


simolus3 (approved):
Thanks! Yeah IIRC this used to not be configurable and then the parameter had to be `true` by default for backwards compatibility. Setting this to `false` is in the spirit of what `selectOnly` is supposed to do and definitely the better default value.

North101:
On 2nd thought... I wonder if the option should even be there for `selectOnly`. I feel like if you're using it then you never want join columns. Also should it have a `columns` argument?

simolus3:
Yeah good point, I've removed it in 8d840f3eb36c27b11b1fe38c4e8d1bfea286bd5c.

PR #1932, opened 2022-07-19
Add a `const` constructor to non mutable classes

### LINKED ISSUE #1931: Could `Insertable` and generated `DataClass`es have a `const` constructor ?
I have this table 

```dart
class UserTable extends Table {

}
```

The data class generated is:

```dart
class UserTableData extends DataClass implements Insertable<UserTableData> {
  final String id;
  final String name;
  UserTableData({required this.id, required this.name});
 
  // ...
}
```

Could it be a `const` constructor ? 

```dart
class UserTableData extends DataClass implements Insertable<UserTableData> {
  final String id;
  final String name;
  const UserTableData({required this.id, required this.name}); // <- Const constructor
 
  // ...
}
```

You can checkout https://github.com/ValentinVignal/flutter_app_stable/tree/drift/const-data-class

ValentinVignal:
It should fix https://github.com/simolus3/drift/issues/1931


ValentinVignal:
I might need some direction for the tests, in what file/folder should I write it?

Also, after running the tests, in my local, I have those modified files:
- `drift_dev/build/test_cache/build/c075001b96339384a97db4862b8ab8db.cache.dill.track.dill`
- `drift_dev/build/unit_test_assets/AssetManifest.json`
- `drift_dev/build/unit_test_assets/FontManifest.json`
- `drift_dev/build/unit_test_assets/NOTICES.Z`

Should I commit them?

Are there documentation/examples I should updated?

simolus3:
Thanks for your contribution!

> I might need some direction for the tests, in what file/folder should I write it?

I think `drift_dev/test/writer` would be a good place (maybe a new file `data_class_writer_test.dart` in that folder). You can take a look at the `mutable_classes_integration_test.dart` file in the same folder for ideas, but I think for this test we can just run the builder and use something like `contains('const DataClassName(')` as a matcher.

You should also re-run `dart run build_runner build` in `drift/`. The generated files in there are committed to the repository and essentially serve as goldens.

> Should I commit them?

These files don't sound familiar and probably shouldn't be committed. What tool did you use for testing? Running `dart test` in `drift_dev` doesn't create a `build/` folder for me.

kuhnroyal:
Those files look like a `flutter test` run, in any case, the `build` folder should probably be excluded in `.gitignore`.

ValentinVignal:
Thank you for the feedbacks!

I ignored the `build/` folder in https://github.com/simolus3/drift/pull/1932/commits/f440f1c91a8e7f82a4b60dde5b5bf0792bf83bba

I run the build runner in `drift/` in https://github.com/simolus3/drift/pull/1932/commits/d4364306dd296e5dbc9fa101945c2ff1586a527e

And I added a test in https://github.com/simolus3/drift/pull/1932/commits/9bd2486c852b69f7022d2f30034b038940898017

Tell me if I need to do anything else

drift_dev/.gitignore
```diff
@@ -0,0 +1 @@
+build/
```
simolus3: nit: I think it would be nicer to add `*/build/` to the top-level gitignore instead. That way, we get the same behavior in all top-level packages in this repository. Currently, if you ran `flutter test` in `drift/` you'd probably run into the same issue.

drift_dev/test/writer/data_class_writer_test.dart
```diff
@@ -0,0 +1,111 @@
...
+            {'User', 'UsersCompanion'},
+          ),
+        },
+      );
+    },
+    tags: 'analyzer',
+  );
+}
+
+class _GeneratesConstDataClasses extends Matcher {
```
simolus3, line 46: Nice!!

simolus3 (commented):
Thanks! The code changes look good to me, let's just move the gitignore entry one level up.

Also, now that data classes are constants, we have a bunch of lints in `drift` and in the examples. Hopefully they can be fixed with `dart fix`, otherwise you probably need to add a bunch of `const` keywords where the linter complains about them now.

drift_dev/.gitignore
ValentinVignal: I've moved it to the root `.gitignore` in https://github.com/simolus3/drift/pull/1932/commits/1b001e5fa0f8b6aa71a52ba2530891a1ade98e9d

ValentinVignal:
- https://github.com/simolus3/drift/pull/1932/commits/1b001e5fa0f8b6aa71a52ba2530891a1ade98e9d moves the gitignore of `*/build/` inside the root `.gitignore` file.
- https://github.com/simolus3/drift/pull/1932/commits/26b6fc4203b772a0fdeea05ff772866b0f41dde3 adds the missing `const` keywords the linter was complaining about.
- I also fixed some linter info about some lines being too long in https://github.com/simolus3/drift/pull/1932/commits/7134e892c0a08ce401efb357e2e17db49d42c7fe, and I also added more `const` keywords there

simolus3 (approved):
Thanks again for your contribution, this looks good to me now :+1:

Issue #286, opened 2019-12-19
UTC is returned in local time 


Ufuk94:
First of all, thanks for this amazing package and the hard work you put into it! 

I noticed that DateTime objects in UTC are returned in local time. Storing the date in Iso8601 format instead of MillisecondsSinceEpoch should solve this.

simolus3:
> Storing the date in Iso8601 format instead of MillisecondsSinceEpoch should solve this.

That would have been a good idea, but unfortunately it'd be breaking for existing users now :(

As a starting point, you could use a custom type converter.
```dart
class IsoDateTimeConverter extends TypeConverter<DateTime, String> {
  const IsoDateTimeConverter();
  @override
  DateTime mapToDart(String fromDb) {
    if (fromDb == null) {
      return null;
    } else {
      return DateTime.parse(fromDb);
    }
  }

  @override
  String mapToSql(DateTime value) {
    return value?.toIso8601String();
  }
}
```
In your tables, you can now do:
```dart
// replace this
DateTimeColumn foo => dateTime()();
// with this:
TextColumn foo => text().map(const IsoDateTimeConverter())();
```

Arguably, moor should be able to do this out of the box. I'll think about how that could be done without breaking existing databases.

kuhnroyal:
Maybe add a new ColumnType and moor file mapping?

NightOwlCoder:
The problem I see with this approach is that we lose `DateTime` `where` and `orderBy` operations.

JCKodel:
+1 for this.

Suggestion: In MSSQL, there is a special DateTime type called DateTimeOffset (it will record the date with timezone info and all queries will work as intended, no matter if comparing different timezones).

Moor could have a special type called DateTimeOffset or IsoDateTime that always convert the date to UTC while saving (so queries through different timezones will have no issues) and it always returns UTC date (so the developer knows that he must call value.toLocal() before using the date in the UI).

fgblomqvist:
Wouldn't it make more sense if Moor just added a "storeAsString" option to the existing DateTime column rather than having a new type?

Regardless, I think @JCKodel is right in that the best way to avoid timezone issues is just for Moor to always covert to UTC and for the user to convert to the correct timezone when fetching. Doesn't make any sense whatsoever to rely on what timezone the object was stored in imo.

simolus3:
[...]

PR #1934, opened 2022-07-19
Refactor SQL type system

simolus3:
I'm mainly opening this PR to get an overview of all the breaking changes that this refactoring brings, but all kinds of comments or feedback on this are most welcome. 

- Remove the deprecated `SqlType` class and its subtypes.
- To represent the drift sql type at runtime now, an enum class holding supported type mappings is introduced
- In addition to that static enum, a database-specific `SqlTypes` + `DriftDatabaseOptions` class is responsible for actually mapping from and to SQL. This allows adding options, like whether date time values should be stored as text.
- Remove `DataClass.fromData` factory, one needs to use `table.map` instead now.
- Remove nullable `Expression<T?>` instances. This was an attempt to map nullability in SQL to nullability in Dart, but this never really worked due to joins. Now, expressions always have non-nullable types and `read()` returns a nullable value.

Todos for this PR:

- Document breaking changes in changelog
- More tests around type converter nullability

Future work:

- With this refactoring done, there's a straightforward path to finally tackle #286.
- Now that there's an enum with all allowed types, the duplicate logic for this in `drift_dev` can be removed.


North101:
Great! I'm converting my code over now. Is it intended this is no longer supported?
```dart
db.table.column.equals(Null)
```

North101:
How do you use `storeDateTimeAsText`?

simolus3:
Thanks for trying it out! Please let me know if you run into more weird things with the updated nullability now.

> `db.table.column.equals(Null)`

This seems a bit fishy since `x = NULL` is `NULL` in SQL, right? So I'm in favor of not allowing it (it appears as if this has always been broken). I think there could be a `equalsNullable(D?)` method that does `compare == null ? isNull() : equals(compare)` internally (actually I'm surprised something like that doesn't exist already, I think I remember talking about it at some point but I couldn't find anything).

We could also give `equals` the behavior of `equalsNullable`, but I think it may be a bit confusing since nulls in SQL aren't equal?

> How do you use `storeDateTimeAsText`?

It's not supported yet, I just added it as a proof-of-concept while refactoring the types to find all the locations that need changing. I'm working on it, I think it's best exposed as a build option (since the generator needs to be aware of it for type inference in custom queries).

North101:
> Thanks for trying it out! Please let me know if you run into more weird things with the updated nullability now.
> 
> > `db.table.column.equals(Null)`
> 
> This seems a bit fishy since `x = NULL` is `NULL` in SQL, right? So I'm in favor of not allowing it (it appears as if this has always been broken). I think there could be a `equalsNullable(D?)` method that does `compare == null ? isNull() : equals(compare)` internally (actually I'm surprised something like that doesn't exist already, I think I remember talking about it at some point but I couldn't find anything).
> 
> We could also give `equals` the behavior of `equalsNullable`, but I think it may be a bit confusing since nulls in SQL aren't equal?

I thought so and I think it'll be less of a gotcha but thought i'd bring it up as I was using this to intentionally have nulls not return results. I'll just fix my code :P

> > How do you use `storeDateTimeAsText`?
> 
> It's not supported yet, I just added it as a proof-of-concept while refactoring the types to find all the locations that need changing. I'm working on it, I think it's best exposed as a build option (since the generator needs to be aware of it for type inference in custom queries).

👍

PR #1938, opened 2022-07-21
Option to store `DateTime` values as text

### LINKED ISSUE #286: UTC is returned in local time

simolus3:
This is not nearly done yet, but this PR's description can serve as a to-do list for me :)

Todos:

- [x] Add build option to store date time values as text for a database
- [x] Support this build option in the generator
  - Infer these columns as text for SQL.
  - Respect a `IsDateTime` type-hint on text values during type inference.
- [x] Respect this option when mapping literals and variables to SQL.
- [x] Adapt the datetime APIs as extension on `Expression<DateTime>` to be aware of this option. 
- [x] Run all existing date-time tests against both modes if applicable
- [x] Documentation
  - Document build option
  - Document storing behavior in page about supported types
  - Document how to migrate between the two storage types (and test the migration snippet!)
- [x] Lint for things like `dateTime1 < dateTime2` in SQL queries when the text mode is active.

Finally fixes https://github.com/simolus3/drift/issues/286.

### DIFF
--- a/docs/lib/snippets/migrations/datetime_conversion.dart
+++ b/docs/lib/snippets/migrations/datetime_conversion.dart
@@ -1,0 +1,74 @@
+import 'package:drift/drift.dart';
+
+extension MigrateToTextDateTimes on GeneratedDatabase {
+  // #docregion unix-to-text
+  Future<void> migrateFromUnixTimestampsToText(Migrator m) async {
+    for (final table in allTables) {
+      final dateTimeColumns =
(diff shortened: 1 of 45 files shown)


North101:
Awesome! I'll give the migration a go

PR #1958, opened 2022-08-01
Remove null check when parameter is not nullable in dart

westito:
Képernyőfotó 2022-08-01 - 11 48 46

This was a problem before, I don't remember exactly. But it appears again. When TypeConverter returns non-nullable type and column is nullable, Drift writes null check conditions


simolus3 (approved):
Nice catch, thanks!

simolus3:
Ah, looks like you need to re-run the build in `drift/` as well.

westito:
Sorry for being late. I made a docker test env to run local tests. I can't test with the new SQLite version anymore. It isn't possible to update sqlite on Mac that dart executable use. 
There is some web issues with new date functions. Related to floating numbers I think.
Képernyőfotó 2022-08-08 - 11 24 09
Képernyőfotó 2022-08-08 - 11 25 21
Képernyőfotó 2022-08-08 - 11 25 34

simolus3:
> I can't test with the new SQLite version anymore.

Drift's tests should still work with older sqlite3 versions ideally (tests requiring newer versions should be skipped on older versions). If you can post the tests that are failing with your sqlite install I can skip them on older versions.

>  There is some web issues with new date functions. Related to floating numbers I think.

Yeah, I think this is because `DateTime` uses timestamps internally and those are doubles on the web. I have skipped or adapted those tests in eea63cb5c9d93ae66fc0fd378b82a70bc481883a.

westito:
Képernyőfotó 2022-08-08 - 14 27 02

simolus3:
I've fixed that test in 4856ba4c5b9542e8d7d12b750c75cf246fc2c3dc, but I agree that having a Docker image to test with the latest sqlite3 version is still helpful.

simolus3 (approved):
Looks good to me now, nice catch!

PR #1959, opened 2022-08-01
Fix lint warnings when strict type check enabled

westito:
Képernyőfotó 2022-08-01 - 12 15 30


drift_dev/lib/src/writer/database_writer.dart
```diff
@@ -117,8 +117,9 @@ class DatabaseWriter {
     // Write List of tables
     final schemaScope = dbScope.leaf();
     schemaScope
-      ..write('@override\nIterable<TableInfo> get allTables => ')
-      ..write('allSchemaEntities.whereType<TableInfo>();\n')
+      ..write(
+          '@override\nIterable<TableInfo<Table, dynamic>> get allTables => ')
+      ..write('allSchemaEntities.whereType<<Table, dynamic>>();\n')
```
simolus3: I think this should be

```suggestion
      ..write('allSchemaEntities.whereType<TableInfo<Table, Object?>>();\n')
```

(using `Object?` over `dynamic` is a preference choice here, either way is fine with me)

simolus3 (approved):
Looks good now, thanks!

PR #1990, opened 2022-08-15
Prevent additional checks from failing if nullable value is null

davidhsoftware:
If `isAcceptableValue` is run on a nullable column with an additional check it will fail for null values, e.g.

`TextColumn get description => text().withLength(max: 250).nullable()();`

This behaviour is different from previous versions of Drift

### DIFF
--- a/drift/CHANGELOG.md
+++ b/drift/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 2.0.1
+
+- Fix an error when inserting a null value into a nullable column defined with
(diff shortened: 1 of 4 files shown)


simolus3:
Actually it looks like the `!` was completely unnecessary since the `additionalChecks` function is defined to take a nullable parameter. I've just removed it and added a test which should also fix this.

Merged and released as `2.0.1`, thanks for your contribution!

PR #1999, opened 2022-08-17
Fix the wrong version number on `migration_test`

jack24254029:
It should use `oldVersion` not use `1` on every times.


simolus3 (approved):
Good catch, thanks for the fix!

PR #2000, opened 2022-08-17
Update binary_string_conversion.dart

3alisaki:
string.codeUnits returns 16 bit list

Note: why not just use:
```
import 'dart:convert';
...
utf8.encode(input)
....
utf8.decode(input)
```

### DIFF
--- a/drift/lib/src/web/binary_string_conversion.dart
+++ b/drift/lib/src/web/binary_string_conversion.dart
@@ -22,7 +22,7 @@ class _String2Bin extends Converter<String, Uint8List> {
   @override
   Uint8List convert(String input) {
     final codeUnits = input.codeUnits;
-    final list = Uint8List(codeUnits.length);
+    final list = Uint16List(codeUnits.length);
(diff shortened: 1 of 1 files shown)


simolus3:
The intention here is that, since we can't store bytes in local storage, we store them as an ascii string (so each character has a code between 0 and 255). We _want_ to store a `Uint8List` because that's the way the database bytes are represented.

The current code is a Dart port of what SQL.js recommends: https://github.com/sql-js/sql.js/wiki/Persisting-a-Modified-Database#save-a-database-to-a-string

As to why we're not using `utf8`: Not every sequence of bytes is a valid utf8 encoding that could be decoded to a string. We could probably get away with using `ascii` (or `latin1`?) as a codec if we allow invalid chars, I'll try that one.

simolus3:
I've migrated to `latin1` in 9c30acf8791b0f5aaf91677a8ee2c99c1864d625, thanks for the suggestion!

PR #2001, opened 2022-08-18
typedef scoped functions

North101:
This doesn't currently deal with duplicate names. I wasn't sure how to handle that or you would want things named (if you want this change at all).

### DIFF
--- a/drift/test/generated/custom_tables.g.dart
+++ b/drift/test/generated/custom_tables.g.dart
@@ -1605,7 +1605,7 @@ abstract class _$CustomTablesDb extends GeneratedDatabase {
   }
 
   Selectable<Config> readMultiple(List<String> var1,
-      {ReadMultipleClause? clause}) {
+      {ReadMultiple$clause? clause}) {
     var $arrayStartIndex = 1;
     final expandedvar1 = $expandVar($arrayStartIndex, var1.length);
     $arrayStartIndex += var1.length;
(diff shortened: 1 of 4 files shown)


North101:
I think `typeFor` needs a name change to imply that its writing the `typedef`.

Maybe `writeTypeFor`?

simolus3:
I like the idea of using typedefs.

Regarding the public defaults: The only reason I originally extracted the defaults into top-level methods is because the reference to them needs to be constant in the default value, so we can't use anonymous functions. I'm not sure if they _should_ be part of the public API. If your intention was to pass the default around (so that one can write wrapper functions with the same default), perhaps we should just make the parameter nullable and default to the anonymous function with `??`. Then you can pass `null` if you want the default. Would that solve the problem you had with private defaults?

North101:
> I like the idea of using typedefs.
> 
> Regarding the public defaults: The only reason I originally extracted the defaults into top-level methods is because the reference to them needs to be constant in the default value, so we can't use anonymous functions. I'm not sure if they _should_ be part of the public API. If your intention was to pass the default around (so that one can write wrapper functions with the same default), perhaps we should just make the parameter nullable and default to the anonymous function with `??`. Then you can pass `null` if you want the default. Would that solve the problem you had with private defaults?

Yea I think that would solve it. I was originally thinking that you could pass a `null` as a normal value but I think it must be a function for scoped functions so a null would work.

simolus3:
Are you currently working on using `null` as a default? If you want to push your current state I'm happy to take this over as well, then I can get it released with some other fixes today :)

North101:
Im busy until Sunday unfortunately

simolus3 (approved):
> Im busy until Sunday unfortunately

No worries, thanks for your contribution! I have changed the generated code to use nullable parameters instead. I've also changed the name of the generated typedefs to `<QueryName>$<parameterName>` to make them easier to distinguish. I'm not entirely sure on  that naming scheme since it's weird Dart, but I think there should be some way to tell the query and the actual parameter apart in the type.

PR #2013, opened 2022-08-23
Support for spellfix1 module

FaFre:
This PR adds support for spellfix1 (https://www.sqlite.org/spellfix1.html)

There is one last challenge, spellfix uses hidden columns as input parameters, which are only used to control the behavior of the module (https://www.sqlite.org/spellfix1.html#virtual_table_details). This columns are never included in the output.
```
top
    (HIDDEN) For any query, this value is the same on all rows. It is an integer which is the maximum number of rows that will be output. The actually number of rows output might be less than this number, but it will never be greater. The default value for top is 20, but that can be changed for each query by including a term of the form "top=N" in the WHERE clause of the query. 

scope
    (HIDDEN) For any query, this value is the same on all rows. The scope is a measure of how widely the virtual table looks for matching words. Smaller values of scope cause a broader search. The scope is normally chosen automatically and is capped at 4. Applications can change the scope by including a term of the form "scope=N" in the WHERE clause of the query. Increasing the scope will make the query run faster, but will reduce the possible corrections. 

srchcnt
    (HIDDEN) For any query, this value is the same on all rows. This value is an integer which is the number of words examined using the edit-distance algorithm to find the top matches that are ultimately displayed. This value is for diagnostic use only. 

soundslike
    (HIDDEN) When inserting vocabulary entries, this field can be set to a spelling that matches what the word sounds like. See the DEALING WITH UNUSUAL AND DIFFICULT SPELLINGS section below for details. 

command
    (HIDDEN) The value of the "command" column is always NULL. However, applications can insert special strings into the "command" column in order to provoke certain behaviors in the spellfix1 virtual table. For example, inserting the string 'reset' into the "command" column will cause the virtual table to reread its edit distance weights (if there are any). 
```

Right now we get a warning from the code generator for this columns, because I haven't mapped them yet. But the code seems to get generated anyways.

How should this get handled? 
Adding a new 'hidden' flag to `TableColumn` and treat them accordingly would be an option I could think of.

After this is clear I will continue and also add the tests.


simolus3:
Hidden columns are only used as parameters when a table is queried, right? I think we can mark them as hidden on the `TableColumn` as you suggested. The `CreateTableReader` in `drift_dev` should then ignore those columns and not generate a `DriftColumn` for them.

FaFre:
Maybe, after https://github.com/simolus3/sqlite3.dart/pull/115 is merged, we can extend the `NativeDatabase` constructor to take a value of `LoadableExtension` (https://github.com/simolus3/sqlite3.dart/pull/115/files#diff-95ba3f31fc280ad6ec6309330ea763a440b6419b8955a316921f31a8dcef9b71R510) to register extensions automatically. It is safe to call `sqlite3_auto_extension` multiple times. It will do nothing when the extension is already registered (we do not need to track the state).

EDIT: Nvm, I added the loading to `Database` directly. The extension can be loaded now like this:
```dart
NativeDatabase(
      file,
      setup: (database) {
        database.ensureExtensionLoaded(LoadableExtension.spellfix1);
      },
    );
```
Really sleek would be doing this with the generator directly, but after a quick check that would be to much work for now.

simolus3 (approved):
Thanks for your contribution!

PR #2017, opened 2022-08-25
Expose fts5 rowid

FaFre:
Expose `rowid` of fts5 table when created with external content table (https://www.sqlite.org/fts5.html#external_content_tables).
We need the `rowid` for every insert here, and probably also for other queries.

### DIFF
--- a/sqlparser/lib/src/engine/module/fts5.dart
+++ b/sqlparser/lib/src/engine/module/fts5.dart
@@ -28,6 +28,8 @@ class _Fts5Module extends Module {
     return _Fts5Table(
       name: stmt.tableName,
       columns: [
+        if (stmt.argumentContent.any((arg) => arg.startsWith('content_rowid')))
+          TableColumn('rowid', const ResolvedType(type: BasicType.int)),
         for (var arg in columnNames)
           TableColumn(arg, const ResolvedType(type: BasicType.text)),
       ],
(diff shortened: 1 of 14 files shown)


sqlparser/lib/src/engine/module/fts5.dart
```diff
@@ -28,6 +28,8 @@ class _Fts5Module extends Module {
     return _Fts5Table(
       name: stmt.tableName,
       columns: [
+        if (stmt.argumentContent.any((arg) => arg.startsWith('content_rowid')))
+          TableColumn('rowid', const ResolvedType(type: BasicType.int)),
```
simolus3: I'm not sure why the `rowid` only exists when there's a `content_rowid` option? Shouldn't it be exposed whenever we have a content table in general?

simolus3:
I've also added some static analysis in drift to make sure it creates the content table before the fts5 table and to warn if an unknown table is referenced.

I'm still not sure when the `rowid` column should be available or not. Even for fts5 tables that don't have a content table, I can do `SELECT rowid FROM <thatTable>`. I can also use `rowid` as a column for inserts in that case. Whether that's supposed to work or not I don't know :D

sqlparser/lib/src/engine/module/fts5.dart
FaFre: At first I got the documentation wrong, I double checked and validated that `content_rowid` is actually optional and defaults to 'rowid'

FaFre:
Great :D
After your last comment I read the documentation again and just pushed another commit. It catches empty `content` option, and fallsback to `rowid` when `content_rowid` is not set. Input strings can contain ticks and are also getting cleaned now.

simolus3 (approved):
Looks good to me now, thank you!

PR #2023, opened 2022-08-27
Update existing_databases.md

daniel-hopkins:
the line `final file = File(p.join(dbFo …` needs `path/path.dart` imported as `p`


simolus3 (approved):
Good catch, thank you!

PR #2024, opened 2022-08-28
Add `NULLS FIRST\LAST` to the `OrderingTerm`

Vorkytaka:
SQLite 3.30.0 added the NULLS FIRST and NULLS LAST options to the ORDER BY clause.

For now we need to use some hacks to sort nulls as we want, like use `ifnull` or `coalesce`.
So, i'm think, that it's good to add this new feature, to handle nulls in a good way. :)

I'm not sure how to handle version check, but, as i see, in drift you just warning about version in comments\docs, so i did the same.

### DIFF
--- a/drift/lib/src/runtime/query_builder/components/order_by.dart
+++ b/drift/lib/src/runtime/query_builder/components/order_by.dart
@@ -3,30 +3,28 @@ part of '../query_builder.dart';
 /// Describes how to order rows
 enum OrderingMode {
   /// Ascending ordering mode (lowest items first)
-  asc,
+  asc._('ASC'),
 
   /// Descending ordering mode (highest items first)
-  desc
-}
+  desc._('DESC');
 
-const _modeToString = {
-  OrderingMode.asc: 'ASC',
-  OrderingMode.desc: 'DESC',
-};
+  final String _mode;
+
+  const OrderingMode._(this._mode);
+}
 
 /// Describes how to order nulls
 enum NullsOrder {
   /// Place NULLs at the start
-  first,
+  first._('NULLS FIRST'),
 
   /// Place NULLs at the end
-  last,
-}
+  last._('NULLS LAST');
 
-const _nullsOrderToString = {
-  NullsOrder.first: 'NULLS FIRST',
-  NullsOrder.last: 'NULLS LAST',
-};
+  final String _order;
+
+  const NullsOrder._(this._order);
+}
 
 /// A single term in a [OrderBy] clause. The priority of this term is determined
 /// by its position in [OrderBy.terms].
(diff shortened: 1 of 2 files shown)


drift/lib/src/runtime/query_builder/components/order_by.dart
```diff
@@ -23,9 +37,21 @@ class OrderingTerm extends Component {
...
+  final NullsOrder? nullsOrder;
+
   /// Creates an ordering term by the [expression] and the [mode] (defaults to
   /// ascending).
-  OrderingTerm({required this.expression, this.mode = OrderingMode.asc});
+  OrderingTerm({
+    required this.expression,
+    this.mode = OrderingMode.asc,
+    this.nullsOrder,
+  });
```
simolus3: The name of the field is fine, but I think it would read nicer if the name of the parameter was `nulls`:

```suggestion
    NullsOrder? nulls,
  }): nullsOrder = nulls;
```

That way, users can write `OrderingTerm(..., nulls: NullsOrder.last)` which is a bit closer to the original SQL :)

drift/lib/src/runtime/query_builder/components/order_by.dart
```diff
@@ -14,6 +14,20 @@ const _modeToString = {
...
+  first,
+
+  /// Place NULLs at the end
+  last,
+}
+
+const _nullsOrderToString = {
+  NullsOrder.first: 'NULLS FIRST',
+  NullsOrder.last: 'NULLS LAST',
+};
```
simolus3: Now that we're targeting Dart 2.17, you can actually move this information into the enum:

```dart
enum NullsOrder {
  first._('NULLS FIRST'),
  last._('NULLS LAST');

  final String _sql;
  const NullsOrder._(this._sql);
}
```

drift/lib/src/runtime/query_builder/components/order_by.dart
```diff
@@ -23,9 +37,21 @@ class OrderingTerm extends Component {
...
   /// ascending).
-  OrderingTerm({required this.expression, this.mode = OrderingMode.asc});
+  OrderingTerm({
+    required this.expression,
+    this.mode = OrderingMode.asc,
+    this.nullsOrder,
+  });
 
   /// Creates an ordering term that sorts for ascending values of [expression].
   factory OrderingTerm.asc(Expression expression) {
```
simolus3: I think it would be great if the same named parameter could be added to the `asc` and `desc` factories as well.

drift/test/database/statements/order_by_test.dart
```diff
@@ -0,0 +1,78 @@
...
+  setUp(() {
+    executor = MockExecutor();
+    db = TodoDb(executor);
+  });
+
+  test('when nullsOrder is null it ignored', () async {
+    final query = db.select(db.users);
+    query.orderBy([(tbl) => OrderingTerm(expression: tbl.name)]);
+    await query.get();
+    verify(executor.runSelect(
```
simolus3, line 21: Thanks for these tests! By the way, if you only wanted to test for the SQL generated by an `OrderingTerm`, you could also do `expect(OrderingTerm(expression: tbl.name), generates('name ASC'))`. But doing it like this is fine too, it's always good to test the thing in a full and working context :)

simolus3 (commented):
Thanks for your contribution. Let's also add this API to the `.asc` and `.desc` helper factories, apart from that I just have nitpicked comments.

drift/lib/src/runtime/query_builder/components/order_by.dart
Vorkytaka: Yeah, I was thinking about it, but stay with one code-style.

Vorkytaka:
Pushed new commit with fixes

simolus3 (approved):
LGTM now, thanks for your contribution!

PR #2025, opened 2022-08-29
Prepare for separating ClassElement, EnumElement, and MixinElement.

scheglov:
All three implement InterfaceElement, but in the breaking change for analyzer we will stop implementing ClassElement in EnumElement and MixinElement. This is a minimal change necessary to make google3 green, the might be need for more changes that are not exercised there.


simolus3 (approved):
Thanks! Since you need it in google3 soon, I assume I should also cherry-pick this to `latest-release`?

scheglov:
This sounds reasonable to me, I was told that only the stable version is rolled.

PR #2037, opened 2022-09-07
docs: fix upgrading doc (toDart -> fromSql)

AyumiSashitani:
Looking at `drift/lib/src/runtime/types/converters.dart`, I think it's `fromSql()`, not `toDart()`.
Therefore, I have modified the doc.


simolus3 (approved):
Good catch, thanks!

PR #2044, opened 2022-09-12
Generated functions for returning queries return Iterable<_> instead of List<_>

LeFrosch:
After I upgraded to version 2.1.0 I found the following error:
```
 Error: A value of type 'Iterable<int>' can't be returned from a function with return type 'FutureOr<List<int>>'.
 - 'Iterable' is from 'dart:core'.
 - 'List' is from 'dart:core'.
        }).then((rows) => rows.map((QueryRow row) => row.read<int>('id')));
```

This error is caused by the generated code from returning queries. For example this simple query would generate the following code: 
```
query: INSERT INTO tbl (id) VALUES(10) RETURNING id;
```
```dart 
Future<List<TblData>> query() {
  return customWriteReturning(
    'INSERT INTO tbl (id) VALUES (10) RETURNING id',
    variables: [],
    updates: {tbl}
  ).then((rows) => Future.wait(rows.map(tbl.mapFromRow)));
}
```
And because the `map` function returns an iterable instead of a list the types are not compatible. To fix this I just added a `toList`call in this pull request.

Hope this is helpful :)

### DIFF
--- a/drift_dev/lib/src/writer/queries/query_writer.dart
+++ b/drift_dev/lib/src/writer/queries/query_writer.dart
@@ -218,11 +218,11 @@ class QueryWriter {
     if (update.needsAsyncMapping) {
       _buffer.write('Future.wait(rows.map(');
       _writeMappingLambda(update);
-      _buffer.write('))');
+      _buffer.write(')).toList()');
     } else {
       _buffer.write('rows.map(');
       _writeMappingLambda(update);
-      _buffer.write(')');
+      _buffer.write(').toList()');
     }
     _buffer.write(');\n}');
   }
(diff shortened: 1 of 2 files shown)


drift_dev/lib/src/writer/queries/query_writer.dart
```diff
@@ -218,11 +218,11 @@ class QueryWriter {
     if (update.needsAsyncMapping) {
       _buffer.write('Future.wait(rows.map(');
       _writeMappingLambda(update);
-      _buffer.write('))');
+      _buffer.write(')).toList()');
```
simolus3: I think we don't need it in this case since `Future.wait` takes an iterable and returns a list already.

LeFrosch: Oh yes that's true.

simolus3 (approved):
Thanks for the fix!

PR #2076, opened 2022-09-24
a new app to show implementation with GetX

abdelrahmanelmarakby:


simolus3 (approved):
Thanks for sharing the example!

Issue #2073, opened 2022-09-23
Drift can't resolve to the latest `analyzer` release


Carapacik:
After flutter pub get:

``` 
Because freezed >=2.1.1 depends on analyzer ^5.0.0 and drift_dev 2.1.0 depends on analyzer ^4.5.0, freezed >=2.1.1 is incompatible with drift_dev 2.1.0.
And because no versions of drift_dev match >2.1.0 <3.0.0, freezed >=2.1.1 is incompatible with drift_dev ^2.1.0.
So, because myproject depends on both drift_dev ^2.1.0 and freezed ^2.1.1, version solving failed.
pub get failed (1; So, because myproject  depends on both drift_dev ^2.1.0 and freezed ^2.1.1, version solving failed.)
A clear and concise description of what the bug is. Please briefly describe what behavior you expected, what behavior you saw, and how the bug can be reproduced. 
```

Maybe you should change the analyzer version from `"^4.5.0"` to `^4.5.0:`
https://github.com/simolus3/drift/blob/8c1a62ae1b65373d1403d80569b67cc959394d8b/drift_dev/pubspec.yaml#L33
You can also upgrade the analyzer to `^4.7.0`.

And after analyzer version 4.2.0 package support only on Dart higher than 2.17:
https://github.com/simolus3/drift/blob/8c1a62ae1b65373d1403d80569b67cc959394d8b/drift_dev/pubspec.yaml#L9

Carapacik:
If you are busy, then I can open a PR with this fix.

Carapacik:
My pubspec.yaml
```yaml
environment:
  sdk: '>=2.17.0 <3.0.0'

dependencies:
  flutter:
    sdk: flutter
  freezed_annotation: ^2.1.0
  json_annotation: ^4.7.0
  drift: ^2.1.0
  sqlite3_flutter_libs: ^0.5.10
  path_provider: ^2.0.11
  path: ^1.8.2

dev_dependencies:
  flutter_lints: ^2.0.1
  build_runner: ^2.2.1
  freezed: ^2.1.1
  json_serializable: ^6.4.0
  drift_dev: ^2.1.0
```

simolus3:
I think we're not affected by any of the breaking changes in `analyzer` version 5.x, so we can raise our analyzer dependency to also support `5.x` releases. Feel free to open a PR for that if you want to.

[...]

PR #2084, opened 2022-09-26
[drift_dev] Update analyzer and deprecated methods

Carapacik:
Also updated the minimum version of packages so that packages have null safety support. (#2073)

### DIFF
--- a/drift_dev/pubspec.yaml
+++ b/drift_dev/pubspec.yaml
@@ -30,7 +30,7 @@ dependencies:
   sqlparser: ^0.23.2
 
   # Dart analysis
-  analyzer: '>=4.7.0 <6.0.0'
+  analyzer: '^5.0.0'
   analyzer_plugin: ^0.11.0
   source_span: ^1.5.5
   package_config: ^2.0.0
--- a/drift_dev/pubspec.yaml
+++ b/drift_dev/pubspec.yaml
@@ -1,6 +1,6 @@
 name: drift_dev
 description: Dev-dependency for users of drift. Contains a the generator and development tools.
-version: 2.1.0
+version: 2.2.0-dev
 repository: https://github.com/simolus3/drift
 homepage: https://drift.simonbinder.eu/
 issue_tracker: https://github.com/simolus3/drift/issues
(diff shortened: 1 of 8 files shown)


Carapacik:
@simolus3, should I upgrade the drift_dev version to 2.1.1?

drift_dev/pubspec.yaml
```diff
@@ -6,51 +6,51 @@ homepage: https://drift.simonbinder.eu/
 issue_tracker: https://github.com/simolus3/drift/issues
 
 environment:
-  sdk: '>=2.14.0 <3.0.0'
+  sdk: '>=2.17.0 <3.0.0'
```
simolus3, line 9: > should I upgrade the drift_dev version to 2.1.1?

Actually it should be `2.2.0-dev`, there have been some new features in the meantime. I will strip the `-dev` before releasing, but you can also leave the version as it is.

In general, I don't like raising the minimum dependency constraints if we don't need it. It's fine for the language version I guess, but do we really need a new feature that has been added in `charcode` version `1.3.1`? If we support the lower version, our pubspec should indicate that as well. We want to test against the latest versions we claim to support, which is why the CI runs `pub upgrade`. There's no need to _only_ support the latest version by raising the constraint in the pubspec.

So, can we keep the constraints of the other packages? Given that we don't seem to be affected by any breaking changes in `analyzer`, I suggest doing this:

```dart
  analyzer: '>=4.5.0 < 6.0.0'
  analyzer_plugin: '^0.11.0'
```

If the lower bound has been wrong before and needed to be upgraded - that's fine! But I doubt we need to raise it to the latest version for every single dependency.

simolus3:
Also, don't worry about the breaking CI for now. I'll have to fix that dependency issue.

Carapacik:
Ok, I can fix version and some packages

Carapacik:
@simolus3, to resolve conflicts in #2073, you need an analyzer version higher than 4.7.0. And the analyzer from version 4.2.0 already supports the minimum Dart SDK 2.17.

simolus3:
Now it seems that to resolve the lints I need to use `5.0.0` as a minimum version :D Oh well...

Carapacik:
I think it is necessary to update all dependencies in all projects to Dart 2.17. Because analyzer 5 requires this.

Carapacik:
Maybe you should also ask to update the analyzer in build_value_generator?

In the meantime, you can just write 
```yaml
dependency_overrides:
  analyzer: ^5.0.0
```

Carapacik:
@simolus3, check last commit here

simolus3:
> I think it is necessary to update all dependencies in all projects to Dart 2.17. Because analyzer 5 requires this.

I'm not sure I don't understand what you mean with this. If a package has `sdk: '>=2.12.0 <3.0.0'` in its pubspec, it can still be used with Dart 2.17. So yeah, if we depend on `analyzer: ^5.0.0`, the user will have to be on Dart 2.17 or later, but IMO there's no need to spell that out in the pubspec.

examples/with_built_value/pubspec.yaml
```diff
@@ -3,13 +3,16 @@ publish_to: none
...
   built_value: ^8.0.0
 
 dev_dependencies:
   drift_dev: ^2.0.0-dev
-  built_value_generator: ^8.1.1
+  built_value_generator: ^8.1.3
   build_runner: ^2.0.0
+
+dependency_overrides:
+  analyzer: ^5.0.0
```
simolus3, line 20: Could you add a comment here explaining why the override is necessary? Something like

```suggestion
dependency_overrides:
  # `built_value_generator` doesn't support the latest analyzer yet
  analyzer: ^5.0.0
```

Carapacik:
I added a description to override

simolus3 (approved):
LGTM now, thanks for your contribution!

PR #2086, opened 2022-09-30
Add `const` constructor to `Insertable`

ValentinVignal:
Relates to https://github.com/simolus3/drift/issues/1931

It was not done by https://github.com/simolus3/drift/pull/1932

This allows the dev to do:

```dart
class MyInsertable<T> extends Insertable<T> {
  const MyInsertable(); // <- Const constructor.

  // ... 
}
```

I'm not too sure about the test I have written, tell me if you want me to do it in another way, or move the test somewhere else


simolus3:
> I'm not too sure about the test I have written, tell me if you want me to do it in another way, or move the test somewhere else

I think a simple test like that is fine, we just need to fix the analysis warnings. Can you call `toColumns` or something on that constant instance to use it? Also, perhaps add a comment explaining that the test exists to remember that the constructor has to be const.

And the test should probably be put into the existing `drift/test/database/data_class_test.dart` file.

ValentinVignal:
Thanks @simolus3 , I've moved the test and fixed the lints in https://github.com/simolus3/drift/pull/2086/commits/0a7b32f92e7ff0d502063613547f7e59d36af67d

I also added some `const` keywords in https://github.com/simolus3/drift/pull/2086/commits/7f78ab46421b3f5a5ca599930e1eff9398a95e84

simolus3 (approved):

PR #2102, opened 2022-10-13
Decouple JsonTypeConverter from TypeConverter

North101:
WIP

### DIFF
--- a/drift/lib/src/runtime/types/converters.dart
+++ b/drift/lib/src/runtime/types/converters.dart
@@ -29,12 +29,11 @@ abstract class TypeConverter<D, S> {
 /// A mixin for [TypeConverter]s that should also apply to drift's builtin
 /// JSON serialization of data classes.
 ///
-/// By default, a [TypeConverter] only applies to the serialization from Dart
-/// to SQL (and vice-versa).
-/// When a [BuildGeneralColumn.map] column (or a `MAPPED BY` constraint in
-/// `.drift` files) refers to a type converter that inherits from
-/// [JsonTypeConverter2], it will also be used for the conversion from and to
-/// JSON.
+/// Unlike the old [JsonTypeConverter] mixin, this more general mixin allows
+/// using a different type when serializing to JSON ([J]) than the type used in
+/// SQL ([S]).
+/// For the cases where the JSON serialization and the mapping to SQL use the
+/// same types, it may be more convenient to mix-in [JsonTypeConverter] instead.
 mixin JsonTypeConverter2<D, S, J> on TypeConverter<D, S> {
   /// Map a value from the Data class to json.
   ///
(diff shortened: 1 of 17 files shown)


North101:
Ok I think I have this working now. I tried to keep it backwards compatible but I wasn't able to without doing a lot more changes I wasn't comfortable.

North101:
Nevermind, found a few issues

drift/lib/src/runtime/types/converters.dart
```diff
@@ -35,25 +35,30 @@ abstract class TypeConverter<D, S> {
 /// `.drift` files) refers to a type converter that inherits from
 /// [JsonTypeConverter], it will also be used for the conversion from and to
 /// JSON.
-mixin JsonTypeConverter<D, S> on TypeConverter<D, S> {
+abstract class JsonTypeConverter<D, S, J> extends TypeConverter<D, S> {
```
simolus3: One way to make this non-breaking might be to introduce this as a new superclass (imagine `JsonTypeConverterWithDifferentTypes` but with a less cumbersome name).

Them, `typealias JsonTypeConverter<D, S> = JsonTypeConverterWithDifferentTypes<D, S, S>;` would make this non-breaking I believe? Or just `class JsonTypeConverter<D, S> extends JsonTypeConverterWithDifferentTypes<D, S, S>`. I might still be overlooking something, but maybe that works to make this non-breaking.

simolus3 (commented):
I like the idea and the implementation in the generator.

I have an idea to make this non-breaking (but I didn't fully think this through, might not be entirely sound). Still, I think it could be worth trying out.

drift/lib/src/runtime/types/converters.dart
North101: I'll give this a go later

North101: I think it's backwards compatible now. Thanks for the suggestion!

Not sure what to name `JsonTypeConverterWithDifferentTypes`, it seems a little verbose.

drift/lib/src/runtime/types/converters.dart
```diff
@@ -33,17 +33,47 @@ abstract class TypeConverter<D, S> {
 /// to SQL (and vice-versa).
 /// When a [BuildGeneralColumn.map] column (or a `MAPPED BY` constraint in
 /// `.drift` files) refers to a type converter that inherits from
-/// [JsonTypeConverter], it will also be used for the conversion from and to
+/// [JsonTypeConverterWithDifferentTypes], it will also be used for the conversion from and to
 /// JSON.
-mixin JsonTypeConverter<D, S> on TypeConverter<D, S> {
+mixin JsonTypeConverterWithDifferentTypes<D, S, J> on TypeConverter<D, S> {
```
simolus3: Hard to think of a better name here... Maybe `BaseJsonTypeConverter` could work?

drift/lib/src/runtime/types/converters.dart
simolus3: Nice!

drift/lib/src/runtime/types/converters.dart
North101: Maybe just `JsonTypeConverter2`?

simolus3: Sounds good 👍

simolus3 (approved):
Thanks for your contribution!

North101:
> Thanks for your contribution!

No problem. Thanks!

Maybe in the next breaking release we could rename:
`JsonTypeConverter2` to `JsonTypeConverter`
`JsonTypeConverter` to `SimpleJsonTypeConverter`

PR #2107, opened 2022-10-18
fix DatabaseConnection invoke

saliouseck2009:
if we call DatabaseConnection( ) it need 3 parameters by default so the compiler report this error . So we need to use the fromExecutor constructor if we want to give it one parameter.


simolus3:
The constructor recently changed to only take the executor, this change has already been released as well:

https://github.com/simolus3/drift/blob/3253cd7ead51d7b163542929eb0fc8daee573fce/drift/lib/src/runtime/api/connection.dart#L17

If you're getting this error locally, please run `dart pub global activate melos && melos bootstrap` to get the dependencies in order.

PR #2116, opened 2022-10-29
Use native postgres boolean type instead of integers

### LINKED ISSUE #2113: Error when using boolean columns in postgresql when used as a boolean expression
**Describe the bug**
Having a table such as
```dart
class MyTables extends Table {
  BoolColumn get myBool => boolean()();
}
```
and connecting to a postgresql database, running the following query fails:
```dart
(select(myTables)..where((_) => myTables.myBool)).get();
```
```
Message: [PostgreSQLSeverity.error 42804: argument of WHERE must be type boolean, not type integer ]. Stacktrace: 
#0      _PostgreSQLExecutionContextMixin._query
package:postgres/src/connection.dart:513
#1      _PostgreSQLExecutionContextMixin.query
package:postgres/src/connection.dart:474
#2      _PgDelegate.runSelect
package:drift_postgres/src/pg_database.dart:108
#3      _BaseExecutor.runSelect
package:drift/…/helpers/engines.dart:68
#4      main
```

This is due to postgres, unlike sqlite, not implicitly converting integers to booleans. Instead, it has a dedicated boolean type. So, we could simply change the `sqlTypeName` of `DriftSqlType.bool` to be `boolean` instead of `integer` in postgres, but then we run into issues with the default constraints on bool columns (`CHECK ("col_name" IN (0, 1))`) producing errors since postgres can't compare a boolean to `0` or `1`.

I'd be willing to write up a PR for this if you could give me some guidance on how to approach the issue - being a new user of drift, I'm not aware of how this stuff is usually handled.

abitofevrything:
Makes it so that databases connected to postgresql will use the `boolean` type for boolean columns, instead of integers mapped to booleans as they are in sqlite.

These changes were made to be non-breaking but could use some cleanup in the next major release - mainly changing the `SqlTypes` methods to get the current dialect from somewhere else than an optional positional parameter.

Closes #2113


abitofevrything:
Converted to draft as this PR has issues with backwards compatibility - databases generated before this PR would not be compatible with the contents of this PR.

I'll work on allowing both integer and boolean columns to represent booleans in postgres, and users can migrate simply by changing the column type.

simolus3:
> databases generated before this PR would not be compatible with the contents of this PR.

You mean the generated database code, right? I think this is fine as long we keep the dependency constraints between `drift` and `drift_dev` consistent. The current contract is that we can make "breaking" changes in minor drift versions as long as the generator with the same minor version will generate correct code. It means that users will have to re-run `build_runner build` after upgrading their dependencies, but at least `drift` and `drift_dev` don't get out-of-sync.

abitofevrything:
No, I mean the actual database - the column type from before this PR would still be `integer` which is incompatible.

Since postgres accepts `1` and `0` as Boolean literals, it would be possible to be backwards compatible if it weren't for the postgres library only accepting `bool` for Boolean arguments and `int` for integer arguments in queries - other than that, reading from the db and constants embedded in the query as literals are compatible with both column types, so arguments are the only real blocker.

abitofevrything:
It would be easy for users to migrate though, simply `ALTER TABLE tbl_name ALTER COLUMN old_integer_boolean_column DROP CHECK SET TYPE boolean USING integer::boolean;`, or something along those lines (wrote that from memory).

simolus3:
Oh right, of course. Given that our support for postgres is still somewhat experimental, I think that shouldn't be a problem either. I think it's better to just do this right now instead of adding (and testing and maintaining) a build option just to get the broken behavior.

AlexKenbo:
> It would be easy for users to migrate though, simply `ALTER TABLE tbl_name ALTER COLUMN old_integer_boolean_column DROP CHECK SET TYPE boolean USING integer::boolean;`, or something along those lines (wrote that from memory).

Can you tell me if we are just starting out, can we use your commite right away?
We have been writing backends in Dart for 1.5 years, but we don't work with SQL directly. We work via Hasura :)

Now the application has grown and we need more control on the backend and we don't want to leave Dart for Java.

We already had a nice experience with Moor. We want to try it with Postgres as well. 
We're in business and we want to move in this month, hopefully all will be good.

And ideally we will join in on the package. 

I'd like your opinion @simolus3 - why is this still experimental and what might we encounter?
We already have experience writing complex queries and embedded SQL procedures.

abitofevrything:
Yes, this PR is in a working state. You just need to migrate existing databases, both the actual database contents and the generated database file to use it though.

I'll reopen it later, as if backwards compatibility isn't a huge concern then there is no reason for this to be kept as a draft.

abitofevrything:
After a bit of experimenting, this is the SQL migration query for old integer columns to the new boolean columns:
```sql
ALTER TABLE tbl_name DROP CONSTRAINT col_constraint, ALTER COLUMN col_name TYPE boolean USING col_name::boolean;
```
Where:
- `tbl_name` is the name of the table containing the column
- `col_constraint` is the name of the old `CHECK value IN (0, 1)` constraint (can be found with `\d+` from psql, generally looks like `$tblName_$colName_check`)
- `col_name` is the name of the column to change.

Run that query in a migration for every boolean column in your database, and that's all the migration done.

drift/lib/src/runtime/types/mapping.dart
```diff
@@ -80,12 +83,19 @@ class SqlTypes {
 
   /// Maps the [dart] value into a SQL literal that can be embedded in SQL
   /// queries.
-  String mapToSqlLiteral(Object? dart) {
+  String mapToSqlLiteral(
+    Object? dart, [
+    SqlDialect dialect = SqlDialect.sqlite,
```
simolus3: I wonder if this can be refactored so that the `dialect` is a field in this class instead of a parameter? It seems easy to forget passing this parameter around somewhere, I rather have it be in some central place. `SqlTypes` is marked `@internal`, so we can change its API without calling it a breaking change.

Maybe we can remove the `types` field from `DriftDatabaseOptions` and instead turn it into a `SqlTypes createMapping(SqlDialect dialect)` method. Then the `GenerationContext` could store the options associated with the current dialect by calling that method? That way the dialect only needs to be passed in one place.

drift/test/generated/todos.g.dart
```diff
@@ -760,25 +816,66 @@ class $UsersTable extends Users with TableInfo<$UsersTable, User> {
...
       'is_awesome', aliasedName, false,
       type: DriftSqlType.bool,
-      requiredDuringInsert: false,
-      defaultConstraints: 'CHECK ("is_awesome" IN (0, 1))',
-      defaultValue: const Constant(true));
+      requiredDuringInsert: false, defaultConstraints: (context) {
+    const dialectConstraints = {
+      SqlDialect.sqlite: 'CHECK ("is_awesome" IN (0, 1))',
+      SqlDialect.mysql: 'CHECK ("is_awesome" IN (0, 1))',
+      SqlDialect.postgres: 'CHECK ("is_awesome" IN (0, 1))',
```
simolus3: I assumed we don't need this check anymore in postgres? Am I overlooking something?

drift_dev/lib/src/writer/tables/table_writer.dart
```diff
@@ -50,9 +50,26 @@ abstract class TableOrViewWriter {
...
+    } else if (constraints.values.any((constraint) => constraint.isNotEmpty)) {
       // Use the default constraints supported by drift
-      additionalParams['defaultConstraints'] = asDartLiteral(constraints);
+
+      final literalEntries = [
+        for (final entry in constraints.entries)
+          'SqlDialect.${entry.key.name}: ${asDartLiteral(entry.value)},',
+      ];
+
+      additionalParams['defaultConstraints'] = '''(context) {
```
simolus3: This generates a lot of code. I think we should look for ways to optimize common cases here:

1. If the constraints are the same for all dialects: Maybe there could be a method like
```dart
// Maybe a static method in GeneratedColumn
void Function(GenerationContext) always(String sql) {
  return (context) => context.write(sql);
}
// we can now do something like `defaultConstraints: GeneratedColumn.always('PRIMARY KEY AUTOINCREMENT')`, saving a lot of code
```
2. There could be a similar helper method taking a `Map<SqlDialect, String>`. Essentially we'd be turning the current code snippet into a single method which is then called in the generated code, avoiding a lot of duplication.

simolus3 (commented):
I like the idea and the approach! My only concern is the increased code size, but I think we can fix that.

drift/lib/src/runtime/types/mapping.dart
abitofevrything: Making this change would be breaking - as I highlighted in the PR description.

The reason for this is that making it a field on the class would require knowing the dialect when the `SqlTypes` instance is constructed, which happens to be in `DriftDatabaseOptions`. `DriftDatabaseOptions` has no knowledge of the dialect of the database it is being attached to, so we would have to also add a parameter to `DriftDatabaseOptions` - another breaking change.

Of course, such a parameter could be made optional, but then that would require postgres users to change their database options to add a parameter, which seems redundant since only the executor should care about the dialect...

Therefore, since the `SqlTypes` methods are mostly used internally and not in many places (only when translating sql -> dart or dart -> sql), I think it's best to just have it as a parameter for now.

drift/test/generated/todos.g.dart
abitofevrything: You're right, I forgot to re-run `melos run build` after adding the code to handle this.

simolus3 (approved):
Thank you for your contribution!

simolus3:
@abitofevrything To reduce the size of the generated code, dialects for which we generate code now need to be enabled (in drift 2.11.0). If you only need postgres support, you can use `dialect: postgres` in your `build.yaml`. If you need database code that works on sqlite3 and on postgres as well, you can use [this setup](https://drift.simonbinder.eu/docs/advanced-features/builder_options/#multi-dialect-code-generation). Without that option, the postgres-specific types and constraints would be removed from generated code in the latest drift version.

PR #2121, opened 2022-11-03
Add isInValues and isNotInValues

North101:
Adds `isInValues` and `isNotInValues` extensions.

If the value is `null` it currently gets ignored. I'm not sure of the correct way to handle `null` in that case.

### DIFF
--- a/drift/CHANGELOG.md
+++ b/drift/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 - Add the `JsonTypeConverter2` mixin. It behaves similar to the existing json
   type converters, but can use a different SQL and JSON type.
+- Add `isInValues` and `isNotInValues` methods to columns with type converters.
+  They can be used to compare the column against a list of Dart expressions that
+  will be mapped through a type converter.
 
 ## 2.2.0
 
--- a/drift/lib/src/runtime/query_builder/schema/column_impl.dart
+++ b/drift/lib/src/runtime/query_builder/schema/column_impl.dart
@@ -287,10 +287,7 @@ class GeneratedColumnWithTypeConverter<D, S extends Object>
           check: check,
         );
 
-  /// Compares this column against the mapped [dartValue].
-  ///
-  /// The value will be mapped using the [converter] applied to this column.
-  Expression<bool> equalsValue(D? dartValue) {
+  S? _mapDartValue(D? dartValue) {
     S? mappedValue;
 
     if ($nullable) {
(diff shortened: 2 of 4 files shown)


simolus3 (commented):
Thanks for your work on this, I think this is a good addition! I'm fine with the non-nullable parameter. We could handle `null` values like this though:

```dart
final nonNullValues = <S>[];
var hasNull = false;
for (final value in values) {
  final mapped = _mapDartValue(value);
  if (mapped == null) {
    hasNull = true;
  } else {
    nonNullValues.add(mapped);
  }
}

if (hasNull) {
  isNull() || isIn(nonNullValues);
} else {
  return isIn(nonNullValues);
}
```

So essentially we'd only put non-null values in the `IN` clause and then handle nulls separately with an `or`. But I'm also not sure if that's an intuitive API or if we should just keep it non-nullable :thinking: 

Either way, some tests for this would be great. There could be a `column_test.dart` file [here](https://github.com/simolus3/drift/tree/develop/drift/test/database/expressions) with tests similar to this one: (maybe just use `expect(tbl.syncState.isInValues(...), generates('sync_state IS IN (...)'))`). https://github.com/simolus3/drift/blob/28c90b6ffbb11deeeaeab9523898624ef44c902e/drift/test/integration_tests/drift_files_test.dart#L239-L246

North101:
I wanted to confirm how to handle `null` before writing tests. I'll give that a go tomorrow

North101:
We could maybe add an extra param: `ignoreNull=false`

North101:
I've added checking against NULL and tests. Can you check my logic? I think I have it correct.

```sql
-- column1.isInValues('a', 'b', null)

SELECT column1 IN ('a', 'b') OR column1 IS NULL
FROM (
  VALUES
    ('a'), -- TRUE
    (NULL), -- TRUE
    ('c') -- FALSE
);
```


```sql
-- column1.isInValues('a', 'b')

SELECT column1 IN ('a', 'b') AND column1 IS NOT NULL
FROM (
  VALUES
    ('a'), -- TRUE
    (NULL), -- FALSE
    ('c') -- FALSE
);
```


```sql
-- column1.isNotInValues('a', 'b', null)

SELECT column1 NOT IN ('a', 'b') AND column1 IS NOT NULL
FROM (
  VALUES
    ('a'), -- FALSE
    (NULL), -- FALSE
    ('c') -- TRUE
);
```


```sql
-- column1.isNotInValues('a', 'b')

SELECT column1 NOT IN ('a', 'b') OR column1 IS NULL
FROM (
  VALUES
    ('a'), -- FALSE
    (NULL), -- TRUE
    ('c') -- TRUE
);
```

simolus3:
> I've added checking against NULL and tests. Can you check my logic? I think I have it correct.

Good call adding the `IS NOT NULL` condition as well! This looks good to me.

simolus3 (approved):

PR #2146, opened 2022-11-12
Fix runtime migration verification typo

JosiahMendes:
The original snippet has `kDebugBuild` as the condition without showing the earlier variable assignment, I'm assuming that the docs mean to use the Flutter defined `kDebugMode`.

### DIFF
--- a/docs/lib/snippets/migrations/runtime_verification.dart
+++ b/docs/lib/snippets/migrations/runtime_verification.dart
@@ -5,7 +5,7 @@ import 'package:drift/drift.dart';
 import 'package:drift_dev/api/migrations.dart';
 // #enddocregion
 
-const kDebugBuild = true;
+const kDebugMode = true;
 
 abstract class _$MyDatabase extends GeneratedDatabase {
   _$MyDatabase(QueryExecutor executor) : super(executor);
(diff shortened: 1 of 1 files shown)


docs/lib/snippets/migrations/runtime_verification.dart
```diff
@@ -3,10 +3,10 @@ import 'package:drift/drift.dart';
 // #docregion
 // import the migrations tooling
 import 'package:drift_dev/api/migrations.dart';
+// import foundation to determine if built with debug mode
+import 'package:flutter/foundation.dart';
```
simolus3: Thanks for the pull request ! But actually, I think this is not going to work because `flutter` is not available in the documentation (and I don't want it to be available, not depending on Flutter makes building it way easier).

As you said, `kDebugMode` is supposed to be the one from Flutter. A consequence of not depending on Flutter is that we can't add the import to the full file. If you want to fix this properly, it would be best to add a fake Flutter package in `docs/assets`. We also do this for [path_provider](https://github.com/simolus3/drift/tree/develop/docs/assets/path_provider) for the exact same reason. Maybe `kDebugMode` is known well enough to not need the explicit import shown in the snippet?

JosiahMendes:
Thanks for the review and the comments, I did try implementing a stub mock Flutter package, but it seems like having a dependency on flutter within the docs' `pubspec.yaml` was causing other build issues such as the `melos bootstrap` command to fail, so I reverted those changes and just corrected the variable name to `kDebugMode`.
I would assume that most users would be familiar with it, and for those who aren't a quick Google search or an IDE tip should be sufficient. Apologies for the multiple CI runs as I was unable to reproduce the issue locally!

simolus3:
No worries, and thanks for your contribution!

simolus3 (approved):

PR #2154, opened 2022-11-14
Fix binary encoding of BLOB/BYTEA fields when using postgres.

ignatz:


simolus3:
Hm, does `package:postgres` not support blobs? I'm fine with the workaround in drift, but perhaps we should also take a look at fixing that in the postgres client? Or is this a limitation of the protocol?

ignatz:
Hey Simon,

Apologies for not providing more context. Let me fix that. I created https://github.com/ignatz/drift_postgres_blob_issues to capture the symptoms I'm seeing (including error messages and stack traces). In a nutshell:

1. On batch inserts the postgres server is returning invalid syntax errors (I'm using pg14)
2. On individual inserts, not error but the stored data is jumbled. Arguably worse.
3. (tables are not automatically created but presumably that's unrelated. (1) and (2) are the result of me manually creating the table first)

I'm not familiar with the distribution of responsibilities between drift and postgres. I'd rely on your judgement to decide where and how to fix it. My layman's interpretation of (1) was that the query is malformed, specifically postgres' sql dialect expects binary data to be encoded differently.

To your question:

> Hm, does package:postgres not support blobs? 

Both postgres and `package:postgres` support blobs, they call them byte arrays (https://pub.dev/documentation/postgres/latest/postgres/PostgreSQLDataType.html)

simolus3:
Thanks for providing the details and the repository! I can reproduce the issue.

As it seems, the syntax error comes from `package:postgres` inlining the parameters into the query. So we send `INSERT INTO "foos" ("bar") VALUES (@1)` and the client package turns that into `VALUES ({1, 2, 3, 4})`.
I think the best solution would be to use prepared statements, which seems to happen when we use `.query` instead of `.execute()`. 

And then there's the issue about the parameters having the wrong format. Peeking into the source code of the `postgres` package, it seems like we need to specify the actual type in the SQL. Otherwise, the stuff gets inserted as a string. You can see that in the returned row, `[123, 49, 44, 50, 44, 51, 44, 52, 125]` is the utf8 encoding of `{1,2,3,4}`.

simolus3:
However, it strongly feels like we're fighting `package:postgres` here. Internally it uses indexed parameters (e.g. `$1`), just like sqlite3 does. But we have to jump through the `@1` format only for `package:postgres` to turn that into the right format internally. 
Maybe drift should be using a lower level of abstraction that `package:postgres` could provide for us (https://github.com/isoos/postgresql-dart/issues/72).

simolus3:
I have created a PR to the postgres client which will allow us to send the byte array over the wire directly without any interference from that library: https://github.com/isoos/postgresql-dart/pull/73

This will require some changes to drift (made in the `better-postgres-types`) branch. Once the postgres PR is through, I'll finish that work. I'll also test that with your example to make sure that it indeed fixes the encoding problem. Overall, these changes will make postgres implementation a lot less hacky which is great.

ignatz:
That's amazing. Thanks for the prompt handling and kudos to you!

simolus3:
Sorry for the long wait on this one, it look a while to get the changes integrated into in `package:postgres`. With cc16156c84d3ffe8db3a23bcc9969ab00adea6ff on `develop`, we have a more reliable postgres implementation that sends the correct types for parameters.

ignatz:
Thanks for circling back, much appreciated.

PR #2161, opened 2022-11-21
Remove unnecessary where clause for batch.deleteAll

North101:
Sorry, meant to do change this before being merged


simolus3 (approved):
Ah yeah :D Both are fine, but I agree it's cleaner like this.

PR #2168, opened 2022-11-26
Add `textEnum` column builder and `EnumNameConverter` 

### LINKED ISSUE #2115: add EnumNameConverter
drift offers an inbuilt `EnumIndexConverter`, but I woud prefer them being stored as a String that is their name.

Add something like
```dart
class EnumNameConverter<T extends Enum> extends TypeConverter<T, String> {
  /// All values of the enum.
  final List<T> values;

  /// Constant default constructor.
  const EnumNameConverter(this.values);

  @override
  T fromSql(String fromDb) => values.asNameMap()[fromDb]!;

  @override
  String toSql(T value) => value.name;
}
```
This relies on the dart version which introduced the enum names, but afaik drift already uses dart 2.17 so that should be fine.

ValentinVignal:
Should close https://github.com/simolus3/drift/issues/2115

### DIFF
--- a/drift_dev/lib/src/analysis/resolver/drift/table.dart
+++ b/drift_dev/lib/src/analysis/resolver/drift/table.dart
@@ -19,7 +19,7 @@ import 'find_dart_class.dart';
 
 class DriftTableResolver extends LocalElementResolver<DiscoveredDriftTable> {
   static final RegExp _enumRegex =
-      RegExp(r'^enum\((\w+)\)$', caseSensitive: false);
+      RegExp(r'^enum(name)?\((\w+)\)$', caseSensitive: false);
 
   DriftTableResolver(super.file, super.discovered, super.resolver, super.state);
 
@@ -55,10 +55,10 @@ class DriftTableResolver extends LocalElementResolver<DiscoveredDriftTable> {
 
       final typeName = column.definition?.typeName;
 
-      final enumMatch =
+      final enumIndexMatch =
           typeName != null ? _enumRegex.firstMatch(typeName) : null;
-      if (enumMatch != null) {
-        final dartTypeName = enumMatch.group(1)!;
+      if (enumIndexMatch != null) {
+        final dartTypeName = enumIndexMatch.group(2)!;
         final imports = file.discovery!.importDependencies.toList();
         final dartClass = await findDartClass(imports, dartTypeName);
 
--- a/drift_dev/lib/src/analysis/resolver/drift/table.dart
+++ b/drift_dev/lib/src/analysis/resolver/drift/table.dart
@@ -54,6 +54,7 @@ class DriftTableResolver extends LocalElementResolver<DiscoveredDriftTable> {
       AnnotatedDartCode? defaultArgument;
 
       final typeName = column.definition?.typeName;
+
       final enumMatch =
           typeName != null ? _enumRegex.firstMatch(typeName) : null;
       if (enumMatch != null) {
@@ -72,6 +73,9 @@ class DriftTableResolver extends LocalElementResolver<DiscoveredDriftTable> {
             (msg) => reportError(
                 DriftAnalysisError.inDriftFile(column.definition ?? stmt, msg)),
             dartClass.classElement.thisType,
+            column.type.type == BasicType.int
+                ? EnumType.intEnum
+                : EnumType.textEnum,
           );
         }
       }
--- a/drift_dev/lib/src/analysis/resolver/drift/table.dart
+++ b/drift_dev/lib/src/analysis/resolver/drift/table.dart
@@ -73,9 +73,7 @@ class DriftTableResolver extends LocalElementResolver<DiscoveredDriftTable> {
             (msg) => reportError(
                 DriftAnalysisError.inDriftFile(column.definition ?? stmt, msg)),
             dartClass.classElement.thisType,
-            column.type.type == BasicType.int
-                ? EnumType.intEnum
-                : EnumType.textEnum,
+            type == DriftSqlType.int ? EnumType.intEnum : EnumType.textEnum,
           );
         }
       }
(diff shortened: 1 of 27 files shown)


drift_dev/lib/src/analysis/resolver/drift/table.dart
```diff
@@ -72,6 +73,9 @@ class DriftTableResolver extends LocalElementResolver<DiscoveredDriftTable> {
             (msg) => reportError(
                 DriftAnalysisError.inDriftFile(column.definition ?? stmt, msg)),
             dartClass.classElement.thisType,
+            column.type.type == BasicType.int
+                ? EnumType.intEnum
+                : EnumType.textEnum,
```
ValentinVignal: Is it the correct way to do this conditional check?

drift_dev/test/analysis/resolver/dart/enum_columns_test.dart
```diff
@@ -55,8 +64,14 @@ void main() {
...
-          allOf(contains('Not an enum'), contains('NotAnEnum')))),
+      containsAllInOrder(
+        [
+          isA<DriftAnalysisError>().having((e) => e.message, 'message',
+              allOf(contains('Not an enum'), contains('NotAnEnum'))),
+          isA<DriftAnalysisError>().having((e) => e.message, 'message',
+              allOf(contains('Not an enum'), contains('NotAnEnum'))),
+        ],
+      ),
     );
```
ValentinVignal: Do you know a better way to verify an iterable contains 2 occurrences of a matcher?

drift_dev/lib/src/analysis/resolver/drift/table.dart
simolus3: Let's use the local `type` variable instead. You can compare it against `DriftSqlType.int`.

drift_dev/test/analysis/resolver/dart/enum_columns_test.dart
simolus3: I would just extract the matcher into a local variable and use it twice:

```dart
final notAnEnumError = isDriftError(allOf(contains('Not an enum'), contains('NotAnEnum')));

expect(file.allErrors, [notAnEnumError, notAnEnumError);
```

simolus3 (commented):
The implementation looks good to me, nice work! I have some minor comments on how to compare the type and for the matcher.

We should also mention this in the changelog and the documentation, but I can take care of that as well.

drift_dev/lib/src/analysis/resolver/drift/table.dart
ValentinVignal: I've used `type` in [🎨 Better type comparison and matcher](https://github.com/simolus3/drift/pull/2168/commits/4030db4c5f6eb7795f3f21c2a65065f025bdbdc0)

drift_dev/test/analysis/resolver/dart/enum_columns_test.dart
ValentinVignal: I extracted the matcher in [🎨 Better type comparison and matcher](https://github.com/simolus3/drift/pull/2168/commits/4030db4c5f6eb7795f3f21c2a65065f025bdbdc0)

ValentinVignal:
> We should also mention this in the changelog and the documentation, but I can take care of that as well.

I can try to finish the work 😃 

I upgraded the change logs in  [⬆️ Upgrade version number](https://github.com/simolus3/drift/pull/2168/commits/d7ef3746776919fb33b17c3b3f790743b293f73b) but also in [Merge remote-tracking branch 'upstream/develop' into textEnum](https://github.com/simolus3/drift/pull/2168/commits/a993d433c598218daa1ec349cc7b5c50d06f5a01) (because of some conflicts). I'm not too sure I've done it correctly.

I also tried to update the documentation in [📝 Update the documentation](https://github.com/simolus3/drift/pull/2168/commits/3f0e40b8a7a74cecb80e096076a6620a5ff2b87d). Tell me what you think about it.

Feel free to push on my branch if you think it is easier/faster.

drift_dev/pubspec.yaml
```diff
@@ -1,6 +1,6 @@
 name: drift_dev
 description: Dev-dependency for users of drift. Contains a the generator and development tools.
-version: 2.3.0
+version: 2.4.0-dev
```
simolus3, line 3: Please also change the `drift: ` dependency in this package to `>=2.4.0 <2.5.0`.

simolus3 (commented):
Thanks for also updating the documentation!

In the generator, it would also be good to have a test for drift files - you can take a look at the `parses enum columns` test in `drift_dev/test/analysis/resolver/drift/table_test.dart` file for inspiration.
There should also be tests interacting with `EnumNameConverter` at runtime - similar to `drift/test/database/types/enum_index_converter.dart`.

ValentinVignal:
> In the generator, it would also be good to have a test for drift files - you can take a look at the `parses enum columns` test in `drift_dev/test/analysis/resolver/drift/table_test.dart` file for inspiration.

I've introduced `ENUMNAME` in [:sparkles Add ENUMNAME type](https://github.com/simolus3/drift/pull/2168/commits/3cfa69c9d965cc2667603580590ca020a6424c90). But I don't use this feature in my apps so I don't really know what I'm doing here... Please be careful when you review this part

> There should also be tests interacting with `EnumNameConverter` at runtime - similar to `drift/test/database/types/enum_index_converter.dart`.

I've added those tests in [✅ Add enum name converter test](https://github.com/simolus3/drift/pull/2168/commits/ed82e46f2dfffa9998031df038f2d8de384d84a7)

simolus3 (approved):
Thanks a lot for your contribution, I think this is a helpful feature for sure!

PR #2181, opened 2022-12-01
Add the `column_name_case` option

### LINKED ISSUE #1068: Support overriding naming for columns globally (e.g snake_case, camelCase, exact property name, etc...)
I need to make my column names to be the same as my property names, but using named in each column for each table can be tedious and prone to error by not adding it

add an option to let moor know I want my column names to be created in a specific manner globally.

ValentinVignal:
Should close https://github.com/simolus3/drift/issues/1068

Add the `column_name_case` option to the builder
The possible options are 
- `preserve`
- `camel_case`
- `CONSTANT_CASE`
- `snake_case`
- `PascalCase`
- `lowercase`
- `UPPERCASE`

### DIFF
--- a/drift_dev/pubspec.yaml
+++ b/drift_dev/pubspec.yaml
@@ -1,6 +1,6 @@
 name: drift_dev
 description: Dev-dependency for users of drift. Contains a the generator and development tools.
-version: 2.4.0-dev
+version: 2.5.0-dev
 repository: https://github.com/simolus3/drift
 homepage: https://drift.simonbinder.eu/
 issue_tracker: https://github.com/simolus3/drift/issues
--- a/drift_dev/pubspec.yaml
+++ b/drift_dev/pubspec.yaml
@@ -1,6 +1,6 @@
 name: drift_dev
 description: Dev-dependency for users of drift. Contains a the generator and development tools.
-version: 2.5.0-dev
+version: 2.4.0-dev
 repository: https://github.com/simolus3/drift
 homepage: https://drift.simonbinder.eu/
 issue_tracker: https://github.com/simolus3/drift/issues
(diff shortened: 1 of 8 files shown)


drift_dev/pubspec.yaml
```diff
@@ -1,6 +1,6 @@
 name: drift_dev
 description: Dev-dependency for users of drift. Contains a the generator and development tools.
-version: 2.4.0-dev
+version: 2.5.0-dev
```
ValentinVignal: I'm not too sure about this one, what do you want the version number to be?

simolus3: Let's leave it as `2.4.0-dev` - I just use the `-dev` suffix as a reminder that this state hasn't been released yet. In the end, all new features will be released as part of `2.4.0`.

simolus3 (commented):
Thanks for the PR!

I wonder if this should also apply to table names? E.g. if I have `class UserGroupMemberships extends Table` class in Dart, should the option affect the table's SQL name too?

ValentinVignal:
> I wonder if this should also apply to table names? E.g. if I have class UserGroupMemberships extends Table class in Dart, should the option affect the table's SQL name too?

Sure, would it be better to create another option for that? Or do you think I should reuse the same one (and then, rename it) ?

simolus3:
> Sure, would it be better to create another option for that? Or do you think I should reuse the same one (and then, rename it) ?

I can't think of a case where the table and columm name casing should be different. If it turns out to be something that users want, we can still introduce a separate option for it later.

ValentinVignal:
@simolus3 I renamed to option into `case_from_dart_to_sql` in [🚚 Rename column_name_case into case_from_dart_to_sql](https://github.com/simolus3/drift/pull/2181/commits/496635afb53bbafc3857f8bb6c41db42a479aee3) and applied it to tables in [♻️ apply the recase option to table name](https://github.com/simolus3/drift/pull/2181/commits/8a34a41fba2da0c8b96a5d9a3243cba98393bd07)

However, the test file I create is `drift_dev/test/analysis/resolver/dart/column_test.dart`, but it is now also testing the names of the tables. Do you want me to rename it and/or move it somewhere else?

ValentinVignal:
For the `CI / sqlparser`, I see those logs 🤔 

```
Run dart analyze --fatal-infos --fatal-warnings
Analyzing sqlparser...
The analysis server timed out while shutting down.
No issues found!
The analysis server shut down unexpectedly.
Please report this at dartbug.com.
Error: Process completed with exit code 1.
```

And I don't think I'm allowed to re-run the job

simolus3 (approved):
Thanks for also applying this to table names and for the helpful tests! LGTM :rocket:

simolus3:
> And I don't think I'm allowed to re-run the job

I just re-ran it, it's probably a rare bug in the analysis server and not our fault.

PR #2184, opened 2022-12-05
Various fixes for modular code gen

North101:
This PR contains various fixes for modular code gen.

How to reproduce the issues I've fixed:

- Enable modular code gen
- Add views to .drift files (which listens to a table from another file)
- Add .drift file with table but NO queries (select/insert/update/delete) and then add it to a join in another file
- add query that DOES insert/update/delete

### DIFF
--- a/drift_dev/lib/src/analysis/driver/state.dart
+++ b/drift_dev/lib/src/analysis/driver/state.dart
@@ -3,7 +3,6 @@ import 'package:meta/meta.dart';
 import 'package:path/path.dart' show url;
 import 'package:sqlparser/sqlparser.dart' hide AnalysisError;
 
-import '../resolver/intermediate_state.dart';
 import '../results/database.dart';
 import '../results/element.dart';
 import '../results/file_results.dart';
--- a/drift_dev/lib/src/analysis/driver/state.dart
+++ b/drift_dev/lib/src/analysis/driver/state.dart
@@ -3,10 +3,12 @@ import 'package:meta/meta.dart';
 import 'package:path/path.dart' show url;
 import 'package:sqlparser/sqlparser.dart' hide AnalysisError;
 
+import '../resolver/intermediate_state.dart';
 import '../results/database.dart';
 import '../results/element.dart';
 import '../results/file_results.dart';
 import '../results/query.dart';
+import 'driver.dart';
 import 'error.dart';
 
 class FileState {
@@ -18,6 +20,8 @@ class FileState {
   final Map<DriftElementId, ElementAnalysisState> analysis = {};
   FileAnalysisResult? fileAnalysis;
 
+  bool? _needsModularAccessor;
+
   FileState(this.ownUri);
 
   String get extension => url.extension(ownUri.path);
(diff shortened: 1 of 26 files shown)


drift_dev/lib/src/analysis/driver/state.dart
```diff
@@ -33,7 +33,9 @@ class FileState {
     final hasImports = discovery?.importDependencies.isNotEmpty == true;
     final hasQuery = analyzedElements.any((e) => e is DefinedSqlQuery);
 
-    return hasImports || hasQuery;
+    return hasImports ||
+        hasQuery ||
+        true; // TODO: This is required for other cases than just the above
```
North101: One of my tables was needed for the typedef function of a dynamic `$where` expression. This function was returning `false` which resulted in the `ModularAccessor` class not being created.

simolus3: If you have a scoped `$where`, that can only appear in a query so `hasQuery` should be true in that case right? Maybe this is another bug?

I'm happy to take a look at the original bug as well, but I think I don't fully understand how this could happen yet.

North101:
I've removed my attempts to fix the unneeded modular accessor getters.

A workaround is just to create a query in the .drift file

North101:
How to reproduce the issues I've fixed:

- Enable modular code gen
- Add views to .drift files (which listens to a table from another file)
- Add .drift file with table but NO queries (select/insert/update/delete)
- add query that DOES insert/update/delete

simolus3:
Thanks for the additional fixes. I've added a view to the `examples/modular` project and I can see that the problems there are fixed.  I've also separated the `users.drift` file into one table file and a query file in an attempt to reproduce the other issues, but I'm not entirely sure if I got that right.

> A workaround is just to create a query in the .drift file

If it's easy to reproduce, can you bring the `examples/modular` example in this repository into a state where an accessor should be generated but isn't? I'd love to take a look at what's happening in the analyzer for that case, this sounds like something that really should be fixed.

North101:
> If it's easy to reproduce, can you bring the examples/modular example in this repository into a state where an accessor should be generated but isn't? I'd love to take a look at what's happening in the analyzer for that case, this sounds like something that really should be fixed.

Sure can do. The problem is actually the opposite. An
`<Table>Drift get <table>Drift => this.accessor('table')` is created when `<Table>Drift` doesn't exist (because the `.drift` file didn't have any queries) because a different `.drift` file referenced the table.

simolus3 (approved):
I think I managed to fix the unnecessary accessors being generated. Thanks for all your help!

North101:
> I think I managed to fix the unnecessary accessors being generated. Thanks for all your help!

Its fixed all of my previous issues, however it has introduced a new issue.

I have a `.drift` file which has:

- imports other `.drift` files (for foreign keys)
- creates table

Then i have another `.drift` file which:
- imports the previous `.drift` file
- uses it in a join

Then second is trying to create an accessor to the first, but first doesn't have a `ModularAccessor` class.

I can create a project later to reproduce the issue if I have time.

North101:
Here is a sample project:
https://github.com/North101/moor/tree/modular_gen_bug/examples/modular2

I think its the circular import. If I move the query to `table1.drift`, the error moves to `table1.drift.dart`. If I then remove the import in `table2.drift` then the error is gone.

simolus3:
Thanks! Should be fixed in fb3d832f0f4b145dc197e35cb75629f4fb6f38be.

North101:
Thats all of my issues solved. Thanks!

PR #2189, opened 2022-12-07
A couple bugfixes for complex migration

JuanMAlcarazXYZ:
I was trying to change a TextColumn to DateTimeColumn and initialize all values with a date 30 days in the past, like this:

```
await m.alterTable(TableMigration(myTable, columnTransformer: {
  myTable.date: Variable<DateTime>(
    DateTime.now().add(const Duration(days: -30)),
  ),
}));
```

This table has a generated column, and it couldn't complete the insert with this exception:

_SqliteException(1): cannot INSERT into generated column "generated", SQL logic error._

Debugging I realized it was actually trying to insert a value to the generated column on the temporary table, so I changed it to skip columns with a not null `generatedAs` property, like this:

```
for (final column in table.$columns) {
  if (column.generatedAs != null) continue;
```

After that, it failed again with this exception:

_ArgumentError (Invalid argument (params[1]): Allowed parameters must either be null or bool, int, num, String or List\<int\>.: Instance of 'Variable\<DateTime\>')_

This was happening because it passed `introducedVariables` to the INSERT statement instead of `boundVariables` like it does in line 119 when creating a table, so it made sense to pass the same argument to the same method and changed it.

Old code:
```
context.buffer.write(' FROM ${context.identifier(tableName)};');
await _issueCustomQuery(context.sql, context.introducedVariables);
```
Line 119:
```
return _issueCustomQuery(context.sql, context.boundVariables);
```

Hope it helps, greetings!

### DIFF
--- a/drift/lib/src/runtime/query_builder/migration.dart
+++ b/drift/lib/src/runtime/query_builder/migration.dart
@@ -203,9 +203,7 @@ class Migrator {
       context.buffer.write('INSERT INTO $temporaryName (');
       var first = true;
       for (final column in table.$columns) {
-        if (column.generatedAs != null) {
-          continue;
-        }
+        if (column.generatedAs != null) continue;
 
         final transformer = migration.columnTransformer[column];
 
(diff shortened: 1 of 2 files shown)


simolus3 (approved):
Hi, thanks a lot for your contribution! I agree that the changes are necessary, I've also added two tests for the issues (generated column and using a variable).

PR #2198, opened 2022-12-12
Add functions: iif, ifnull, nullif

North101:


drift/lib/src/runtime/query_builder/expressions/misc.dart
```diff
@@ -0,0 +1,10 @@
+import '../query_builder.dart';
+
+/// The iif(X,Y,Z) function returns the value Y if X is true, and Z
+/// otherwise. The iif(X,Y,Z) function is logically equivalent to and
+/// generates the same bytecode as the CASE expression "CASE WHEN X THEN Y
+/// ELSE Z END".
+Expression<T> iif<T extends Object>(
+    Expression<T> x, Expression<bool> y, Expression<T> z) {
```
simolus3: Shouldn't it be `Expression<bool> x`?

I'm also considering whether this should be an extension on `Expression<bool>`? Maybe it could be `Expression<T> match({required Expression<T> ifTrue, required Expression<T> ifFalse})` (maybe with positional parameters to reduce syntactic overhead, not sure yet). Most users probably aren't available of `iif` as a SQL function and an extension with a more Dart-y name could give this better visibility.

drift/lib/src/runtime/query_builder/expressions/null_check.dart
```diff
@@ -29,6 +29,22 @@ Expression<T> coalesce<T extends Object>(List<Expression<T>> expressions) {
...
+  return FunctionCallExpression<T>('IFNULL', [x, y]);
+}
+
+/// The nullif(X,Y) function returns its first argument if the arguments are
+/// different and NULL if the arguments are the same. The nullif(X,Y) function
+/// searches its arguments from left to right for an argument that defines a
+/// collating function and uses that collating function for all string
+/// comparisons. If neither argument to nullif() defines a collating function
+/// then the BINARY collating function is used.
+Expression<T> nullif<T extends Object>(Expression<T> x, Expression<T> y) {
```
simolus3: I think this could also be a method / an extension on `Expression<T>`, so that you'd be able to write `x.nullIf(y)`.

drift/lib/src/runtime/query_builder/expressions/null_check.dart
```diff
@@ -29,6 +29,22 @@ Expression<T> coalesce<T extends Object>(List<Expression<T>> expressions) {
   return FunctionCallExpression<T>('COALESCE', expressions);
 }
 
+/// The ifnull() function returns a copy of its first non-NULL argument, or
+/// NULL if both arguments are NULL
```
simolus3: I think the Dartdoc for `ifnull` and `coalesce` should reference each other to highlight that a more general case with more arguments exists as well.

simolus3 (commented):
Thanks for adding these functions.  I'm wondering if `iif` and `nullif` would be better represented as an extension for discoverability with auto-complete.

And of course it would be good to have tests for them, you can use the `generates()` matcher to check that a expression generates certain SQL.

North101:
Pushed suggested changes.

drift/lib/src/runtime/query_builder/expressions/null_check.dart
```diff
@@ -29,6 +29,21 @@ Expression<T> coalesce<T extends Object>(List<Expression<T>> expressions) {
...
+/// to null. See [coalesce] if you need more than 2.
+Expression<T> ifNull<T extends Object>(
+    Expression<T> first, Expression<T> second) {
+  return FunctionCallExpression<T>('IFNULL', [first, second]);
+}
+
+/// Extension defines the `nullIf` members.
+extension SqlNullIf<T extends Object> on Expression<T> {
+  /// Returns null if both [matcher] matches this expression.
+  Expression<T> nullIf(Expression<T> matcher) {
```
simolus3: Given that it's available on all expressions, let's actually make this a regular method on `Expression`.

North101: Done

simolus3 (approved):

PR #2199, opened 2022-12-13
Fixes for modular generation

kuhnroyal:
This fixes 2 problems with modular generation:
- fix duplicate result set classes being generated in a single accessor
- fix type converters missing import prefixes

### DIFF
--- a/drift_dev/lib/src/writer/writer.dart
+++ b/drift_dev/lib/src/writer/writer.dart
@@ -198,7 +198,9 @@ abstract class _NodeOrWriter {
     final buffer = StringBuffer();
 
     for (final lexeme in code.elements) {
-      if (lexeme is DartTopLevelSymbol) {
+      if (lexeme is AnnotatedDartCode) {
+        buffer.write(dartCode(lexeme));
+      } else if (lexeme is DartTopLevelSymbol) {
         final uri = lexeme.importUri;
 
         if (uri != null) {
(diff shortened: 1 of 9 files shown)


drift_dev/lib/src/writer/writer.dart
```diff
@@ -198,7 +198,9 @@ abstract class _NodeOrWriter {
     final buffer = StringBuffer();
 
     for (final lexeme in code.elements) {
-      if (lexeme is DartTopLevelSymbol) {
+      if (lexeme is AnnotatedDartCode) {
```
simolus3: An `AnnotatedDartCode` shouldn't have another `AnnotatedDartCode` as an element, that's probably a bug in another part of the generator.

I think we should add an assertion to the constructor of `AnnotatedDartCode` checking that each element is either a `String` or a `DartTopLevelSymbol`. Then we can figure out where the wrong element comes from and fix that.

simolus3 (commented):
Thanks for the fixes! I'm a bit worried about `AnnotatedDartCode` apparently nesting - if you have an example where that happens I can also take a look at fixing that.

drift_dev/lib/src/writer/writer.dart
kuhnroyal: That happens when the type converter is read: https://github.com/simolus3/drift/blob/develop/drift_dev/lib/src/writer/writer.dart#L115
And this results in the `AnnotatedDartCode.toString()` being used which is missing the import prefix.

kuhnroyal:
Yea that makes more sense, I run a quick test.

kuhnroyal:
Works on my project 👍

simolus3 (approved):
Thanks again for the fixes!

PR #2220, opened 2022-12-18
Validate enum default value in drift files

North101:
For enums in drift files you need to use the converted value. This PR checks if the index (for ENUM) or the field (for ENUMNAME) is a valid value.


North101:
The original version allowed you to reference the enum like you would a table for the default value but I felt it was a bit hacky so changed it to only do a validation check on the value.

simolus3 (approved):
Very neat idea :+1: I've extended the check to also apply to statements (e.g. `INSERT INTO foo VALUES ('not a valid enum')`).

PR #2222, opened 2022-12-18
Fix: remove illegal cast

### LINKED ISSUE #2221: [drift_postgres] type 'int' is not a subtype of type 'bool' in type cast
Bug report for the experimental feature `drift_postgres`

**Describe the bug**

The following select will fail on a model with BoolColumn:
```
// Check if user already exists
final dbUser = await (db.select(db.users)
          ..where((tbl) => tbl.email.equals(requestedMail)))
        .getSingleOrNull();
```

Error:
```
ERROR - 2022-12-18 23:42:10.397046
POST /auth/register
Error thrown by handler.
type 'int' is not a subtype of type 'bool' in type cast

SqlTypes.read                                                     package:drift/…/types/mapping.dart:134
$UsersTable.map                                                package:helltax_database/src/postgres_database.g.dart:321
MapAndAwait.mapAsyncAndAwait.<fn>.<fn>  package:drift/…/utils/async.dart:10
dart:async                                                                   new Future.sync
MapAndAwait.mapAsyncAndAwait.<fn>            package:drift/…/utils/async.dart:10
dart:async                                                                   Future.wait
MapAndAwait.mapAsyncAndAwait                     package:drift/…/utils/async.dart:10
SimpleSelectStatement._mapResponse               package:drift/…/select/select.dart:73
```

Model:
```
class Users extends Table {
  TextColumn get id => text().clientDefault(() => uuid.v4())();
  TextColumn get email => text().unique()();
  TextColumn get passwordHash => text()();
  BoolColumn get emailConfirmed => boolean().withDefault(Constant(false))();
  DateTimeColumn get createdAt =>
      dateTime().clientDefault(() => DateTime.now())();
  DateTimeColumn get updatedAt =>
      dateTime().clientDefault(() => DateTime.now())();

  @override
  Set<Column> get primaryKey => {id};
}
```

ynnob:
Fixes #2221 

Removed illegal cast int as bool:
`return sqlValue as T;`

Also removed _dialect check. I don't think it is needed here. Couldn't really tell what you were trying to archive here.

### DIFF
--- a/drift/lib/src/runtime/types/mapping.dart
+++ b/drift/lib/src/runtime/types/mapping.dart
@@ -134,11 +134,7 @@ class SqlTypes {
     // ignore: unnecessary_cast
     switch (type as DriftSqlType<Object>) {
       case DriftSqlType.bool:
-        if (_dialect == SqlDialect.sqlite) {
-          return (sqlValue != 0) as T;
-        } else {
-          return sqlValue as T;
-        }
+        return (sqlValue != 0) as T;
       case DriftSqlType.string:
         return sqlValue.toString() as T;
       case DriftSqlType.bigInt:
--- a/drift/lib/src/runtime/types/mapping.dart
+++ b/drift/lib/src/runtime/types/mapping.dart
@@ -134,7 +134,7 @@ class SqlTypes {
     // ignore: unnecessary_cast
     switch (type as DriftSqlType<Object>) {
       case DriftSqlType.bool:
-        return (sqlValue != 0) as T;
+        return (sqlValue != 0 && sqlValue != false) as T;
       case DriftSqlType.string:
         return sqlValue.toString() as T;
       case DriftSqlType.bigInt:


ynnob:
Is there a situation where we would need to check for _dialect?
Because in general a conversion int as bool should never work right?

drift/lib/src/runtime/types/mapping.dart
```diff
@@ -134,11 +134,7 @@ class SqlTypes {
     // ignore: unnecessary_cast
     switch (type as DriftSqlType<Object>) {
       case DriftSqlType.bool:
-        if (_dialect == SqlDialect.sqlite) {
-          return (sqlValue != 0) as T;
-        } else {
-          return sqlValue as T;
-        }
+        return (sqlValue != 0) as T;
```
simolus3: In postgres, we store Dart bools with the native boolean type. So when selecting a boolean column, `sqlValue` would be a boolean here. That's where the original cast comes from.

I think this is correct for some comparisons where we seem to be getting an integer back from the database, but we should still make sure to interpret `false` as `false`.

Maybe `(sqlValue != 0 && sqlValue != false) as T;`?

ynnob: Yes you are correct. For whatever reason i can not reporduce my own issue here.
So i am not really sure what happend here but i am 100% sure that yesterday the sqlvalue was 0 instead of false. Now today the sqlvalue is false and can be casted as bool. 

Anyway, I will change the return to your suggestion because that should cover all relevant cases. 👍

simolus3 (approved):

PR #2234, opened 2022-12-26
Support existing row classes for query types

### LINKED ISSUE #1134: Custom row classes
This tracks enhancements to the custom row classes feature in moor 4.3. Custom row classes let users specify an existing class that moor will use as a data class:

```dart
@UseRowClass(MyRowClass)
class MyTable extends Table {
}

class MyRowClass {
  final int id;

  MyRowClass(this.id);
}
```

Goals

- [x] Basic support for custom row classes
- [x] Proper type checking
- [x] Support custom row classes for tables declared in moor files
- [x] Support named constructors
- [x] Aid in generating `toColumns` if desired (not urgent, we can delegate that method to the generated companion)
- [x] Support something similar for generated queries?

simolus3:
Support the `WITH ExistingType` syntax for defined queries as well. Finally closes #1134.

Some remaining todos:

- [x] Make the `DartType` for existing table/view row classes available and serializable.
- [X] Support passing nested types to an existing query row type. The model needs to be able to express different inner row types for nested queries.
- [X] Check nested types if they're available (i.e. not generated by drift)
- [X] The default record type should also generate a nested record type for nested queries.
- [x] Support existing row types in the generator
- [x] Tests: One for each error message, test records and existing classes in the generator, integration test in `drift/`.
- [x] Documentation!


North101:
Let me know when you want to test this. I've got a bunch of queries I can try this with

simolus3:
@North101 That would be great! I'll write some documentation and then merge this to `develop`, but it's ready for the examples in drift so feel free to throw a couple of real-world usages at it :)

PR #2241, opened 2022-12-29
Add toColumns mixin when using user generated classes

North101:
At the moment when using user generated classes, you need to write your own code `toColumns`. Because user generated columns must match the type and name the row, I feel like this is something that can be generated by drift as a mixin.

This is more a proof of concept. Any suggestions on how you'd like it implemented?

### DIFF
--- a/docs/pages/docs/Advanced Features/builder_options.md
+++ b/docs/pages/docs/Advanced Features/builder_options.md
@@ -73,6 +73,8 @@ At the moment, drift supports these options:
   For more information on these modes, see [datetime options]({{ '../Getting started/advanced_dart_tables#datetime-options' | pageUrl }}).
 * `case_from_dart_to_sql` (defaults to `snake_case`): Controls how the table and column names are re-cased from the Dart identifiers.
   The possible values are  `preserve`, `camelCase`, `CONSTANT_CASE`, `snake_case`, `PascalCase`, `lowercase` and `UPPERCASE` (default: `snake_case`).
+* `write_to_columns_mixins`: Whether the `toColumns` method should be written as a mixin instead of being added directly to the data class.
+   This is useful when using [existing row classes]({{ 'custom_row_classes.md' | pageUrl }}), as the mixin is generated for those as well.
 
 ## Assumed SQL environment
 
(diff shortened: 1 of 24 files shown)


simolus3:
In the Dart API, we actually already have this with [`generateInsertable: true`](https://drift.simonbinder.eu/api/drift/userowclass/generateinsertable). So I think the right question is just how we expose it in a `.drift` file.

I'm open for ideas here, perhaps that should just be turned on by default for `.drift` files?

North101:
> In the Dart API, we actually already have this with [`generateInsertable: true`](https://drift.simonbinder.eu/api/drift/userowclass/generateinsertable). So I think the right question is just how we expose it in a `.drift` file.
> 
> I'm open for ideas here, perhaps that should just be turned on by default for `.drift` files?

I completely missed that Maybe it could be a build option instead? I can't see a reason not generate it tbh cause otherwise you'd be writing it by hand and presumably producing the exact same code

North101:
After looking into `generateInsertable` and testing it out by forcing it to be true, I kind of prefer generating mixins like in this PR. `toInsertable` creates a wrapper that creates a companion that then returns a map, while the mixin just adds the toColumns` method to the class.

North101:
using the `toInsertable()` function, it causes ~6000 unnecessary class creations (~3000 inserts * 2) when I load my initial data from an api.

In the PR i've added a `write_to_columns_mixins` builder option which generates a mixin that can be added to custom row classes and prevents these unnecessary class creations. This can also be used when using `@UseRowClass`.

simolus3:
I agree with the ergonomics of this. Let's also deprecate the `generateInsertable` parameter on `@UseDataClass` with a message suggesting to use the new build option instead.

North101:
A couple of thoughts:

- We could always generate the mixin and apply it to the data classes.
- I'm duplicating code for the toColumns method from `data_class_writer.dart`. Maybe there is a better place to put this code?
- Currently the name is `${table.nameOfRowClass}Columns` but maybe theres a better name for it? `${table.baseDartName}ToColumns`?

simolus3 (approved):
> Maybe there is a better place to put this code?

I've just moved it into an extension shared between the two writers - the method only really needs a list of columns anyway.

> maybe theres a better name for it? `${table.baseDartName}ToColumns`?

I like `ToColumns` as a suffix :+1: 

Looks good to me, thanks for all your contributions!

PR #2246, opened 2022-12-31
Fixes for schema generate

North101:
- Fixes writing `toColumns` mixin when `writeDataClasses` is disabled.
- Fixes writing `toCompanions` method when `writeCompanions` is disabled.
- Fixes not writing `@override` when using `toColumns` mixin.
- Fixes writing companion with row type that doesn't exist when `writeCompanions` is enabled and `writeDataClasses` is disabled. `dymamic` is used instead.
- Sorts schema versions


North101:
Not quite sure why the last change caused that test to reorder its imports

simolus3 (approved):

PR #2259, opened 2023-01-09
Update faq.md

constantinLu:


simolus3:
Thanks for the PR! I think there's no point in referencing the old package, I'll just recommend the current one.

simolus3 (approved):

PR #2276, opened 2023-01-23
Fix bitwiseAnd generated operation

tenhobi:
According to https://github.com/simolus3/drift/blob/2c63c1a64e6288e7fea6bf745fff172d63bdd3cc/drift/test/database/expressions/bitwise_test.dart#L23, the result of `bitwiseAnd` should be `&`.


simolus3 (approved):
Good catch, thanks!

Issue #1235, opened 2021-05-29
Query does not complete if the database is created outside WidgetTester.runAsync


Mike278:
Sorry Simon, I've got another one! These tests pass with moor 3.4.0 and flutter 1.20.4, but tests 3 and 4 do not complete with moor 4.3.2 and flutter 2.2.1. I know in general you're supposed to mock or avoid async in widget tests, so this may be a case of "that should have never worked", but it'd at least be good to understand what's going on here. 

```dart
import 'dart:async';

import 'package:flutter_test/flutter_test.dart';
import 'package:moor/ffi.dart';
import 'package:moor/moor.dart';

class TestDb extends GeneratedDatabase {
  TestDb() : super(SqlTypeSystem.withDefaults(), VmDatabase.memory());
  @override
  final List<TableInfo> allTables = const [];
  @override
  final int schemaVersion = 1;
}

Future<int> selectOne(TestDb _db) => _db
    .customSelect('select 1 a')
    .map((row) => row.read<int>('a'))
    .getSingle();

void main() {
  moorRuntimeOptions.dontWarnAboutMultipleDatabases = true;

  testWidgets('1', (tester) async {
    final db = TestDb();
    expect(await selectOne(db), 1); // completes
  });

  testWidgets('2', (tester) async {
    await tester.runAsync(() async {
      final db = TestDb();
      expect(await selectOne(db), 1); // completes
    });
  });

  testWidgets('3', (tester) async {
    final db = TestDb();
    await tester.runAsync(() async {
      expect(await selectOne(db), 1);  // does not complete
    });
  });

  testWidgets('4', (tester) async {
    final db = await tester.runAsync(() async {
      return TestDb();
    });
    expect(await selectOne(db), 1); // does not complete
  });
}

```

simolus3:
Sorry, I tried to investigate what's going wrong here but I'm pretty much clueless.

I think it has something to do with the custom zones used by Flutters test library. I've added some print statements to the moor's mutex where the deadlock is appearently coming from (I already don't like where this is going):

```diff
diff --git a/moor/lib/src/utils/synchronized.dart b/moor/lib/src/utils/synchronized.dart
index aebe8b6b..bd6a62bf 100644
--- a/moor/lib/src/utils/synchronized.dart
+++ b/moor/lib/src/utils/synchronized.dart
@@ -7,22 +7,29 @@ class Lock {
   /// Waits for previous [synchronized]-calls on this [Lock] to complete, and
[...]

PR #2297, opened 2023-02-04
DON'T MERGE Widget test with memory database deadlock

davidmartos96:
Hello!
I've recently encountered with an issue with drift when doing a widget test.
With a `NativeDatabase.memory` I get a deadlock when trying to cleanup the test. `db.close()` gets called and it hangs forever.
After some debugging it looks like the `Lock` from `synchronized` in the method `close()` is never entered.

In this "PR" I've included a failing simple test and a few logs to help debugging. 
I tried to debug the problem, but I'm not sure why it doesn't obtain the lock.

Here are the logs from running the test:

```
❯ flutter test
00:01 +0: deadlock test                                                                                               
Opening - 3703 - lock requested
Opening - 3703 - lock obtained
Drift: Sent CREATE TABLE IF NOT EXISTS "users" ("id" INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT, "name" TEXT NOT NULL, "birth_date" INTEGER NOT NULL, "profile_picture" BLOB NULL); with args []
Opening - 3703 - will release lock
Drift: Sent SELECT 1 with args []
closing db
Closing - 9536 - lock requested

(HERE IT HANGS)
```

Flutter version 3.7.1

### DIFF
--- a/drift/lib/src/runtime/executor/helpers/engines.dart
+++ b/drift/lib/src/runtime/executor/helpers/engines.dart
@@ -1,4 +1,5 @@
 import 'dart:async';
+import 'dart:math';
 
 import 'package:drift/drift.dart';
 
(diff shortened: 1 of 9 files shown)


davidmartos96:
After further investigation, it looks like this requires wrapping the test with `tester.runAsync`. That solves the problem.
In fact, running that test with a regular `test` and not `testWidgets` it works as expected.

This is probably not a bug in drift. 

I'll close this.

Mike278:
Sounds similar to #1235

davidmartos96:
> Sounds similar to #1235

Indeed. Looks like it's the same issue. Thank you for the link.

Maybe it would be worth adding a note on the drift Testing docs regarding widget tests in particular.

Unfortunately it's really tricky to know what can cause the lock if one doesn't know about `runAsync`

PR #2308, opened 2023-02-08
Remove hackyvariables from Postgres tests

davidmartos96:
I took a stab at improving the support for the Postgres dialect in the integration tests.

I fixed the variable binding test which were being skipped, but other issues remain which I will briefly mention.

One is the query `amountOfGoodFriends` which uses `WHERE f.really_good_friends = 1`.
That expression is invalid in postgres, as a bool is being compared with an int. Do you have any idea in mind with how to solve this?

The other is nested transaction support. Is there any reason why the Postgres implementation is not using the same logic as the SAVEPOINTs in the nested transactions for regular drift + sqlite? 

I've encountered with https://github.com/isoos/postgresql-dart/issues/85 so I was looking for a way to handle nested transactions at the library level.

As for the PR, casting the SELECT expression in the test was the main fix, but for UInt8List it's not enough, as the output is `"{1,2,3,..}"` (with curly braces) instead of the bytes 1,2,3...

### DIFF
--- a/drift/lib/src/runtime/query_builder/expressions/variables.dart
+++ b/drift/lib/src/runtime/query_builder/expressions/variables.dart
@@ -74,15 +74,23 @@ class Variable<T extends Object> extends Expression<T> {
     var explicitStart = context.explicitVariableIndex;
 
     var mark = '?';
+    var suffix = '';
     if (context.dialect == SqlDialect.postgres) {
       explicitStart = 1;
       mark = '@';
+
+      if (value is List<int>) {
+        // We need to explicitly bind the variable as byte array. Otherwise
+        // a Uint8List like [1,2,3] would bind to "{1,2,3}" as bytes
+        suffix += ":bytea";
+      }
     }
 
     if (explicitStart != null) {
       context.buffer
         ..write(mark)
-        ..write(explicitStart + context.amountOfVariables);
+        ..write(explicitStart + context.amountOfVariables)
+        ..write(suffix);
       context.introduceVariable(
         this,
         mapToSimpleValue(context),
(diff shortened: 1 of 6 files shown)


simolus3:
Thanks for the contribution! Recently, I helped with the postgres package to implement a new API that we can use to specify types without explicitly encoding them in the statement. This will hopefully make variables in postgres much more reliable. I have done some work on integrating that in drift in the `postgres3` branch. Since it will probably take a while to get that released though, I think this is a good improvement in the meantime.

For the `amountOfGoodFriends` query, we should just use `TRUE` instead of `1`.

> Is there any reason why the Postgres implementation is not using the same logic as the SAVEPOINTs in the nested transactions for regular drift + sqlite?

I think it was broken in some older versions of the postgres package, which encountered an illegal internal state when doing custom transaction management. We can probably remove the postgres-specific transaction implementation now.

davidmartos96:
@simolus3 I wasn't aware that the v3 API would support binding types without encoding, that's great to hear.

> For the amountOfGoodFriends query, we should just use TRUE instead of 1.

I changed the 1 to TRUE to make it compatible.

> I think it was broken in some older versions of the postgres package, which encountered an illegal internal state when doing custom transaction management. We can probably remove the postgres-specific transaction implementation now.

I changed the Transaction delegate to be the `NoTransactionDelegate`. Would that be the one that lets Drift use Savepoints internally?

Only 3 tests are failing now due to syntax error in the `friendsOf` query.

simolus3 (approved):
> I changed the Transaction delegate to be the NoTransactionDelegate. Would that be the one that lets Drift use Savepoints internally?

Exactly!

I managed to fix the remaining tests with the postgres v3 API, but since it's not clear how long it takes for that to get stable and released, this is a welcome improvement. Thanks again for your contribution!

davidmartos96:
@simolus3 You are welcome! Thanks to you, your work on the drift ecosystem, superb as always 😄

PR #2316, opened 2023-02-15
fix parsing of nested cte's

FaFre:
Nested CTE's are currently throwing an generator error.

```
nestedCte:
  SELECT
    *
    FROM
      (
	      WITH cte AS (
	          SELECT 1 AS val
	      )
	      SELECT * from cte
      )
```

Resulting error: `Null check operator used on a null value`


simolus3 (approved):
Thanks!

Issue #1149, opened 2021-04-20
Provide a way to access auto-generated rowid on dataclasses


micimize:
From the [sqlite docs](https://www.sqlite.org/lang_createtable.html#rowid):
> The rowid (and "oid" and "_rowid_") is omitted in WITHOUT ROWID tables. WITHOUT ROWID tables are only available in SQLite version 3.8.2 (2013-12-06) and later. A table that lacks the WITHOUT ROWID clause is called a "rowid table".

This poses an issue if one wants to access `rowid` on a table with a custom `primaryKey`, but not as a user-defined column

simolus3:
FWIW this has gotten a bit better now:

- In selects, you can access the [rowid column](https://drift.simonbinder.eu/api/drift/rowidextension/rowid) in Dart and add it via a join or `selectOnly`.
- Recently (see the linked PR), the `rowid` would also be exposed on companions so it can be set to custom values for inserts or updates.

micimize:
@simolus3 Is there a concise way to use `selectOnly` to get a `select rowId, * from table`?

That said, I don't entirely remember what I wanted `rowId` for in  and haven't touched the Flutter app I was working on in a while, so I can't say if that solves my issue. Though, I'm guessing I explicitly wanted to access `rowId` from companions.

simolus3:
> Is there a concise way to use `selectOnly` to get a `select rowId, * from table`?

`select(table).addColumns([table.rowId])` should work. You'll get a `TypedResult` back on which you can use `readTable(table)` to get the row class for the table and `read(table.rowid)` to get its rowid.

PR #2318, opened 2023-02-17
Expose rowid in results when using FTS5 with content table

FaFre:
It is required to provide the rowid when inserting elements into FTS5 (referring the entry in the referenced table). So it is required to expose rowid.

### DIFF
--- a/sqlparser/lib/src/engine/module/fts5.dart
+++ b/sqlparser/lib/src/engine/module/fts5.dart
@@ -122,7 +122,8 @@ class Fts5Table extends Table {
   }) : super(
           name: name,
           resolvedColumns: [
-            if (contentTable != null && contentRowId != null) RowId(),
+            if (contentTable != null && contentRowId != null)
+              RowId(includedInResults: true),
             ...columns,
             _Fts5RankColumn(),
             _Fts5TableColumn(name),
--- a/sqlparser/lib/src/analysis/schema/column.dart
+++ b/sqlparser/lib/src/analysis/schema/column.dart
@@ -183,10 +183,11 @@ class ViewColumn extends Column with DelegatedColumn implements ColumnWithType {
 class RowId extends TableColumn {
   // note that such alias is always called "rowid" in the result set -
   // "SELECT oid FROM table" yields a sinle column called "rowid"
-  RowId() : super('rowid', const ResolvedType(type: BasicType.int));
+  RowId({this.includedInResults = false})
+      : super('rowid', const ResolvedType(type: BasicType.int));
 
   @override
-  bool get includedInResults => false;
+  final bool includedInResults;
 }
 
 /// A column that is created by an expression. For instance, in the select
--- a/sqlparser/test/engine/module/fts5_test.dart
+++ b/sqlparser/test/engine/module/fts5_test.dart
@@ -34,9 +34,10 @@ void main() {
           .read(result.root as TableInducingStatement);
 
       expect(table.name, 'foo');
-      expect(table.resultColumns,
-          [isA<Column>().having((c) => c.name, 'name', 'bar')]);
-      expect(table.findColumn('oid'), isA<RowId>());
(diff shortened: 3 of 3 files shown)


sqlparser/lib/src/engine/module/fts5.dart
```diff
@@ -122,7 +122,8 @@ class Fts5Table extends Table {
   }) : super(
           name: name,
           resolvedColumns: [
-            if (contentTable != null && contentRowId != null) RowId(),
+            if (contentTable != null && contentRowId != null)
+              RowId(includedInResults: true),
```
simolus3, line 126: This will also have the effect of including the `rowid` column when expanding `*` columns. From what I can see, that's not correct though:  `SELECT * FROM fts_table_with_external_content` does not include the rowid column, I need to be explicit and say `SELECT rowid, * FROM ...`.

Do you perhaps have a statement that doesn't work at the moment as an example? Maybe we need to expose this rowid column in a different way.

FaFre: I need this column to insert entries into the fts table, where I need to address the rowid colum directly like:
```sql
INSERT INTO fts(rowid, name) VALUES(0, 'test');
```
The inserted rowid links the entry to the entry with same rowid in the content table.

simolus3: It looks like drift is able to resolve the rowid correctly though, this statement is analyzed fine:
https://github.com/simolus3/drift/blob/039ff942f77888adce75b406234ecb4fdbd60264/examples/modular/lib/src/search.drift#L16-L18

FaFre: Yes, I also use a trigger `ON DELETE` on my content table and that works for the analyzer. For the insertion I use dart logic to insert the entries manually into the fts table via the generated `DataClass`. There the `rowid` is only available when setting the `includedInResults` to `true`. Sorry for my unclear previous answer.

simolus3:
Ah I think I understand you now. However, we still can't let the `rowid`  be `includedInResults`: Otherwise, we'd expect the `rowid` to show up when we do `SELECT * FROM table` - and it doesn't! So we'd generate invalid code for those queries.

In 5166e5d29484747846f253adb2329429bfa3d65c, I've added the `rowid` as a field on companions. So you when you do an insert with companions, you can set the `rowid` to the desired value (see the tests in that commit for an example).
Should we also expose the rowid on generated row classes? Ideally, we probably should, but there are two problems in the way:

1. We need to adapt the query builder, so that doing a `select` on a table does `SELECT rowid, *` instead of `SELECT *`.
2. For compiled queries in drift files, we return an instance of the generated row class if all columns match. So doing `SELECT * FROM table` will return an instance of `TableData`. But if `TableData` has a `rowid` column, where would that come from? So we'd have to transform the query at compile time to add the `rowid` column.

So unfortunately, even though this PR looks simple, I think it would break code when selecting from fts5 tables. Properly adding the rowid to those row classes is more complicated (contributions towards that are welcome though!), but you can now use companions to set the rowid.

FaFre:
Generally all non-virtual tables that are not explicitly created with `WITHOUT ROWID` and do not  declare an `INTEGER PRIMARY KEY` do have a rowid (https://www.sqlite.org/rowidtable.html & https://www.sqlite.org/lang_createtable.html#rowid). I for my part have written custom SQL's when working with rowid in selects and joins (like for fts). The insert was the only case I could not find a workaround (working witch batch-inserts). My problem is solved here and I think working with rowid besides fts will be a very rare occasion for most user, as there weren't any other requests in the issues so far besides #1149 that only got two upvotes. 

Thank you a lot :)

FaFre:
For some reasons no rowid is added to my companions. I generate from drift file and have two options set in my `build.yaml`:

```yaml
options:
          skip_verification_code: true
          named_parameters: true
```

PR #2332, opened 2023-02-28
Fix `schema generate` instruction for migration tests

gnprice:
Without these flags, following the example code below leads to an error:

    Unsupported operation: TableInfo.map in schema verification code

because the generated code has a stub implementation of `map`.

The README in `examples/migrations_example`, which has very similar test code, says to use these flags.  So use them here.

### DIFF
--- a/docs/pages/docs/Advanced Features/migrations.md
+++ b/docs/pages/docs/Advanced Features/migrations.md
@@ -396,7 +396,11 @@ This can be used to insert data before a migration. After the migration ran, you
 Note that you can't use the regular database class from you app for this, since its data classes always expect the latest
 schema. However, you can instruct drift to generate older snapshots of your data classes and companions for this purpose.
 To enable this feature, pass the `--data-classes` and `--companions` command-line arguments to the `drift_dev schema generate`
-command.
+command:
+
+```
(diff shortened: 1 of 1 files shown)


simolus3:
Thanks! The documentation mentions those flags in the [verifying data integrity](https://drift.simonbinder.eu/docs/advanced-features/migrations/#verifying-data-integrity) section below, but I agree that it should be more prominent. On the other hand, they're not necessary for everyone, so maybe we shouldn't recommend them in the default command.

Perhaps we could point out these flags directly under there by writing something like "If you need older versions of generated data classes and companions, for instance because you want to run queries against older versions of your schema in a migration, you can pass the `--data-classes --companions` option to  `schema generate`"?

gnprice:
> The documentation mentions those flags in the [verifying data integrity](https://drift.simonbinder.eu/docs/advanced-features/migrations/#verifying-data-integrity) section below,

Ah, I see! That's the section I was referring to for writing the test code, but I'd completely glossed over the mention of those flags in the text.

In that case I think the right fix is probably to just provide a version of the command that has those flags, as a complete command in a code block, like the other `schema generate` command is. That'll make it a lot more conspicuous for readers scanning a bit too quickly, as I was. Just pushed a revised version that does that.

simolus3 (approved):
Thanks!

PR #2343, opened 2023-03-11
Fix melos package filters

blaugold:
As part of Melos 3.0.0 we changed all configuration options to camel case to be more consistent and renamed some options to be more self documenting.


simolus3 (approved):
Thanks!

PR #2349, opened 2023-03-14
Upsert constraint support DO NOTHING

FaFre:
This PR adds `ON CONFLICT (...) DO NOTHING` support for upserts.

Yes, there is `INSERT OR IGNORE INTO` but this will swallow other errors too:
`... No error is returned for uniqueness, NOT NULL, and UNIQUE constraint errors when the IGNORE conflict resolution algorithm is used` (https://www.sqlite.org/lang_conflict.html)

With `DO NOTHING` we explicitly target constraint errors.


simolus3 (approved):
Thanks! Could you add a changelog entry as well?

FaFre:
@simolus3 done!

simolus3:
Perfect, thanks for your contribution!

PR #2357, opened 2023-03-25
Update docu for drift file imports

FaFre:
First of all, is this a bug or a feature? :D


simolus3:
> First of all, is this a bug or a feature? :D

I think I've changed my mind about this a couple of times already, but I think it'd be more intuitive if this just worked. a6549425ef364eaafc8e1494eb8ca3be469ed75 should fix this.

FaFre:
Great! Can confirm, it is working now without adding in the annotation.

PR #2377, opened 2023-04-07
Expose CASE WHEN without base expression

### LINKED ISSUE #2376: CASE WHEN without base expression
> There are two fundamental forms of the CASE expression: those with a base expression and those without.
https://www.sqlite.org/lang_expr.html#the_case_expression

`CASE WHEN` without the base expression is not available.
I suppose that this was addressed in [this](https://github.com/simolus3/drift/issues/806) issue, but it was misunderstood.

Possible solutions are:
- Make base expression nullable in`CaseWhenExpression`
- Introduce totally new class `CaseWhenExpression`, and make the old one internal-only since it is called from the expression and must be hidden.
- Just remove `@internal` annotation and make constructor more straightforward (i.e. named parameters)

Result class example:
```

import '../query_builder.dart';

/// A `CASE WHEN` expression in sqlite.
///
/// This class supports when expressions with or without a base expression.
class CaseWhenExpression<T extends Object> extends Expression<T> {
  /// The optional base expression. If it's set, the keys in [whenThen] will be
  /// compared to this expression.
  final Expression? base;

  /// The when entries for this expression. This expression will evaluate to the
  /// value of the entry with a matching key.
  final List<MapEntry<Expression, Expression>> whenThen;

  /// The expression to use if no entry in [whenThen] matched.
  final Expression<T>? orElse;

  /// Creates a `CASE WHEN` expression from the independent components.
  CaseWhenExpression({this.base, required this.whenThen, this.orElse}) {
    if (whenThen.isEmpty) {
      throw ArgumentError.value(whenThen, 'when', 'Must not be empty');
    }
  }

  @override
  void writeInto(GenerationContext context) {
    context.buffer.write('CASE ');
    base?.writeInto(context);

    for (final entry in whenThen) {
      context.buffer.write(' WHEN ');
      entry.key.writeInto(context);
      context.buffer.write(' THEN ');
      entry.value.writeInto(context);
    }

    final orElse = this.orElse;
    if (orElse != null) {
      context.buffer.write(' ELSE ');
      orElse.writeInto(context);
    }

    context.buffer.write(' END');
  }
}

```

MichaelDark:
Resolves #2376

### DIFF
--- a/drift/lib/src/runtime/query_builder/expressions/case_when.dart
+++ b/drift/lib/src/runtime/query_builder/expressions/case_when.dart
@@ -4,18 +4,22 @@ import '../query_builder.dart';
 
 /// `CASE WHEN` expression **without** base expression in sqlite.
 ///
-/// This expression evaluates to the first value in [CaseWhenExpression.when]
+/// This expression evaluates to the first value in [BaseCaseWhenExpression.orderedCases]
 /// for which its key evaluates to `true`.
 ///
 /// See [BaseCaseWhenExpression] for more info.
-class CaseWhenExpression<R extends Object> extends BaseCaseWhenExpression<R> {
+class CaseWhenExpression<R extends Object>
+    extends BaseCaseWhenExpression<bool, R> {
   /// Creates a `CASE WHEN` expression from the independent components.
   ///
-  /// If [when] is empty - throws [ArgumentError]
+  /// This expression evaluates to the first value in [cases]
+  /// for which its key evaluates to `true`.
+  ///
+  /// If [cases] is empty - throws [ArgumentError]
   CaseWhenExpression({
-    required Map<Expression<bool>, Expression<R>> when,
+    required List<CaseWhen<bool, R>> cases,
     Expression<R>? orElse,
-  }) : super._(null, when.entries.toList(), orElse);
+  }) : super._(null, cases, orElse);
 }
 
 /// `CASE WHEN` expression **with** base expression in sqlite.
@@ -25,20 +29,36 @@ class CaseWhenExpression<R extends Object> extends BaseCaseWhenExpression<R> {
 /// See [BaseCaseWhenExpression] for more info.
 @internal
 class CaseWhenExpressionWithBase<T extends Object, R extends Object>
-    extends BaseCaseWhenExpression<R> {
+    extends BaseCaseWhenExpression<T, R> {
   /// Creates a `CASE WHEN` expression from the independent components.
   ///
-  /// If [when] is empty - throws [ArgumentError]
+  /// If [cases] is empty - throws [ArgumentError]
   CaseWhenExpressionWithBase(
     Expression<T> base, {
-    required Map<Expression<T>, Expression<R>> when,
+    required List<CaseWhen<T, R>> cases,
     Expression<R>? orElse,
-  }) : super._(base, when.entries.toList(), orElse);
+  }) : super._(base, cases, orElse);
+}
+
+/// A single when-then case for `CASE WHEN` expression in sqlite.
+///
+/// TODO check if it should be replaced with multiple return values.
+///
+/// [Multiple return values GitHub issue](https://github.com/dart-lang/language/issues/68)
+class CaseWhen<T extends Object, R extends Object> {
+  /// Expression to use after `WHEN`
+  final Expression<T> when;
+
+  /// Expression to use after `THEN`
+  final Expression<R> then;
+
+  /// Creates a single case with expression for when and expression for then.
+  const CaseWhen(this.when, {required this.then});
 }
 
 /// `CASE WHEN` expression in sqlite.
 ///
-/// This expression evaluates to the first value in [BaseCaseWhenExpression.whenThen]
+/// This expression evaluates to the first value in [BaseCaseWhenExpression.orderedCases]
 /// for which its key evaluates to `true`.
 ///
 /// Supports `CASE WHEN` expression with base expression and without base expression.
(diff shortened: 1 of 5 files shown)


drift/lib/src/runtime/query_builder/expressions/case_when.dart
```diff
@@ -1,30 +1,56 @@
...
+/// A `CASE WHEN` expression (with base expression) in sqlite.
+class CaseWhenExpressionWithBase<T extends Object, R extends Object>
+    extends _CaseWhenExpression<R> {
+  /// Creates a `CASE WHEN` expression from the independent components.
+  CaseWhenExpressionWithBase(
+    Expression<T> base, {
+    required Map<Expression<T>, Expression<R>> when,
+    Expression<R>? orElse,
+  }) : super._(base, when.entries.toList(), orElse);
+}
```
MichaelDark: Separate classes are needed because of the type system.
For variant WITHOUT base - we do not care about the type of `when` expression (or maybe explicitly specify it to `Expression<bool>`).
For variant WITH base - `when` expressions type must match `base`'s type.

drift/lib/src/runtime/query_builder/expressions/case_when.dart
```diff
@@ -1,30 +1,56 @@
...
+class CaseWhenExpression<R extends Object> extends _CaseWhenExpression<R> {
+  /// Creates a `CASE WHEN` expression from the independent components.
+  CaseWhenExpression({
+    required Map<Expression, Expression<R>> when,
+    Expression<R>? orElse,
+  }) : super._(null, when.entries.toList(), orElse);
+}
+
+/// A `CASE WHEN` expression (with base expression) in sqlite.
+class CaseWhenExpressionWithBase<T extends Object, R extends Object>
```
simolus3, line 29: If we have this as a separate class, I think it shouldn't be exposed - the `caseMatch` extension suffices.

drift/lib/src/runtime/query_builder/expressions/case_when.dart
```diff
@@ -1,30 +1,56 @@
-import 'package:meta/meta.dart';
-
 import '../query_builder.dart';
 
+/// A `CASE WHEN` expression (without base expression) in sqlite.
+class CaseWhenExpression<R extends Object> extends _CaseWhenExpression<R> {
+  /// Creates a `CASE WHEN` expression from the independent components.
+  CaseWhenExpression({
+    required Map<Expression, Expression<R>> when,
```
simolus3: The keys should have a type of `Expression<bool>` here, right?

For `caseMatch`, we can sort of assume that the order doesn't matter, but for a when expression without a base, it really does. So maybe we should change the parameter of `when` to `List<MapEntry<Expression<bool>, Expression<R>>>` directly?

drift/lib/src/runtime/query_builder/expressions/case_when.dart
```diff
@@ -1,30 +1,56 @@
-import 'package:meta/meta.dart';
-
 import '../query_builder.dart';
 
+/// A `CASE WHEN` expression (without base expression) in sqlite.
```
simolus3: The documentation here should link `[Expression.caseMatch]` to mention the variant that does have a base expression.

It should also mention what `when` does, e.g. "this expression evaluates to the first value in [when] for which its key evaluates to `true`".

simolus3 (commented):
I like the idea of having a case expression without a base, thanks for the contribution :+1:

drift/lib/src/runtime/query_builder/expressions/case_when.dart
MichaelDark: `@internal` should me enough 👍🏻

drift/lib/src/runtime/query_builder/expressions/case_when.dart
MichaelDark: > The keys should have a type of Expression<bool> here, right?   

That's true, will be fixed.

> So maybe we should change the parameter of when.  

No, I think it is a bad idea because of the usage. 
Variant with `List`:
```dart
CaseWhenExpression<int>(when: [
  MapEntry(const CustomExpression("'id' IS 1"), const Constant(1)),
  MapEntry(const CustomExpression("'id' IS 2"), const Constant(2)),
]
```
vs variant with `Map`:
```dart
CaseWhenExpression<int>(when: {
  const CustomExpression("'id' IS 1"): const Constant(1),
  const CustomExpression("'id' IS 2"): const Constant(2),
}),
```

The order will be described in the doc

simolus3: My point is that `Map`s aren't guaranteed to keep their order when we call `.entries`. It happens that the default map implementation in Dart (`LinkedHashMap`) does preserve insertion order, but it's not guaranteed that map literals will always use that type. Users might also pass a different map (e.g. a `HashMap` they got from somewhere) which then uses another order and subtly breaks the generated SQL.

In the future, we'll be able to use records for something like:

```dart
CaseWhenExpression<int>(when: [
  (const CustomExpression("'id' IS 1"), const Constant(1)),
  (const CustomExpression("'id' IS 2"), const Constant(2)),
]),
```

Until then, I think we should use a list directly since the order of entries is important here (even though it makes the API a bit ugly to use at the moment),

MichaelDark:
Ok, I will convert it to list.
In future it might be wrapped with some builder so that the final usage
will be like:

```
caseExpression().when(expr1.then(value)).when(expr2.thenExp(expr3))…
```

But not in the scope of this PR :)
For now - I will just change type to list

drift/lib/src/runtime/query_builder/expressions/case_when.dart
```diff
@@ -2,35 +2,113 @@ import 'package:meta/meta.dart';
...
+  /// If it is set, the keys in [orderedCases] will be compared to this expression.
   final Expression? base;
 
-  /// The when entries for this expression. This expression will evaluate to the
-  /// value of the entry with a matching key.
-  final List<MapEntry<Expression, Expression>> whenThen;
+  /// The when-then entries for this expression.
+  ///
+  /// This expression will evaluate to the value of the entry with a matching key.
+  final List<CaseWhen<T, R>> orderedCases;
```
MichaelDark: The fact that the list will be processed in order is specified explicitly _in the base class_.

drift/lib/src/runtime/query_builder/expressions/case_when.dart
```diff
@@ -2,35 +2,113 @@ import 'package:meta/meta.dart';
...
+    }
 
-    for (final entry in whenThen) {
+    for (final specificCase in orderedCases) {
       context.buffer.write(' WHEN ');
-      entry.key.writeInto(context);
+      specificCase.when.writeInto(context);
       context.buffer.write(' THEN ');
-      entry.value.writeInto(context);
+      specificCase.then.writeInto(context);
```
MichaelDark, line 107: https://dart.dev/guides/language/effective-dart/design#consider-making-the-code-read-like-a-sentence

drift/lib/src/runtime/query_builder/expressions/case_when.dart
```diff
@@ -2,35 +2,113 @@ import 'package:meta/meta.dart';
...
+/// [Multiple return values GitHub issue](https://github.com/dart-lang/language/issues/68)
+class CaseWhen<T extends Object, R extends Object> {
+  /// Expression to use after `WHEN`
+  final Expression<T> when;
+
+  /// Expression to use after `THEN`
+  final Expression<R> then;
+
+  /// Creates a single case with expression for when and expression for then.
+  const CaseWhen(this.when, {required this.then});
```
MichaelDark, line 52: then is marked as a named parameter to mark explicitly that the last expression specified in constructor is used after `THEN`

simolus3 (approved):
Looks good to me now, I really like the idea with the `CaseWhen` class. Thanks!

MichaelDark:
@simolus3 It would be great to have this feature as a minor version upgrade by Monday morning ;)

FYI I am planning to add a small `AS` expression in the next PR, probably tomorrow:
```dart
CaseWhenExpression(…).as(“additional_column”)
```

It is needed for a case when `CASE WHEN` is used as an additional column in select statement:
```
SELECT name, title, CASE WHEN … END AS additional_column FROM …
```

It needs additional discussion anyway.

simolus3:
> It would be great to have this feature as a minor version upgrade by Monday morning ;)

That might be a little too soon, but my plan is to release a new version including your contribution in the next few days :)

>  is used as an additional column in select statement

When you add columns to a select statement voa `addColumn`, drift will actually pick an alias for you and you can read that colum via `TypedResult.read`. Did you encounter a case where that didn't work?

PR #2417, opened 2023-05-09
Require analyzer 5.12.0, use InvalidType and DynamicType.

scheglov:


simolus3:
Thanks for the help! But I think we arguably shouldn't extend `TypeVisitor` to begin with then: It's documentation states that the clients shouldn't extend or implement it and it's clear that it's not stable for us to use it that way.

So I've made similar changes in 45dc0994402419ba8309640bdc0e89fee472d702, but without using a visitor - so if there's a new type being added there's no breakage. I will get this released and onto `latest-release` asap.

_edit_: I've released `drift` and `drift_dev` versions 2.8.0 and updated the `latest-release` branch.

PR #2434, opened 2023-05-22
Prepared statements cache

### LINKED ISSUE #2390: Caching prepared statements
At the moment all statements are compiled just before they are getting executed. By caching the compiled statements we can get some performance benefit, especially for frequently executed ones and batches. Are there some arguments against this behavior?

davidmartos96:
Fixes #2390 

I took a stab at implementing a cache for the prepared statements. A few benchmarks are included to compare caching vs no caching.
These are the results on my machine. The query is run against an empty database so that the prepare statement time is measurable vs the execution time.
```
Running 100000 queries (cached prepared statements):     1116418.5 us;
Running 100000 queries (non-cached prepared statements): 2063917.0 us;
```
Measuring just the preparation time of a single query against a large database with a few joins I got around the following times:
```
No-cache: ~30 us each preparation
Cache: ~0 us after caching
```

I created it as a Draft as some questions remain open.
Which should be the disposing strategy? I've taken a look at other sqlite library implementations and some use a LRU cache with a max size ([rusqlite](https://github.com/rusqlite/rusqlite)) while others like https://github.com/diesel-rs/diesel have an unrestricted cache (Reasoning behind this decision https://github.com/diesel-rs/diesel/pull/279)

I opted for a LRU, using the google library `quiver`. If we take this route, we will need to create a PR on the quiver package to have some way of knowing when an item is removed from the map, so that `dispose` can be called.


drift/lib/src/sqlite3/database.dart
```diff
@@ -149,3 +182,37 @@ class _SqliteVersionDelegate extends DynamicVersionDelegate {
     return Future.value();
   }
 }
+
+class _PreparedStatementsCache {
+  final LruMap<String, CommonPreparedStatement> _cache = LruMap(
```
simolus3: I wonder if we can just use a `LinkedHashMap` for this instead of depending on `quiver`? (Where `entries.first` is the LRU and `entries.last` is the MRU)?

When we insert a new statement and the map is full, we'd remove the first key and dispose it. If it's already in the map, we need to remove it and add it again to move it to the back. But I might be overlooking something here that would cause this to break :D

simolus3 (commented):
Awesome, thanks for your contribution!

I think we might be able to roll our own very simple LRU cache with a fixed size if needed (happy to help if you want to). That would avoid the problem of statements being disposed without us realizing it.

drift/lib/src/sqlite3/database.dart
davidmartos96: That's a nice idea, I'll give it a try. Since we only need a few methods, and not the full Map interface it could be simple enough.

davidmartos96: I've added a custom LRU based on LinkedHashMap + a small test.

davidmartos96:
I'm not sure about the LRU size we can use. I don't know how to properly measure the memory usage.
rusqlite uses a 16 size LRU, whereas diesel goes unlimited.

Another question is the current statement preparation on the batch call. Should that go through the cache, or should we leave it as is? Prepare and dispose independently.

simolus3:
>  rusqlite uses a 16 size LRU, whereas diesel goes unlimited.

I think 64 is not unreasonable. To find an optimal number, we'd have to run benchmarks on real-world projects, but 64 doesn't sound too high.

Diesel has an unlimited cache, but allows individual SQL components (the equivalent of Drift's Dart API) decide whether they can be cached or not. This sounds a lot harder to get right than just picking a upper limit, especially since we'd have to duplicate this in `drift_dev` for compile-time analyzed queries based on SQL.

simolus3:
> Another question is the current statement preparation on the batch call. Should that go through the cache, or should we leave it as is? Prepare and dispose independently.

Again, probably something that would need realistic benchmarks to figure out :D Batches are designed for use cases where few statements are run many times, so the prepare cost should amortize within the batch operation itself.

simolus3 (approved):
Looks great, thanks for your contribution!

davidmartos96:
> Looks great, thanks for your contribution!

@simolus3 No problem!

PR #2459, opened 2023-06-05
fix: use async syntax to fix timeout when closing database in unit tests

### LINKED ISSUE #2458: Unit test timeouts after upgrading to `v2.8.1`
Unit test timeouts after upgrading to `v2.8.1`.
I investigated code and looks like the problem is that database timeouts when trying to close it.

Test:
```dart
    testWidgets(
      'no customers in DB - should show empty state, no fab, no searchbar',
      (tester) async {
        //deps.database = NativeDatabase.memory(logStatements: true)
        await initPump(tester);
        await tester.pumpAndSettle();
        expect(find.text(l10n.noCustomers), findsOneWidget);
        expect(find.byType(FloatingActionButton), findsNothing);
        expect(find.byType(SearchBar), findsNothing);
        await deps.database.close();
      },
    );

```
I checked what was changed, this is new code: 
https://github.com/simolus3/drift/blob/e481f29138a77d737aa78f0769bd286bf24609ab/drift/lib/src/sqlite3/database.dart#L105-L109

After changing this part to:
```dart
  @override
  @mustCallSuper
  Future<void> close() async {
    return _preparedStmtsCache.disposeAll();
  }
```

test passed, closing database action was successful. But it is quite strange that this impacts unit test execution. Should I create PR with this change or is a problem somewhere else?

sidlatau:
Closes #2458

### DIFF
--- a/drift/lib/src/sqlite3/database.dart
+++ b/drift/lib/src/sqlite3/database.dart
@@ -104,8 +104,8 @@ abstract class Sqlite3Delegate<DB extends CommonDatabase>
 
   @override
   @mustCallSuper
-  Future<void> close() {
-    return Future(_preparedStmtsCache.disposeAll);
+  Future<void> close() async {
+    _preparedStmtsCache.disposeAll();
   }
 
   /// Synchronously prepares and runs [statements] collected from a batch.


simolus3 (approved):
Thanks!

Issue #2464, opened 2023-06-06
Database locked with prepared statements cache


YeungKC:
When I upgraded to 2.8.1, I got an error about database locked, that wasn't there before, The detailed error is :

```

SqliteException(5): while executing, cannot commit transaction - SQL statements in progress, database is locked (code 5)
  Causing statement: COMMIT TRANSACTION, parameters: 
```

PS: Always happened with COMMIT TRANSACTION.

I found that it was caused by the #2434, the change inside is actually very simple, just caching commonly used statements, but when not using cache, there is a slight difference in logic compared to the previous code, specifically that `CommonPreparedStatement.dispose()` is not called.

https://github.com/simolus3/drift/commit/45353a7693c928f06d81b357428585514729c417#diff-acd603f6276d61fb2f09c053ff4d50f15ffcc3271ec14940c4d377554cc01e2dL121

When I directly changed this code add manually called `CommonPreparedStatement.dispose()` when cache was disabled, I did not get the database locked anymore.

Finally, I curious, is `CommonPreparedStatement.dispose()` not supposed to be called? if using cache, is it unnecessary?

simolus3:
The reason we're not disposing the statement directly is that the statement is potentially being cached to be re-used later, so we want to keep it around.

Can you enable `logStatements: true` and post the statements that happen inside the transaction?

@davidmartos96 Have you seen this error before or do you have some clues on how this might happen? I guess we might not be fully executing some other statements before.

I've messed up by forgetting to add this feature to the changelog, given that is has caused some problems I'm leaning towards a patch release that disables statement caching by default until the next minor drift release.

davidmartos96:
@simolus3 I did not see such error unfortunately. But I think the problem is that the individual stmt.dispose() was not added when the cache statements flag is false. That was a slip up on my part.

It would be great to add an extra test that fails with a non caching database.

Yes, maybe it can get delayed until the next update.

YeungKC:
> The reason we're not disposing the statement directly is that the statement is potentially being cached to be re-used later, so we want to keep it around.↳
> 
> Can you enable `logStatements: true` and post the statements that happen inside the transaction?↳
> 
> @davidmartos96 Have you seen this error before or do you have some clues on how this might happen? I guess we might not be fully executing some other statements before.↳
> 
> I've messed up by forgetting to add this feature to the changelog, given that is has caused some problems I'm leaning towards a patch release that disables statement caching by default until the next minor drift release.↳

[...]

PR #2465, opened 2023-06-06
Add individual stmt dispose when not caching statements

davidmartos96:
I've included the try/finally when not using the prepared statements cache.

@simolus3 Do you want to use this PR to change the default flag of `cachePreparedStatements` to false? 

I'm not sure what could cause the SQLITE_BUSY error from #2464 in this context.


simolus3:
Thanks for the quick PR! It looks like we've had the same thought, I've made the same changes in dd3f04bf09f8262348e748532844afc8304dca3f. I've also added the option for background databases and disabled it by default.

davidmartos96:
:sweat_smile: Yes, I've just seen your changes

PR #2510, opened 2023-07-06
Bugfix: Only remove available implementations entries if they're available

brian-superlist:
Hey, thanks so much for the wonderful library! It's an absolute fave of mine in the Flutter ecosystem. Incredibly well built, thank you for all of your effort.

At Superlist, we've just started using the new Web Support and I ran into one funny issue.

I added the headers to get `opfs` to work. Everything looked good while logged in. However, I then realized the headers block our Google Auth Popup from working, and so I had to remove them. Of course, that also meant we had to fall back from the `opfsShared` shared implementation to a `sharedIndexedDb` implementation.

However, after I removed the headers, the library fell back to an `inMemory` implementation instead of a `sharedIndexedDb` implementation. Why? Because the previous files "_existsInOpfs" even though that implementation was no longer available due to the change to the headers.

Therefore, the `List<WasmStorageImplementation> availableImplementations` went from:

```
[
    WasmStorageImplementation.inMemory,
   WasmStorageImplementation.sharedIndexedDb,
   WasmStorageImplementation.unsafeIndexedDb,
  ];
```

to an empty array. The fix: you should only remove entries in that array if the previous database exists *and* it's still an available implementation.

I've also added to the documentation for folks to be careful about using this with Google Auth popups and linked to the issue. Hope this helps make this awesome library a bit better, or also happy to hear about alternative ways to handle the problem if this is not a good solution.

### DIFF
--- a/docs/pages/docs/Other engines/web.md
+++ b/docs/pages/docs/Other engines/web.md
@@ -108,12 +108,21 @@ Unfortunately, there's no way (that I'm aware of) to add these headers onto `flu
 Drift will fall back to a (slightly slower) implementation in that case (see [storages](#storages)),
 but we recommend researching and enabling these headers in production if possible.
 
-Note that Safari 16 has an [unfortunate bug](https://bugs.webkit.org/show_bug.cgi?id=245346)
+{% block "blocks/alert" title="Downsides of COOP and COEP" color="danger" %}
+While these headers are required for the origin-private FileSystem Access API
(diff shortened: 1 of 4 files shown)


simolus3 (approved):
Thank you for the kind words and the bugfix! I agree with your solution.

Also thanks for adding a warning to the documentation, I've highlighted it with a warning block since this is likely a painful realization for developers otherwise (given that it doesn't happen with `flutter run`)

PR #2529, opened 2023-07-22
Fix Assertion failure during setting up router in example app

Amir-P:
Fixes the following assertion failure:

```
The following assertion was thrown building DefaultSelectionStyle:
Assertion failed:
file://.pub-cache/hosted/pub.dev/go_router-9.0.0/lib/src/parser.dart:63:12
parser.dart:63
routeInformation.state != null
is not true
The relevant error-causing widget was:
  MaterialApp
MaterialApp:file://drift/examples/app/lib/main.dart:33:26
main.dart:33
```


simolus3 (approved):
Huh, I don't remember that breaking. Thanks for the fix!

PR #2548, opened 2023-08-01
Adds isolate setup

### LINKED ISSUE #2545: Should NativeDatabase.createInBackground offer a preSetup parameter to call open.overrideFor(OperatingSystem.android, openCipherOnAndroid)?
I'm trying to combine the instructions of the getting started to open an encrypted database in an isolate. The problem is that `open.overrideFor(OperatingSystem.android, openCipherOnAndroid)` needs to be called in the isolate that handles the database. However, I haven't found a way to do that with `NativeDatabase.createInBackground`.

As a workaround, I ended up copying the code of `NativeDatabase.createBackgroundConnection` and `_NativeIsolateStartup` just to add that one line in `_NativeIsolateStartup.start`.

If you're ok, I'm happy to submit a PR to add an optional parameter (`override`, `preSetup`?) to that later, we can only do this:

```dart
NativeDatabase.createInBackground(
  File(...),
  setup: (db) { ... },
  override: () => open.overrideFor(OperatingSystem.android, openCipherOnAndroid),
)
```

GP4cK:
Closes https://github.com/simolus3/drift/issues/2545.

I don't know how to write a test for this since it's executed in a different isolate... I found one test for `createInBackground` but it doesn't test the `setup` function... Any tips?

### DIFF
--- a/drift/CHANGELOG.md
+++ b/drift/CHANGELOG.md
@@ -3,6 +3,7 @@
 - Add support for subqueries in the Dart query builder.
 - Add `isInExp` and `isNotInExp` to construct `IS IN` expressions with arbitrary
   expressions.
+- Add `isolateSetup` to `NativeDatabase.createInBackground()` to override libraries.
 
 ## 2.10.0
 
--- a/drift/lib/native.dart
+++ b/drift/lib/native.dart
@@ -9,6 +9,7 @@
 /// For more information other platforms, see [other engines](https://drift.simonbinder.eu/docs/other-engines/vm/).
 library drift.ffi;
 
+import 'dart:async';
 import 'dart:io';
 import 'dart:isolate';
 
@@ -39,7 +40,7 @@ typedef DatabaseSetup = void Function(Database database);
 /// ```
 /// open.overrideFor(OperatingSystem.android, openCipherOnAndroid)
 /// ```
-typedef IsolateSetup = void Function();
+typedef IsolateSetup = FutureOr<void> Function();
 
 /// A drift database implementation based on `dart:ffi`, running directly in a
 /// Dart VM or an AOT compiled Dart/Flutter application.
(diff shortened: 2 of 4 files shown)


drift/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 - Add support for subqueries in the Dart query builder.
 - Add `isInExp` and `isNotInExp` to construct `IS IN` expressions with arbitrary
   expressions.
+- Add `isolateSetup` to `NativeDatabase.createInBackground()` to override libraries.
```
GP4cK, line 6: I saw 2.11 is not yet published so I added this here.

drift/lib/native.dart
```diff
@@ -31,6 +31,16 @@ export 'package:sqlite3/sqlite3.dart' show SqliteException;
...
 
+/// Signature of a function that can perform setup work on the isolate before
+/// opening the database.
+///
+/// This could be used to override libraries.
+/// For example:
+/// ```
+/// open.overrideFor(OperatingSystem.android, openCipherOnAndroid)
+/// ```
+typedef IsolateSetup = void Function();
```
simolus3: I wonder if we should support asynchronous functions as well? We could allow returning a `FutureOr<void>`. I think making `_NativeIsolateStartup.start` async to await it should be fine, the communication is asynchronous either way.

simolus3 (commented):
> don't know how to write a test for this since it's executed in a different isolate

Maybe we could do something like this:

```dart
final receivePort = ReceivePort();
final sendPort = receivePort.sendPort;

NativeDatabase.createInBackground(
  ...,
  isolateSetup: () => sendPort.send('hi there'),
);
```

Then we can assert that the setup function was called by listening for messages on the receive port in the main isolate?

drift/lib/native.dart
GP4cK: Done in [Make IsolateSetup return FutureOr<void>](https://github.com/simolus3/drift/pull/2548/commits/6e94fd3694d830aa2ba5e8bc54350460bf9b2618)

simolus3 (approved):
Thanks for the contribution, LGTM!

PR #2554, opened 2023-08-09
Support listing and deleting wasm databases

simolus3:
This adds the `WasmDatabase.probe` API, which can be used to list existing databases and delete them. It can also be used to open a database with a specific storage implementation which can be useful for apps needing a higher degree of customization.

@FaFre Feel free to take a look [at the API](https://github.com/simolus3/drift/pull/2554/files#diff-be9f1be52a4a84dba71787a14fea605e8bbf79a62b87f0905a6c06d1dc759d3e) to see if would satisfy your requirements. Unfortunately we can't efficiently iterate through IndexedDb databases, so we can only give you a complete list of OPFS databases.

### DIFF
--- a/drift/lib/src/web/wasm_setup.dart
+++ b/drift/lib/src/web/wasm_setup.dart
@@ -23,6 +23,7 @@ import 'package:sqlite3/wasm.dart';
 
 import 'broadcast_stream_queries.dart';
 import 'channel.dart';
+import 'wasm_setup/shared.dart';
 import 'wasm_setup/protocol.dart';
 
 /// Whether the `crossOriginIsolated` JavaScript property is true in the current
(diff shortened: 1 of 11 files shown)


FaFre:
Wow! That's great :star_struck:
It perfectly fits my needs, thanks for recognizing my feedback.

To give some feedback after switching to this implementation:

- I created an extension to easily get my desired implementation:
```drift
extension _ on WasmProbeResult {
  WasmStorageImplementation? getPreferredImplementation() {
    if (availableStorages.contains(WasmStorageImplementation.opfsShared)) {
      return WasmStorageImplementation.opfsShared;
    } else if (availableStorages
        .contains(WasmStorageImplementation.opfsLocks)) {
      return WasmStorageImplementation.opfsLocks;
    }

    return null;
  }
}
```
- I also decided to cache `WasmProbeResult` to `open` several databases from this instance. I hope to squeeze a bit of performance out of this, because my users will swap databases quite often.

simolus3:
Caching the result is probably a good call - I don't want to do it in drift because the list of existing databases could have changed, but it good to do it when you know they haven't. We have to re-run the whole feature detection chain every time which tries to spawn workers, so it's not exactly cheap (doing it once is fine though).

PR #2565, opened 2023-08-14
remove leading slashes from URI paths

elitree:
In order to allow for drift web setups where the app is served from a subdirectory, the leading slashes have been removed from the Uri.parse() calls.

### DIFF
--- a/extras/integration_tests/web_wasm/README.md
+++ b/extras/integration_tests/web_wasm/README.md
@@ -3,5 +3,5 @@ Integration tests for `package:drift/native.dart`.
 To run the tests automatically (with us managing a browser driver), just run `dart test`.
 
 To manually debug issues, it might make sense to trigger some functionality manually.
-You can run `dart run tool/server_manually.dart` to start a web server hosting the test
-content on http://localhost:8080.
+You can run `dart run tool/serve_manually.dart` to start a web server hosting the test
+content on <http://localhost:8080>.
--- a/extras/integration_tests/web_wasm/web/main.dart
+++ b/extras/integration_tests/web_wasm/web/main.dart
@@ -11,8 +11,8 @@ import 'package:web_wasm/src/database.dart';
 import 'package:sqlite3/wasm.dart';
 
 const dbName = 'drift_test';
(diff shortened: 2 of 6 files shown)


extras/integration_tests/web_wasm/web/main.dart
```diff
@@ -82,18 +82,18 @@ Future<Uint8List?> _initializeDatabase() async {
 
       // Let's first open a custom WasmDatabase, the way it would have been
       // done before WasmDatabase.open.
-      final sqlite3 = await WasmSqlite3.loadFromUrl(Uri.parse('/sqlite3.wasm'));
+      final sqlite3 = await WasmSqlite3.loadFromUrl(Uri.parse('sqlite3.wasm'));
       final fs = await IndexedDbFileSystem.open(dbName: dbName);
       sqlite3.registerVirtualFileSystem(fs, makeDefault: true);
 
-      final wasmDb = WasmDatabase(sqlite3: sqlite3, path: '/app.db');
+      final wasmDb = WasmDatabase(sqlite3: sqlite3, path: 'app.db');
```
simolus3, line 89: I actually prefer that the path here and the one for `xOpen` stay absolute. These refer to the path within the virtual file system implementation and aren't resolved to anything by the browser.

(The default working directory we assume is `/` so both work, but this change is unrelated to the issue of not being able to load the worker and it's safer to use absolute paths)

extras/integration_tests/web_wasm/README.md
```diff
@@ -3,5 +3,5 @@ Integration tests for `package:drift/native.dart`.
 To run the tests automatically (with us managing a browser driver), just run `dart test`.
 
 To manually debug issues, it might make sense to trigger some functionality manually.
-You can run `dart run tool/server_manually.dart` to start a web server hosting the test
-content on http://localhost:8080.
\ No newline at end of file
+You can run `dart run tool/serve_manually.dart` to start a web server hosting the test
+content on <http://localhost:8080>.
```
simolus3, line 7: Nice, good catch!

simolus3 (commented):
Looks good, let's just keep using absolute paths for the `WasmDatabase` constructor and to use `xOpen`.

PR #2577, opened 2023-08-24
add a dependency from the default target to the custom target

jakemac53:
See https://github.com/dart-lang/build/pull/3556#issuecomment-1691965824 - this should fix the customer tests failure.

I will have to figure out how we want to roll this out longer term based on the breaking nature.

### DIFF
--- a/drift/build.yaml
+++ b/drift/build.yaml
@@ -10,6 +10,8 @@ targets:
     sources:
       exclude:
         - "test/generated/custom_tables.dart"
+    dependencies:
+      - :custom
     builders:
       drift_dev:
         options:


jakemac53:
Actually, it looks like I still get the weird analyzer error even with this change using the currently published version of `drift_dev`, but using a path dependency I do not.

simolus3:
So is `build_runner` supposed to only build the `$default` target now and not the others? Or is that the idea of your change? I've mainly split them into different targets because they're using different builder options, the idea was that they should be able to run concurrently.

If adding a dependency from the `$default` target is required now, I agree with these changes. It would be good to know if that's really the idea here though, because I'd have to fix a couple other `build.yaml` files in this repo.

jakemac53:
It will still build all the targets - this just defines the ordering more explicitly (there was some implicit default ordering previously).

Builders in different phases can in fact run concurrently, if they don't actually try to read files from one of the targets they depend on. But if they do try to read those files they will wait on them to be built.

In practice previously I think you may have been having some imports be unavailable, which happened to not break things, but could have caused some weird problems eventually 🤷 . This could technically reduce parallelism if you had some builders resolving code in the `default` target, running on files that imported the file produced by the other target. They will now wait for that file to be produced before completing.

jakemac53:
I did also make a change to the builder such that it will only warn and not error when it encounters unresolvable assets. But that will mean slower build times for anybody depending on your package and using a resolver (which may or may not actually matter). As well as probably some very spammy output 😬 . We might want to invest in some logic for de-duplication of logs actually.

But, you wouldn't be broken if you don't accept this PR and I land the other proposed changes at least.

simolus3:
Ah, now I understand what's happening here: `test/generated/data_classes.dart` is part of the default target, but imports a file with a part statement generated from the other target. So in that case, there _is_ a source-level dependency between the targets missing in the build configuration.

But I think for the drift package, moving that file to the custom target expresses the intent better, because there should be no dependencies between the targets. So thanks for taking the time to explain this to me! I'll close this PR in favor of 84e251121ee8d2f263514fc0c413a48fe0441dcd.

simolus3:
> Actually, it looks like I still get the weird analyzer error even with this change using the currently published version of `drift_dev`, but using a path dependency I do not.

While the published version of `drift_dev` is using the analyzer in a very inefficient way by manually crawling the import graph and resolving every library along the way to find potential definitions, I think this shouldn't crash the resolver or the analyzer.

It works with the path package because that version contains an optimization to use the resolver in more predictable ways (essentially only resolving the input id). I'll also publish that soon, but maybe there are other builders that might run into this crash in some situations.

PR #2581, opened 2023-08-26
fix code generation for `enum` with generics

nikitadol:
Sometimes you need to use `enum` with generics:
```dart
enum MyEnum<T> {
  ...
}
```

But in this case, the code generator generates the wrong code:
```dart
  static JsonTypeConverter2<MyEnum<Object?>, int, int> $convertername =
      const EnumIndexConverter<MyEnum<Object?>>(MyEnum<Object?>.values);
```

Since this part is not considered valid in Dart:
```dart
MyEnum<Object?>.values
```

Correct use:
```dart
MyEnum.values
```

This PR fixes this behavior


simolus3 (approved):
Good catch and thanks for the fix, LGTM!

PR #2599, opened 2023-09-07
drift_network_bridge_dev

Heinrich-vanNieuwenhuizen:
**drift.network_bridge**
Considering some scenarios where different database (especially sqlite, but not limited to) has to be synced or checked remotely with other databases. Likewise, cases where the devices should be able to work offline. An additional custom layer is usually introduced. I believe it would be better if this is integrated into the drift package removing the necessity for a custom layer.

Experimental bridge that will allow for network connectivity. 

Comes with an out-of-the-box mqtt implementation.
On the host device
```
  final db = Database(DatabaseConnection(NativeDatabase.memory(logStatements: true)));
  final gate = MqttDatabaseGateway('127.0.0.1', 'unit_device', 'drift/test_site');
  await gate.serve(db);
```
  
On the client's device
```
  final gw = MqttDatabaseGateway('127.0.0.1', 'unit_device', 'drift/test_site');
  final remoteDb = Database(await gw.createRemoteConnection());
```

**Areas of assistance required**
If this package is introduced. I think it will help if contributions can be made 

- On the network side, making the connection more robust and better error handling
- Other network protocols 
- Syncing and comparison of local to remote tables.


simolus3:
Syncing drift databases over the network is an awesome idea, thanks for sharing this!

> Likewise, cases where the devices should be able to work offline

The communication protocol in `package:drift/remote.dart` has not been designed with this use-case in mind, the idea is to provide a live database connection. This is also a very challenging problem to solve, especially when considering things like conflict resolution.

> I believe it would be better if this is integrated into the drift package removing the necessity for a custom layer.

If there's something that needs to change in the drift package to facilitate this, then that's definitely something worth considering. But I don't see how this fits into the core drift package, which is a essentially a database client.

Since your network implementation only imports public drift APIs and didn't require changes in drift, I think it would be a great contribution to have this as a standalone package. If that requires support from my side I'm happy to look at ways this could be made easier in drift, but I don't think it falls into the things that drift is trying to solve.

FaFre:
Wow that's really interresting! Is this also working with multiple clients reading and writing to the same database at the same time (1-n)?

Heinrich-vanNieuwenhuizen:
Hi @FaFre, the design is set up so that each client should create its own channel, making it technically feasible. I'll update you once we have a production-ready version. If issues arise, we're considering the use of mutexes as a fallback.

PR #2603, opened 2023-09-08
doc: add docs around moving to step-by-step migrations

### LINKED ISSUE #2601: Request: Docs for moving to stepwise migration tooling
It's not clear from docs the best way to move from older migration patterns to the new stepwise migration pattern. 

We chose not to use codegen for generating tables and data classes, so we largely use `.drift` files and raw SQL for tables, indexes, simple and complex queries. We also use our own data classes imported into the drift files.

Because of this - we didn't have schema generation or schema dumps prior to our current schema level.

We've gone back and added schema for our deployed version and the migrations / schema between the latest deployed. We've added the new stepwise migration steps, along with tests, and really like how it works - it's much closer to how we handle schema updates on other platforms.

The pattern we've adopted is to give a floor to the `from` value in `migrator.runMigrationSteps` that equals the current migration. This allows our "legacy" migration code to play out, then we start as though we are the latest schema and migrate stepwise from that point on.

If the below looks like a good pattern, adding it to the docs may be useful. Probably as a callout - "if you've already got a DB and looking to add the new stepwise migration" or similar. 

```dart
// Assuming you've already run your migration code up to schema 10 
// and you want to migrate to 11, 12, etc.. with stepwise, 
// pin the floor of the `from` value to 10.
await m.runMigrationSteps(
    from: math.max(10, from),
    to: to,
    steps: migrationSteps(...));
```

It could also be that you'd like to add some handling for this kind of case to the "runMigrationSteps" - like a `to10` step that handles cases before stepwise was added. Not sure this is better than the above - and I think the current functionality is fine considering this is all that is required to "move" to stepwise at a given schema point.

Happy to do the work to add to the docs, but didn't know if there was a need or opportunity to improve the docs or code here, which is why I'm opening this as a feature request.

dballance:
Resolves https://github.com/simolus3/drift/issues/2601

### DIFF
--- a/docs/lib/snippets/migrations/migrations.dart
+++ b/docs/lib/snippets/migrations/migrations.dart
@@ -1,3 +1,5 @@
+import 'dart:math' as math;
+
 import 'package:drift/drift.dart';
 
 // #docregion stepbystep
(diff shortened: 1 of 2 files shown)


dballance:
@simolus3 - hopefully this gives you a good starting point. Happy to make changes - will leave it to you to merge!

Thank you - and thank you for your efforts on this project!

simolus3 (approved):
Looks great, thanks for the docs and the snippet!

PR #2615, opened 2023-09-17
Add missing "not" in documentation for TypeConverter

paldepind:
I'm pretty sure this is what it's supposed to say 🙂


simolus3 (approved):
Right, good catch. Thanks!

PR #2617, opened 2023-09-18
Preventing getter name conflicts in step-by-step migrations

### LINKED ISSUE #2610: Step-by-step migration generation fails with specific table names
In my database schema I have a table called `entities`. When attempting to move to the new step-by-step style migrations, the generated code does not compile. This is because the generated schemas end up with two conflicting versions of a getter named `entities`: one inherited from `VersionedSchema` and one for my table of the same name. Here is a truncated example of the generated code to show the problem:

```dart
final class _S25 extends i0.VersionedSchema {
  _S25({required super.database}) : super(version: 25);
  @override
  late final List<i1.DatabaseSchemaEntity> entities = [
    campaigns,
    folders,
    entities,
    // ...
  ];
  // ...
  late final Shape26 entities = Shape26(
    // ...
  );
}
```

This issue will occur for tables with any of the following names:

- database
- version
- entities

### Officially disallow these table names

The current list of 3 forbidden table names isn't that restrictive, so it's always possible to say this is working as ~implemented~ intended and those table names just aren't supported. Personally if this is the case I would prefer if _all_ the Drift generators errored out or at least reported a warning if you used one of the reserved names. It would be a bit unfortunate to have everything working only to run into issues when you try to start using step-by-step migrations.

I'm also not sure this would fix my specific problem. I have ~25 versions of schema history I would like to migrate to the new step-by-step API. The old migrations need to continue to work, but I can't generate the step-by-step code for anything until I rename the `entities` table. Presumably I would need a way to make drift use old-style migrations up until the rename and then use step-by-step migrations afterwards. I think that's probably possible with the current set of APIs but I haven't fully investigated.

### Hide the drift getters in weird names

This is a common tactic I've seen in some Go libraries. Any fields that are reserved for the library but need to be public are given awkward, non-idomatic names to minimize the possibility of a user running into a conflict. So the three listed getters could be renamed to something like `DRIFT_database`, `DRIFT_version`, etc.

### Modify the table names to prevent conflicts

The code generator could modify the getter names to ensure there are never any conflicts. The simplest solution would probably be to just prepend or append some extra text. For example, always appending `Table` would result in my schema having `campaignsTable`, `foldersTable`, `entitiesTable`, etc.

A more complex solution would be to only modify the names of tables that would conflict with a reserved word. So my schema would become `campaigns`, `folders`, `entitiesTable`. This seems cleaner but requires more complex code to keep track of the modified getter names. Imagine I had an additional table called `entities_table`, so `entitiesTable` in camelCase. The generator would need to know that just appending `Table` to `entities` still produces a naming conflict and respond accordingly. I admit this is a bit contrived but if the generator is made more complex to solve this issue I think it makes sense to solve the issue for _every_ case.

One more option would be to generate an additional intermediate class that contains only the table getters. Something like this:

```dart
final class _S25Tables {
  const _S25Tables({
    required this.campaigns,
    required this.folders,
    required this.entities,
  });

  final Shape35 campaigns;
  final Shape17 folders;
  final Shape26 entities;
}

final class _S25 extends i0.VersionedSchema {
  _S25({required super.database}) : super(version: 25);
  @override
  late final List<i1.DatabaseSchemaEntity> entities = [
    tables.campaigns,
    tables.folders,
    tables.entities,
  ];

  late final tables = _S25Tables(
    campaigns: Shape35(
      //...
    ),
    folders: Shape17(
      //...
    ),
    entities: Shape26(
      //...
    ),
  );
}
```

This should solve the problem in every case (except for Dart reserved getters like `hashCode` but nothing to be done there).

Personally I like the nested `tables` object API the best, but would be happy with any of the modified generator options that make reserved table names no longer a thing. I'm happy to hack on the solution and open a PR, but wanted to get your feedback on the entire issue before going down the code generation rabbit hole.

jmatth:
Closes #2610.

I added a couple methods to `Scope` to keep track of which names have been used and allow the caller to specify how non-conflicting names should be calculated. As is this fixes the issue in my codebase with a table named `entities`.

In #2610 you mentioned:

> This is not only a problem with table names, but also with the getter names for columns potentially colliding with columns inherited from VersionedTable (and its superclasses).

I began looking into addressing that as well but as far as I can tell any name collisions there will fail while generating Drift code for runtime. Both `VersionedTable` and table the table classes generated for runtime extend `Table` and mixin `TableInfo`.

Are there tests for the code generation I should be updating or adding to? I see there are tests for various parts of the `drift_dev` package but couldn't find any that covered `SchemaVersionWriter`.

### DIFF
--- a/drift_dev/lib/src/writer/schema_version_writer.dart
+++ b/drift_dev/lib/src/writer/schema_version_writer.dart
@@ -325,6 +325,9 @@ class SchemaVersionWriter {
       final versionClass = '_S$versionNo';
       final versionScope = libraryScope.child();
 
+      // Reserve all the names already in use in [VersionedSchema] and its
+      // superclasses. Without this certain table names would cause us to
+      // generate invalid code.
       versionScope.reserveNames([
         'database',
         'entities',
--- a/drift_dev/lib/src/writer/schema_version_writer.dart
+++ b/drift_dev/lib/src/writer/schema_version_writer.dart
@@ -71,21 +71,6 @@ class SchemaVersionWriter {
   final Map<String, String> _columnCodeToFactory = {};
   final Map<_TableShape, String> _shapes = {};
 
-  // A list of member names already in use in the generated class. This is used
-  // to prevent table getters from conflicting with other class members.
-  final Set<String> _usedNames = {};
-
-  void _resetUsedNames() {
-    _usedNames.clear();
-    _usedNames.addAll([
-      'database',
-      'entities',
-      'version',
-      'stepByStepHelper',
-      'runMigrationSteps',
-    ]);
-  }
-
   SchemaVersionWriter(this.versions, this.libraryScope) {
     assert(versions.isSortedBy<num>((element) => element.version));
   }
@@ -173,15 +158,6 @@ class SchemaVersionWriter {
     });
   }
 
-  String _getNonCollidingGetterName(DriftSchemaElement element) {
-    var name = element.dbGetterName!;
-    while (_usedNames.contains(name)) {
-      name = '$name${_suffixForElement(element)}';
-    }
-    _usedNames.add(name);
-    return name;
-  }
-
   String _suffixForElement(DriftSchemaElement element) => switch (element) {
         DriftTable() => 'Table',
         DriftView() => 'View',
@@ -191,8 +167,10 @@ class SchemaVersionWriter {
       };
 
   String _writeWithResultSet(
-      DriftElementWithResultSet entity, TextEmitter writer) {
-    final getterName = _getNonCollidingGetterName(entity);
+    String getterName,
+    DriftElementWithResultSet entity,
+    TextEmitter writer,
+  ) {
     final shape = _shapeClass(entity);
     writer
       ..write('late final $shape $getterName = ')
@@ -298,19 +276,20 @@ class SchemaVersionWriter {
     required DriftSchemaElement element,
     required TextEmitter definition,
   }) {
-    String name;
+    final name = definition.parent!.getNonConflictingName(
+      element.dbGetterName!,
+      (name) => name + _suffixForElement(element),
+    );
 
     if (element is DriftElementWithResultSet) {
-      name = _writeWithResultSet(element, definition);
+      _writeWithResultSet(name, element, definition);
     } else if (element is DriftIndex) {
-      name = _getNonCollidingGetterName(element);
       final index = definition.drift('Index');
 
       definition
         ..write('final $index $name = ')
         ..writeln(DatabaseWriter.createIndex(definition.parent!, element));
     } else if (element is DriftTrigger) {
-      name = _getNonCollidingGetterName(element);
       final trigger = definition.drift('Trigger');
 
       definition
@@ -345,7 +324,14 @@ class SchemaVersionWriter {
       final versionNo = version.version;
       final versionClass = '_S$versionNo';
       final versionScope = libraryScope.child();
-      _resetUsedNames();
+
+      versionScope.reserveNames([
+        'database',
+        'entities',
+        'version',
+        'stepByStepHelper',
+        'runMigrationSteps',
+      ]);
 
       // Write an _S<x> class for each schema version x.
       versionScope.leaf()
--- a/drift_dev/lib/src/writer/schema_version_writer.dart
+++ b/drift_dev/lib/src/writer/schema_version_writer.dart
@@ -71,6 +71,21 @@ class SchemaVersionWriter {
   final Map<String, String> _columnCodeToFactory = {};
   final Map<_TableShape, String> _shapes = {};
 
+  // A list of member names already in use in the generated class. This is used
+  // to prevent table getters from conflicting with other class members.
+  final Set<String> _usedNames = {};
+
+  void _resetUsedNames() {
+    _usedNames.clear();
+    _usedNames.addAll([
+      'database',
+      'entities',
+      'version',
+      'stepByStepHelper',
+      'runMigrationSteps',
+    ]);
+  }
+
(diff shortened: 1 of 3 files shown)


drift_dev/lib/src/writer/writer.dart
```diff
@@ -298,6 +298,10 @@ class Scope extends _Node {
   /// This can be used to generated methods which must have a unique name-
   int counter = 0;
 
+  /// The set of names already used in this scope. Used by methos like
+  /// [getNonConflictingName] to prevent name collisions.
+  final Set<String> _usedNames = {};
```
simolus3, line 303: > I added a couple methods to `Scope` to keep track of which names have been used and allow the caller to specify how non-conflicting names should be calculated

Nice idea!

drift_dev/lib/src/writer/schema_version_writer.dart
```diff
@@ -314,6 +325,14 @@ class SchemaVersionWriter {
       final versionClass = '_S$versionNo';
       final versionScope = libraryScope.child();
 
+      versionScope.reserveNames([
```
simolus3, line 331: Let's add a small comment here explaining that these names are inherited from the schema class, which is why we want to prevent them from being used for schema entities.

drift_dev/lib/src/writer/writer.dart
```diff
@@ -298,6 +298,10 @@ class Scope extends _Node {
   /// This can be used to generated methods which must have a unique name-
   int counter = 0;
 
+  /// The set of names already used in this scope. Used by methos like
```
simolus3: Small typo:

```suggestion
  /// The set of names already used in this scope. Used by methods like
```

simolus3 (commented):
Thanks for the contribution, this looks great!

> Are there tests for the code generation I should be updating or adding to? I see there are tests for various parts of the `drift_dev` package but couldn't find any that covered `SchemaVersionWriter`.

Yeah, at the moment I'm using the generated code in `examples/migrations` and `examples/app` as an integration test for the writer, there aren't any unit tests for it. This is probably a good opportunity to get started with that, maybe something like this can serve as a starting point:

```dart
import 'package:drift_dev/src/analysis/options.dart';
import 'package:drift_dev/src/analysis/results/results.dart';
import 'package:drift_dev/src/writer/import_manager.dart';
import 'package:drift_dev/src/writer/schema_version_writer.dart';
import 'package:drift_dev/src/writer/writer.dart';

void main() {
  final fakeUri = Uri.parse('drift:hidden');

  test('avoids conflict with getters in schema class', () async {
    final imports = LibraryImportManager();
    final writer = Writer(
      const DriftOptions.defaults(),
      generationOptions: GenerationOptions(imports: imports),
    );
    imports.linkToWriter(writer);

    final table = DriftTable(
      DriftElementId(fakeUri, 'database'),
      DriftDeclaration(fakeUri, -1, ''),
      columns: [
        DriftColumn(
          sqlType: DriftSqlType.int,
          nullable: false,
          nameInSql: 'foo',
          nameInDart: 'foo',
          declaration: DriftDeclaration(fakeUri, -1, ''),
        ),
      ],
      baseDartName: 'database',
      nameOfRowClass: 'MyTableData',
    );

    SchemaVersionWriter(
      [
        SchemaVersion(
          1,
          [table],
          const {},
        ),
        SchemaVersion(
          2,
          [table],
          const {},
        ),
      ],
      writer.child(),
    ).write();

    expect(writer.writeGenerated(), contains('...'));
  });
}

```

drift_dev/test/writer/schema_version_writer_test.dart
```diff
@@ -0,0 +1,104 @@
...
+      expect(output, matches(containsTableRegex(tableName, withSuffix: true)));
+    }
+
+    // Tables that conflict with modified table names should themselves be
+    // modified to prevent the conflict. We can't check for nonexistence here
+    // because the the entire point is the name conficts with an in-use table
+    // name, so we only check for the existence of the doubly modified name.
+    for (final tableName in secondaryProblemTables.map((t) => t.baseDartName)) {
+      expect(output, matches(containsTableRegex(tableName, withSuffix: true)));
+    }
```
simolus3, line 100: Looks like this is already done in the first loop in line 93, should we just move the comment before that line and remove the loop?

drift_dev/test/writer/schema_version_writer_test.dart
```diff
@@ -0,0 +1,104 @@
...
+          ],
+          const {},
+        ),
+      ],
+      writer.child(),
+    ).write();
+
+    final output = writer.writeGenerated();
+
+    print(output);
```
simolus3: Let's remove this debug print.

simolus3 (commented):
The implementation looks good to me, I just have two small comments on the test.

drift_dev/test/writer/schema_version_writer_test.dart
jmatth: The two loops are verifying slightly different behavior. The first loop checks that for a reserved word like `database`, we define a getter named `databaseTable` and _not_ one named `database`. Earlier in the test I also create tables to conflict with the modified names, such as a table already named `databaseTable`. This second loop checks that we generate a getter named `databaseTableTable` to avoid the conflict that would otherwise occur from our earlier modification of `database`. However, we can't check for the nonexistence of the original name (`databaseTable`) like we did in the first loop because we know that it will exist; it's why there would be a conflict if we didn't append another `Table` suffix.

simolus3: Ah right! I was confused about the seemingly duplicate expectation, but I've overlooked the different iterable. Thanks for the explanation!

simolus3 (approved):
Thanks for the contribution!

PR #2621, opened 2023-09-19
example app - updated gradle wrapper to 7.5 to match filepicker plugin

crimesp:
Wasn't building on android due to a version conflict between existing 7.4.2 and the filepicker plugin 7.5 version

https://github.com/miguelpruivo/flutter_file_picker/blob/master/android/gradle/wrapper/gradle-wrapper.properties

builds tested: android,linux,chrome


simolus3 (approved):
Thanks for the fix!

PR #2625, opened 2023-09-20
[drift_dev] Add the static const `$name` to table

### LINKED ISSUE #2623: Add a static equivalent of the table getter `actualTableName`

In the project I'm working on. We record "updates"/"changes" of the local database in an outbox. In this outfox, there are 2 entries.
1. The name of the SQL table that was modified.
2. The modifications that were applied.

Let's say I have a table `Todos`:

```dart
class Todos extends Table {
  IntColumn get id => integer()();
  TextColumn get title => text()();
}
```

and let's say I have a method 

```dart
Future<void> recordChangesInOutbox(String tableName, Map<String, dynamic> changes);
```

When I want to record some changes I have to specify the table name `'todos'`. And I either have to hardcode it:

```dart
recordChangesInOutbox('todos', {});
```

Or I can use the generated `$TodosTable` by drift.
Unfortunately the getter `actualTableName` is not static so it forces me to instantiate the table:

```dart
recordChangesInOutbox($TodosTable(attachedDatabase).actualTableName, {});
```

It would be useful to also generate a static getter to get the actual table name, for example:

```dart
static String get actualTableNameStatic;
```

so I don't have to instantiate the table:

```dart
recordChangesInOutbox($TodosTable.actualTableNameStatic, {});
```

ValentinVignal:
Fixes https://github.com/simolus3/drift/issues/2623

### DIFF
--- a/docs/lib/snippets/modular/drift/example.drift.dart
+++ b/docs/lib/snippets/modular/drift/example.drift.dart
@@ -39,10 +39,10 @@ class Todos extends i0.Table with i0.TableInfo<Todos, i1.Todo> {
   @override
   List<i0.GeneratedColumn> get $columns => [id, title, content, category];
   @override
-  String get aliasedName => _alias ?? 'todos';
+  String get aliasedName => _alias ?? actualTableName;
   @override
-  String get actualTableName => 'todos';
-  static String get actualTableNameStatic => 'todos';
+  String get actualTableName => $name;
+  static const String $name = 'todos';
   @override
   i0.VerificationContext validateIntegrity(i0.Insertable<i1.Todo> instance,
       {bool isInserting = false}) {
@@ -281,10 +281,10 @@ class Categories extends i0.Table with i0.TableInfo<Categories, i1.Category> {
   @override
   List<i0.GeneratedColumn> get $columns => [id, description];
   @override
-  String get aliasedName => _alias ?? 'categories';
+  String get aliasedName => _alias ?? actualTableName;
   @override
-  String get actualTableName => 'categories';
-  static String get actualTableNameStatic => 'categories';
+  String get actualTableName => $name;
+  static const String $name = 'categories';
   @override
   i0.VerificationContext validateIntegrity(i0.Insertable<i1.Category> instance,
       {bool isInserting = false}) {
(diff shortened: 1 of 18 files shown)


ValentinVignal:
I didn't modify any test. I suppose some generated files are being used as tests. Tell me if you want me to write a specific test for that

docs/lib/snippets/modular/drift/example.drift.dart
```diff
@@ -42,6 +42,7 @@ class Todos extends i0.Table with i0.TableInfo<Todos, i1.Todo> {
   String get aliasedName => _alias ?? 'todos';
   @override
   String get actualTableName => 'todos';
+  static String get actualTableNameStatic => 'todos';
   @override
```
simolus3, line 46: I'm not sure if `actualTableNameStatic` is such a good name for this, but I also can't come up with something much better. Perhaps we can call it `$name`?

I think it would also be nice to have this as a constant field instead of a getter. To avoid repetition, I think we should also modify `actualTableName` and `aliasedName` to return `actualTableName` (or `$name`) instead of duplicating the string literal.

simolus3:
> I suppose some generated files are being used as tests

Pretty much, yeah. You could write a test that invokes the static field somewhere and asserts it has the expected value. That makes it easier to remember that the field is part of the public API and shouldn't be changed in future versions of the generator.

docs/lib/snippets/modular/drift/example.drift.dart
ValentinVignal: Yeah... I was also not happy with `actualTableNameStatic` 😅  ... 

I renamed it in `$name` in [revamp: Rename actualTableNameStatic into $name](https://github.com/simolus3/drift/pull/2625/commits/d001ad533f91b990e13f3fd10a4c948a26b4f221)

ValentinVignal:
I added a test in [test: Add a test using .$name](https://github.com/simolus3/drift/pull/2625/commits/aa2613c42dd716c20ddf47f59714fc7ad4d9b810). Is it what you were expecting?

simolus3 (approved):
Looks good, thanks!

PR #2663, opened 2023-10-03
CustomExpression.dialectSpecific when used would add in empty arguments when generating queries

Leylan24:
This is a fix when using the CustomExpression.dialectSpecific. If there is more than one dialect it would add no parameters when generating the queries. I just added to writeInto function where it will add the dialect specific content to the buffer.


simolus3 (approved):
Ouch, my bad! :facepalm: Thanks for the fix!

PR #2700, opened 2023-10-30
Edge cases for deriving the singular data class name from table names that are plural

bascoe10:
Complete todo with covering edge cases in deriving the singular data class name from a plural table name


simolus3 (commented):
Thanks :+1: The changes look good to me, but it would be nice to have tests for them. Tests like these could be expanded to verify the `table.nameOfRowClass` for these cases:

https://github.com/simolus3/drift/blob/ddc864a7c21c0b3845f8b1493ab64e121058c0b1/drift_dev/test/analysis/resolver/dart/table_test.dart#L118-L125

bascoe10:
I agree. I will add some test cases.

simolus3:
Let me know if you need any help with the tests cases :)

bascoe10:
> Let me know if you need any help with the tests cases :)

I got caught up with some other things. I have updated my PR with some test cases.

drift_dev/test/analysis/resolver/dart/table_test.dart
```diff
@@ -142,6 +158,42 @@ class InvalidConstraints extends Table {
...
+      final table = result!.result as DriftTable;
+
+      expect(table.nameOfRowClass, 'Piano');
+    });
+
+    test("singularizes table name that end in -s", () async {
+      final result = await findTable('Socks');
+      final table = result!.result as DriftTable;
+
+      expect(table.nameOfRowClass, 'Sock');
```
simolus3: Could you move this into the first `singularizes table name that end in -s` test so that the test verifies both cases (or just remove this since it seems to test the same thing if I'm not overlooking anything)? Having two tests with the same name might be confusing.

simolus3 (approved):
Thanks for the tests! Let's avoid the duplicate, but apart from that this looks good.

drift_dev/test/analysis/resolver/dart/table_test.dart
bascoe10: I must have gotten carried away with the test cases. I removed the duplicated test case.

Issue #2731, opened 2023-11-16
Add column-name-list syntax spec


tibotix:
**Is your feature request related to a problem? Please describe.**
Lately i've noticed the [column-name-list](https://sqlite.org/syntax/column-name-list.html) syntax is not supported by the drift sqlparser.
Namely, the following type of queries are not supported inside .drift files:

```sql
UPDATE foo SET (a,b) = (b,a);
UPDATE foo SET (a,b) = (SELECT a,b FROM bar WHERE foo.id=bar.id);
UPDATE foo SET (a,b) = (SELECT 3+3,4);
UPDATE foo SET (a,b)=(b,a), (c,d)=(d,c);
INSERT INTO demo VALUES (1,2) ON CONFLICT DO UPDATE SET (a,b)=(b,a);
```

This feature is describe in more detail on the [lang_update](https://sqlite.org/lang_update.html) page where it says:

> Beginning in SQLite [version 3.15.0](https://sqlite.org/releaselog/3_15_0.html) (2016-10-14), an assignment in the SET clause can be a [parenthesized list of column names](https://sqlite.org/syntax/column-name-list.html) on the left and a [row value](https://sqlite.org/rowvalue.html) of the same size on the right.

**Describe the solution you'd like**
I would like to have the sqlparser recognize SetComponents with a column-name-list. Ideally, also type inference with variables should work. Following test cases can be made for this:

```sql
CREATE TABLE foo (
   a INTEGER PRIMARY KEY,
   b TEXT NOT NULL
);

UPDATE foo SET (a,b) = (:v1,:v2);
-- v1 -> BasicType.int
-- v2 -> BasicType.text

UPDATE foo SET (a,b) = (SELECT :v1,:v2);
-- v1 -> BasicType.int
-- v2 -> BasicType.text
```

PR #2732, opened 2023-11-17
Add column-name-list syntax spec

tibotix:
This PR aims to solve issue #2731 .

Please read the commit messages / code changes for more details on the actual implementation.

A few notes:
This PR only targets the sqlparser.
All tests pass except the ones for type-resolving variables when using subqueries. Please feel free to suggest any ideas on how to implement this (see the TODO in `TypeResolver.visitMultiColumnSetComponent`). I am not as familiar with the code base as you, and there are still some things i don't quite understand, so there definitely may be better ways to implement some things. I tried to be as similar as possible to other code in the code base based on what i've read in the past few hours. 
Anyway, i would love to get some feedback and details on how to procceed with this :)

### DIFF
--- a/sqlparser/lib/src/analysis/types/resolving_visitor.dart
+++ b/sqlparser/lib/src/analysis/types/resolving_visitor.dart
@@ -131,10 +131,26 @@ class TypeResolver extends RecursiveVisitor<TypeExpectation, void> {
   }
 
   @override
-  void visitSetComponent(SetComponent e, TypeExpectation arg) {
+  void visitSingleColumnSetComponent(
+      SingleColumnSetComponent e, TypeExpectation arg) {
     visit(e.column, const NoTypeExpectation());
     _lazyCopy(e.expression, e.column);
     visit(e.expression, const NoTypeExpectation());
+  }
+
+  @override
+  void visitMultiColumnSetComponent(
+      MultiColumnSetComponent e, TypeExpectation arg) {
+    visitList(e.columns, const NoTypeExpectation());
+    if (e.rowValue is Tuple) {
+      final expressions = (e.rowValue as Tuple).expressions;
+      for (final (idx, expression) in expressions.indexed) {
+        _lazyCopy(expression, e.columns.elementAtOrNull(idx));
+      }
+    } else if (e.rowValue is SubQuery) {
+      // TODO: handle subquery case
+    }
+    visit(e.rowValue, const NoTypeExpectation());
   }
 
   @override
--- a/sqlparser/lib/src/analysis/types/resolving_visitor.dart
+++ b/sqlparser/lib/src/analysis/types/resolving_visitor.dart
@@ -142,15 +142,20 @@ class TypeResolver extends RecursiveVisitor<TypeExpectation, void> {
   void visitMultiColumnSetComponent(
       MultiColumnSetComponent e, TypeExpectation arg) {
     visitList(e.columns, const NoTypeExpectation());
-    if (e.rowValue is Tuple) {
-      final expressions = (e.rowValue as Tuple).expressions;
-      for (final (idx, expression) in expressions.indexed) {
-        _lazyCopy(expression, e.columns.elementAtOrNull(idx));
-      }
-    } else if (e.rowValue is SubQuery) {
-      // TODO: handle subquery case
+
+    final targets = e.resolvedTargetColumns ?? const [];
+    for (final column in targets) {
+      _handleColumn(column, e);
     }
-    visit(e.rowValue, const NoTypeExpectation());
+
+    final expectations = targets.map((r) {
+      if (r != null && session.graph.knowsType(r)) {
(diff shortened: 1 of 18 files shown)


sqlparser/lib/src/reader/parser.dart
```diff
@@ -1686,19 +1689,49 @@ class Parser {
     )..setSpan(withClause?.first ?? updateToken, _previous);
   }
 
+  SingleColumnSetComponent _singleColumnSetComponent() {
+    final columnName =
+        _consume(TokenType.identifier, 'Expected a column name to set')
+            as IdentifierToken;
```
simolus3: Let's use the opportunity and change this to `_consumeIdentifier` as well.

sqlparser/lib/utils/node_to_text.dart
```diff
@@ -1062,12 +1062,21 @@ class NodeSqlBuilder extends AstVisitor<void, void> {
...
   @override
-  void visitSetComponent(SetComponent e, void arg) {
+  void visitSingleColumnSetComponent(SingleColumnSetComponent e, void arg) {
     visit(e.column, arg);
     symbol('=', spaceBefore: true, spaceAfter: true);
     visit(e.expression, arg);
   }
 
+  @override
+  void visitMultiColumnSetComponent(MultiColumnSetComponent e, void arg) {
```
simolus3, line 1072: Let's also add a test for this in `test/utils/node_to_text_test.dart`

sqlparser/lib/src/analysis/types/resolving_visitor.dart
```diff
@@ -131,12 +131,28 @@ class TypeResolver extends RecursiveVisitor<TypeExpectation, void> {
...
+  void visitMultiColumnSetComponent(
+      MultiColumnSetComponent e, TypeExpectation arg) {
+    visitList(e.columns, const NoTypeExpectation());
+    if (e.rowValue is Tuple) {
+      final expressions = (e.rowValue as Tuple).expressions;
+      for (final (idx, expression) in expressions.indexed) {
+        _lazyCopy(expression, e.columns.elementAtOrNull(idx));
+      }
+    } else if (e.rowValue is SubQuery) {
+      // TODO: handle subquery case
```
simolus3: I think this can be handled similarly to what we're doing for inserts (see `visitInsertStatement`). Basically, visit the columns first which should give us their type since they're direct references to a schema we already know. Then, visit the RHS with a `SelectTypeExpectation` which is deconstructed in `visitTuple` and `visitSelectStatement`, so we can get away with a single visit and don't need to distinguish the two cases here at all.

sqlparser/test/parser/update_test.dart
```diff
@@ -26,6 +26,51 @@ void main() {
...
+                    ExpressionResultColumn(
+                      expression: Reference(columnName: 'a'),
+                    ),
+                  ],
+                      from: TableReference('bar', as: 'b'),
+                      where: BinaryExpression(
+                        Reference(entityName: 'b', columnName: 'id'),
+                        token(TokenType.equal),
+                        Reference(entityName: 'foo', columnName: 'id'),
+                      ))))
```
simolus3: Let's add a few commas between the parentheses here so that `dart format` makes this look a little more tree-like :)

simolus3 (commented):
This looks great, thanks for the contribution!

I've left a comment with what I think should resolve the type inference issue - let me know if that doesn't turn out to work.

sqlparser/lib/src/analysis/types/resolving_visitor.dart
tibotix: Thank you. I have mostly copied the code from `visitInsertStatement` now and all tests pass :).
It would be nevertheless helpful if you could look into the current implementation of `TypeResolver.visitMultiColumnSetComponent`. I am especially unsure about the calls to `_handleColumn` , as the tests pass with and without them.

tibotix:
Thanks for your quick reply. i have implemented your suggestions and the type inference turns out to be working now as well. Great :) Let me know if the changes fit your ideas.

sqlparser/lib/src/analysis/types/resolving_visitor.dart
simolus3: > I am especially unsure about the calls to `_handleColumn` , as the tests pass with and without it.

I think it doesn't make a difference here since we'll always have references to schema columns with a fixed type. For some other column references like those to a CTE or a compound select, calling `_handleColumn` is necessary to resolve their types. It also closes the gap between the AST model and the graph of column references.

So it doesn't hurt to have it.

simolus3 (approved):
This looks great, thanks again for the contribution!

sqlparser/lib/src/analysis/types/resolving_visitor.dart
tibotix: Alright, thanks for your explanation.

PR #2786, opened 2023-12-12
#2743: Ignore empty custom constraints + revert if statement for readability

### LINKED ISSUE #2743: Does `drift_dev schema dump` has no backward compatibility from 1.1.0?
I have some JSONs and dart generated files for migrations.

There was no issue before, but currently i'm update to the `drift 2.13.1` with `drift_dev 2.13.2`, update my table and re-run `bump` and `generate`.

And from this point my tests to this new database version is failed because of the schemas.
For example, like this:
```sh
Schema does not match
folder_table:
 columns:
  color:
   Not equal: `` (expected) and `NULL` (actual)
  deleted_datetime:
   Not equal: `` (expected) and `NULL` (actual)
  order_by_index:
   Not equal: `` (expected) and `NULL` (actual)
```

And, as i see, there is really big difference between JSON from this version and new one.

Maybe, there is an issue with some drift builder's options, but i don't use them.

Can it be fixed somehow from my side?

Vorkytaka:
What if we check custom constraints for null and empty first, and fill in defaults otherwise?
Looks like it can fix #2743 without breaking backwards compatibility, but i'm not sure. 😁

### DIFF
--- a/drift/lib/src/runtime/query_builder/schema/column_impl.dart
+++ b/drift/lib/src/runtime/query_builder/schema/column_impl.dart
@@ -126,7 +126,11 @@ class GeneratedColumn<T extends Object> extends Column<T> {
       into.buffer.write('$escapedName ${type.sqlTypeName(into)}');
     }
 
-    if ($customConstraints == null) {
+    if ($customConstraints != null && $customConstraints!.isNotEmpty) {
+      into.buffer
+        ..write(' ')
+        ..write($customConstraints);
+    } else {
       if (!isSerial) {
         into.buffer.write($nullable ? ' NULL' : ' NOT NULL');
       }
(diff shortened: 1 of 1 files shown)


simolus3:
I don't think these changes are sound.

First, they'd only help if you do a schema transition from a `NULL` constraint to something that doesn't have any column constraints. E.g. if you had `NULL REFERENCES ...` before and only `REFERENCES ...` now, this would still be broken.

I also don't think this is backwards-compatible, since you would now get the `NULL` constraint even if you just write `my_column TEXT,` (without a `NULL`) in a drift file - `$customConstraints` would be empty here as well.

-----

Column constraints in the database should match what you wrote in a drift file. That wasn't the case for `NULL` constraints, which has been fixed with 08ceb939bc2869c9c31b8f4c18828278f68d459d. Luckily it's only a harmless constraint which doesn't actually do anything, but it's still pretty bad. The mentioned commit properly fixes the issue though, so I prefer that solution.

PR #2811, opened 2023-12-29
Convert web encoding to utf8

Dall127:
Update web implementation to use UTF-8 for encoding, and updating unit tests

### DIFF
--- a/drift/lib/src/web/binary_string_conversion.dart
+++ b/drift/lib/src/web/binary_string_conversion.dart
@@ -9,13 +9,13 @@ class _BinaryStringConversion extends Codec<List<int>, String> {
   const _BinaryStringConversion();
 
   @override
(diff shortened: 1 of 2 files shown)


simolus3:
Could you explain the need for these changes a bit more? The intention here is to store a `Uint8List` as a string because that's the interface the underlyng web API is giving us. Obviously, the string shouldn't be much larger than the original byte array. So a byte value `x` should be represented by the char code `x` as well. This is exactly what `latin1` is doing. Note that the byte data we want to store is not always a string, this needs to work for all byte sequences, including those that aren't valid utf8.

Dall127:
That's a valid point; In full honesty, I just barely tested out the purpose on my live site and it didn't actually fix the issue, so I apologize for jumping the gun on making a merge request. I'll make an issue here in a moment describing the actual problem I'm trying to solve.

PR #2820, opened 2024-01-05
[bugfix/sum-and-total-aggregate]

tt88dev:
# Bugfix feature - String columns to support MAX and MIN

### We have the following table:

```dart
class example extends Table {
  TextColumn get string_type => text().withDefault(const Constant('N'))(); //N,Y
  RealColumn get quantity => real()();
}
```

#### Example Table:

| id  | name   | string_type | quantity |
| --- | ------ | ----------- | -------- |
| 1   | TREE01 | N           | 700.0    |
| 2   | TREE01 | N           | 800.0    |
| 3   | TREE01 | N           | 600.0    |
| 4   | TREE01 | Y           | 200.0    |

### We need to execute the following SQL:

```sql
SELECT
	"example"."name" AS "example.name",
	SUM("example"."string_type" IS 'N') AS "c1",
	SUM("example"."quantity") AS "c2"
FROM
	"example"
GROUP BY
	"example"."name"
```

#### Result:

| example.name | c1  | c2     |
| ------------ | --- | ------ |
| TREE01       | 3   | 2300.0 |

### The following is how to use it.

```dart
    final sumStringTypeEqN = example.string_type.equals('N').sum(); //<-- Fix this LINE
    final sumQuantity = example.quantity.sum();
    final queryExample = selectOnly(
      example,
    )
      ..addColumns([
        example.name,
        sumStringTypeEqN,
        sumQuantity,
      ])
      ..groupBy([
        example.name,
      ])
     ;
```

### DIFF
--- a/drift/lib/src/runtime/query_builder/expressions/aggregate.dart
+++ b/drift/lib/src/runtime/query_builder/expressions/aggregate.dart
@@ -44,6 +44,26 @@ extension BaseAggregate<DT extends Object> on Expression<DT> {
   Expression<DT> min({Expression<bool>? filter}) =>
       _AggregateExpression('MIN', [this], filter: filter);
 
+  /// Calculate the sum of all non-null values in the group.
+  ///
+  /// If all values are null, evaluates to null as well. If an overflow occurs
+  /// during calculation, sqlite will terminate the query with an "integer
+  /// overflow" exception.
+  ///
+  /// See also [total], which behaves similarly but returns a floating point
+  /// value and doesn't throw an overflow exception.
+  /// {@macro drift_aggregate_filter}
+  Expression<int> sum({Expression<bool>? filter}) =>
+      _AggregateExpression('SUM', [this], filter: filter);
+
+  /// Calculate the sum of all non-null values in the group.
+  ///
+  /// If all values in the group are null, [total] returns `0.0`. This function
+  /// uses floating-point values internally.
+  /// {@macro drift_aggregate_filter}
+  Expression<double> total({Expression<bool>? filter}) =>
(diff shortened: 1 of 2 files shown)


simolus3:
Thanks for the detailed descriptions and you contribution. However, I don't think we should have sum and total on all types (or perhaps I didn't understand the motivation behind these changes properly).

Is this for booleans? Instead of

> `example.string_type.equals('N').sum();`

I would probably use `countAll(filter: example.string_type.equals('N'))` instead.

I think it would be confusing to expose `sum` and `total` on all expression types, they only really work on numbers. If you insist on writing your queries with `sum`, you can use `example.string_type.equals('N').dartCast<int>().sum();`.

tt88dev:
> Is this for booleans? Instead of
> 
> >

Actually, I'm not entirely sure if this is intended for booleans or strings. For instance, in the first case:

```dart
final sumDiscrepancyOnlyNotZero = (example.stock - example.quantity).isNotValue(0).sum();
```

It seems like it might involve boolean logic since it's checking for non-zero values. However, in the second case:

```dart
final sumStringTypeEqN = example.string_type.equals('N').sum();
```

It appears to be related to strings, specifically checking for the equality with 'N'.

Could you clarify whether this functionality is meant for booleans or strings?

Upon further analysis, it seems to lean towards boolean operations.

However, by using the following modification:

```dart
final sumDiscrepancyOnlyNotZero = (example.stock - example.quantity).isNotValue(0).dartCast<int>().sum();
final sumStringTypeEqN = example.string_type.equals('N').dartCast<int>().sum();
```

I believe the current requirements are met.

This revision may not be essential.

Thank you for your assistance.

PR #2830, opened 2024-01-10
feat: add coverage ignore for generated files

Ragueel:
With these changes, generated files will always be ignored when running tests with coverage.


simolus3:
Thanks for the contribution! However, I don't think we should always opt generated files out of test coverage - maybe some people do want to know how much of their generated code is involved in tests. Making this optional with a builder option would be preferable IMO, but then there is the option from `source_gen` already so I'm not sure whether this is necessary.

neiljaywarner:
@simolus3  shoudl this be closed so new people to the project don't see it as an orange flag?
@Ragueel 
thanks

dickermoshe:
Seems like we don't plan on merging this, I'm going to close this for now...

PR #2831, opened 2024-01-11
docs: add one-liner dependency addition instructions in the documentation

utamori:
When a beginner wants to try this library, it is a bit troublesome because of the many dependencies
I think it would be helpful if you also included a one-liner on how to add it, so all you have to do is copy this and run it.

### DIFF
--- a/docs/pages/docs/setup.md
+++ b/docs/pages/docs/setup.md
@@ -45,6 +45,12 @@ dev_dependencies:
   build_runner: ^{{ versions.build_runner }}
 ```
 
+Alternatively, you can achieve the same result using the following command:
+
+```
+flutter pub add drift sqlite3_flutter_libs path_provider path dev:drift_dev dev:build_runner
+```
+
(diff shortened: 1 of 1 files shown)


docs/pages/docs/setup.md
```diff
@@ -45,6 +45,12 @@ dev_dependencies:
   build_runner: ^{{ versions.build_runner }}
 ```
 
+Alternatively, you can achieve the same result using the following command:
+
+```
+flutter pub add drift sqlite3_flutter_libs path_provider path dev:drift_dev dev:build_runner
```
simolus3: The `dart` command also works with Flutter, so let's use

```suggestion
dart pub add drift sqlite3_flutter_libs path_provider path dev:drift_dev dev:build_runner
```

simolus3 (commented):
Thanks!

simolus3 (approved):

PR #2837, opened 2024-01-16
Update pubspec.yaml

ghost:


simolus3 (approved):
These changes don't hurt, but since we're using the carret syntax we already support all the latest dependencies (and we should get them with `dart pub upgrade`). Could you explain why these changes are needed?

simolus3:
Looks like you're just posting these PRs to lots of repositories with no follow-up.

PR #2855, opened 2024-01-22
fix: Fix error message for non nullable converter

### LINKED ISSUE #2854: Wrong message when a non nullable converter is used on a nullable column:
```console
This column is nullable, but the type converter has a non-nullable SQL type, meaning that it won't be able to map `null` from the database to Dart.
Try wrapping the converter in `JsonTypeConverter.asNullable`
   ╷
49 │   IntColumn get myColumn => integer().map(const MyConverter()).nullable()();
   │         
```

But `JsonTypeConverter.asNullable` doesn't exist, it should be replaced with `JsonTypeConverter2.asNullable`

ValentinVignal:
Fixes https://github.com/simolus3/drift/issues/2854

### DIFF
--- a/drift_dev/lib/src/analysis/resolver/shared/dart_types.dart
+++ b/drift_dev/lib/src/analysis/resolver/shared/dart_types.dart
@@ -325,7 +325,7 @@ AppliedTypeConverter? readTypeConverter(
           "potentially map to `null` which can't be stored in the database.");
     } else if (!canBeSkippedForNulls) {
       final alternative = appliesToJsonToo
-          ? 'JsonTypeConverter.asNullable'
+          ? 'JsonTypeConverter2.asNullable'
           : 'NullAwareTypeConverter.wrap';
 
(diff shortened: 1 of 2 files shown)


simolus3 (approved):
Thanks!

ValentinVignal:
I quickly looked around and wasn't able to find if you have already written some tests for reported errors.

Where / how do you want me to write a test for that?

simolus3:
I think this assertion just needs to be expanded to cover the full message:

https://github.com/simolus3/drift/blob/261c151fbf87a580b07bc8565afbd9d7f16e9a07/drift_dev/test/analysis/resolver/dart/type_converter_test.dart#L85-L86

ValentinVignal:
Thank you @simolus3 , I implemented a test in [test: Update the tests](https://github.com/simolus3/drift/pull/2855/commits/b03805d3995ae56581dbcf5015c5013fe1928d01)

PR #2909, opened 2024-03-04
In variable helper, check for UserDefinedSqlType, not CustomSqlType

davidhole:
I am using ElectricSQL and the Electric Dart package. Their custom types for Postgres all implement `UserDefinedSqlType` and not `CustomSqlType`.

Please see: https://github.com/SkillDevs/electric_dart/blob/master/packages/electricsql/lib/src/client/conversions/custom_types.dart

This helper is therefore not being caught by this type check and the custom type is not being passed through to the variable, which means that the custom type is not being correctly mapped into the resulting SQL. This affects all queries as the query builders all use the `variable` helper

As the `Variable` constructor is expecting `UserDefinedSqlType`, it seems more correct that this is the type that it is checking against.

### DIFF
--- a/drift/lib/src/runtime/query_builder/expressions/custom.dart
+++ b/drift/lib/src/runtime/query_builder/expressions/custom.dart
@@ -24,14 +24,14 @@ class CustomExpression<D extends Object> extends Expression<D> {
   @override
   final Precedence precedence;
 
-  final CustomSqlType<D>? _customSqlType;
+  final UserDefinedSqlType<D>? _customSqlType;
 
   /// Constructs a custom expression by providing the raw sql [content].
   const CustomExpression(
     this.content, {
     this.watchedTables = const [],
     this.precedence = Precedence.unknown,
-    CustomSqlType<D>? customType,
+    UserDefinedSqlType<D>? customType,
   })  : _dialectSpecificContent = null,
         _customSqlType = customType;
 
@@ -41,7 +41,7 @@ class CustomExpression<D extends Object> extends Expression<D> {
     Map<SqlDialect, String> content, {
     this.watchedTables = const [],
     this.precedence = Precedence.unknown,
-    CustomSqlType<D>? customType,
+    UserDefinedSqlType<D>? customType,
   })  : _dialectSpecificContent = content,
         content = '',
         _customSqlType = customType;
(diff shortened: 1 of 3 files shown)


simolus3 (approved):
And to make matters worse I was the one migrating the Electric client to dialect-aware types... Thanks for the fix!

This looks good to me, but could you also add a test for this? I think something like this should work:

```patch
diff --git a/drift/lib/src/runtime/query_builder/expressions/custom.dart b/drift/lib/src/runtime/query_builder/expressions/custom.dart
index 90e1d526..0f7f6bdc 100644
--- a/drift/lib/src/runtime/query_builder/expressions/custom.dart
+++ b/drift/lib/src/runtime/query_builder/expressions/custom.dart
@@ -24,14 +24,14 @@ class CustomExpression<D extends Object> extends Expression<D> {
   @override
   final Precedence precedence;
 
-  final CustomSqlType<D>? _customSqlType;
+  final UserDefinedSqlType<D>? _customSqlType;
 
   /// Constructs a custom expression by providing the raw sql [content].
   const CustomExpression(
     this.content, {
     this.watchedTables = const [],
     this.precedence = Precedence.unknown,
-    CustomSqlType<D>? customType,
+    UserDefinedSqlType<D>? customType,
   })  : _dialectSpecificContent = null,
         _customSqlType = customType;
 
@@ -41,7 +41,7 @@ class CustomExpression<D extends Object> extends Expression<D> {
     Map<SqlDialect, String> content, {
     this.watchedTables = const [],
     this.precedence = Precedence.unknown,
-    CustomSqlType<D>? customType,
+    UserDefinedSqlType<D>? customType,
   })  : _dialectSpecificContent = content,
         content = '',
         _customSqlType = customType;
diff --git a/drift/test/database/expressions/custom_types_test.dart b/drift/test/database/expressions/custom_types_test.dart
index c63bc69a..29b1240a 100644
--- a/drift/test/database/expressions/custom_types_test.dart
+++ b/drift/test/database/expressions/custom_types_test.dart
@@ -44,6 +44,23 @@ void main() {
     expect(exp, generates('?', [10]));
     expect(exp.driftSqlType, isA<_NegatedIntType>());
   });
+
+  test('also supports dialect-aware types', () {
+    const a = CustomExpression(
+      'a',
+      customType: DialectAwareSqlType<int>.via(
+        fallback: _NegatedIntType(),
+        overrides: {SqlDialect.postgres: DriftSqlType.int},
+      ),
+      precedence: Precedence.primary,
+    );
+
+    expect(a.equals(3), generates('a = ?', -3));
+    expect(
+        a.equals(3),
+        generatesWithOptions('a = ?',
+            variables: [3], dialect: SqlDialect.postgres));
+  });
 }
 
 class _NegatedIntType implements CustomSqlType<int> {
```

davidhole:
Thanks @simolus3,

Added that test with a couple of tweaks, all looks good / passing :)

simolus3:
Thanks!

PR #2910, opened 2024-03-05
feat: Add option `use_sql_column_name_as_json_key`

### LINKED ISSUE #2847: DataClass toJson/fromJson with column names
Hello again Simon!
There is a feature request for drift we got from someone using `electricsql`. The user expected the `DataClass` `fromJson` to use the column names and not the dart field names as the map keys.

I see that there is an option in the generator to use the column names when using drift files (`use_column_name_as_json_key_when_defined_in_moor_file`), but not when using dart table definitions.
Is there any particular reason why is implemented that way and not for both? Looking at the name it looks like a very old option, back when the package was called `moor`. 

I've made the easy change locally in the following location to work on Dart tables too and it seems to work ok, no analysis errors or anything.

https://github.com/simolus3/drift/blob/261c151fbf87a580b07bc8565afbd9d7f16e9a07/drift_dev/lib/src/analysis/results/column.dart#L115

ValentinVignal:
Should close https://github.com/simolus3/drift/issues/2847

### DIFF
--- a/drift_dev/test/writer/data_class_writer_test.dart
+++ b/drift_dev/test/writer/data_class_writer_test.dart
@@ -267,6 +267,58 @@ extension ItemToInsertable on i1.Item {
       result.writer,
     );
   });
+
+  test(
+    'generates fromJson and toJson with the sql column names as json keys',
+    () async {
+      final writer = await emulateDriftBuild(
+        options: const BuilderOptions({
+          'use_sql_column_name_as_json_key': true,
+        }),
+        inputs: const {
+          'a|lib/main.dart': r'''
+import 'package:drift/drift.dart';
+
+part 'main.drift.dart';
+
+class MyTable extends Table {
+  TextColumn get myFirstColumn => text()();
+  IntColumn get mySecondColumn => integer()();
+}
+
+
+@DriftDatabase(
+  tables: [MyTable],
+)
+class Database extends _$Database {}
+'''
+        },
+      );
+
+      checkOutputs({
+        'a|lib/main.drift.dart': decodedMatches(contains(r'''
+  factory MyTableData.fromJson(Map<String, dynamic> json,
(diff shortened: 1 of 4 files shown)


ValentinVignal:
For the tests, I see you have an `extras/integration_tests/` folder. Should I duplicate one of its sub folder and set the option to `true` ?

simolus3:
Thanks! 

> Should I duplicate one of its sub folder and set the option to `true` ?

We have a simpler way of testing all stages of the generator in a unit test. You can take a look at [this test](https://github.com/simolus3/drift/blob/ca0c70eac92896cdca4a1623924658455ab586db/drift_dev/test/writer/data_class_writer_test.dart#L46-L93) as inspiration, `emulateDriftBuild` takes a `BuilderOptions` instance to which you can pass a map with `'use_sql_column_name_as_json_key': true`.

drift_dev/test/writer/data_class_writer_test.dart
```diff
@@ -267,6 +267,58 @@ extension ItemToInsertable on i1.Item {
...
+    return <String, dynamic>{
+      'my_first_column': serializer.toJson<String>(myFirstColumn),
+      'my_second_column': serializer.toJson<int>(mySecondColumn),
+    };
+  }
+
+''')),
+      }, writer.dartOutputs, writer.writer);
+    },
+    tags: 'analyzer',
```
ValentinVignal, line 320: I'm not sure about this tag, it was also in the test you suggested me to look at. Should I leave it?

simolus3: Yes, it gives the test a little longer to run because the analyzer is sometimes a bit inconsistent in its timing.

simolus3 (approved):
Looks good to me now, thanks!

PR #2913, opened 2024-03-06
feat: Allow to name companion classes

### LINKED ISSUE #2751: More naming options
Right now, it seems difficult for me to find a nice setup for the class names offered by drift.

**Example**
I name my Table class "AppleTable". 
Then, the generated Table will be called "AppleTableTable". 
The actual class I get out of this Table will be called "AppleTableData".
The companion will be called "AppleTableCompanion".

I dislike all of these names, although "AppleTableTable" name does not matter to me, since its not exported anywhere in my code.

I can slightly change this outcome with the `@DataClassName('Apple')`.
With that I then get `Apple` for my model. However, the other two classes stay the same.
This still bothers me, because I think "AppleTableCompanion" is not a great name.

**Another Example**
A different approach is to name my Table "Apples". 
This gives me "ApplesTable" (okay), "Apple" (great) and "ApplesCompanion" (ehh). 
This appears one of the best options (and kind of "cannonical" by drift), though I think I would prefer "AppleCompanion" over "ApplesCompanion".

I think a nicer default behaviour would be great.
Another incredibly simple solution would be to allow directly naming the Companion class in an annotation.
In my other code I oftentimes write a class called "\<Model>Request" which is actually quite similar to a Companion.insert, meaning it is a model that simply lacks the ID. Maybe I would just start naming my Companions that way.

I must also mention that it is possible for me to simply bypass this problem with a bunch of typedefs that map the generated names to my desired names. But this is of course very inconvenient.

ValentinVignal:
Closes https://github.com/simolus3/drift/issues/2751


drift/lib/src/dsl/table.dart
```diff
@@ -308,7 +308,11 @@ final class TableIndex {
 class DataClassName {
   /// The overridden name to use when generating the data class for a table.
   /// {@macro drift_custom_data_class}
-  final String name;
+  final String? name;
+
+  /// The overridden name to use when generating the companion class for a table.
+  /// {@macro drift_custom_data_class}
```
simolus3: I think this hint doesn't really apply to companions as much, let's remove the doc macro application here.

drift_dev/lib/src/analysis/resolver/dart/helper.dart
```diff
@@ -198,11 +198,13 @@ extension TypeUtils on DartType {
 
 class DataClassInformation {
   final String enforcedName;
```
simolus3: The name is not really enforced for `DataClassName.custom` with a null `name`, right? Can we make this field nullable as well and move the ` ?? dataClassNameForClassName(element.name);` to the places currently reading the field? (I haven't checked if that causes a lot of duplication, if it makes code ugly let's keep it as is)

drift_dev/test/writer/data_class_writer_test.dart
```diff
@@ -267,6 +267,56 @@ extension ItemToInsertable on i1.Item {
       result.writer,
     );
   });
+
+  test(
```
simolus3: Thanks for the test! Can we also have one with a null `name` in `DataClassName.custom` to check that it falls back to the default?

simolus3 (commented):
Nice, thank you!
Can you also start a `2.17.0-dev` changelog entry for `drift` and mention this there?

drift_dev/test/writer/data_class_writer_test.dart
ValentinVignal: I added a test in [test: Add a test with a custom companion name and a default data name](https://github.com/simolus3/drift/pull/2913/commits/6b1f54d6473d529080b589fabacd8f46407732fe)

drift_dev/lib/src/analysis/resolver/dart/helper.dart
ValentinVignal: It looks fine, I didn't have to change it in a lot of placed. You can checked it in [refactor: Make enforcedName nullable](https://github.com/simolus3/drift/pull/2913/commits/683ff8ad61fbc54138d4651941a75f92b1671004)

drift/lib/src/dsl/table.dart
ValentinVignal: Sure ! Removed in [docs: Remove unrelated macro](https://github.com/simolus3/drift/pull/2913/commits/27488cf3019dc08b373a3bac79d78c7a20b79acc)

simolus3 (approved):
Awesome, thank you!

PR #2943, opened 2024-04-06
Support bytea sql literal in postgres

davidmartos96:
Currently, when using a bytea literal, the generated Postgres SQL is invalid.

In Postgres we can use a similar syntax to pass the value as HEX.


simolus3 (approved):
Thanks!

davidmartos96:
@simolus3 I've just noticed a small issue in the pushed code.

 We can use the `hexString` variable from above instead of reencoding again in `return "x'${hex.encode(dart)}'"`; for SQLite

simolus3:
Clearly I'm too tired to review PRs right now.. Thanks for pointing that out, fixed in 74fb2696159904e91ad8a98335b1bb9362363b2b.

Issue #2892, opened 2024-02-14
Implement Geopoly Interface


nikitadol:
**Is your feature request related to a problem? Please describe.**
It would be great to be able to have a `geopoly` module which is described here -> https://www.sqlite.org/geopoly.html

**Describe the solution you'd like**
There are 2 options:
- Add only this module
- Allow defining custom modules (as a more universal solution)

simolus3:
I agree that this is something worth having, but it's not a priority for me at the moment if there's no strong demand for it. Are you interested in contributing parts for this if I give you some pointers and help out a bit?

> Allow defining custom modules (as a more universal solution)

You can add support for custom functions in the query builder by writing your own APIs on top of things like `FunctionCallExpression` - that works for geopoly as well.
To make use of these functions in drift files, you can register [additional functions](https://drift.simonbinder.eu/docs/advanced-features/builder_options/#known-custom-functions) when configuring the builder.

But it's harder to generalize support for custom `CREATE VIRTUAL TABLE`  modules, since they have no general syntax and we need to be able to figure out the actual schema statically.

nikitadol:
Hi
I'm sorry for the delay. It's not something urgent. I decided to try to do it myself, but ran into some problems:

1. Such expressions are not parsed correctly:
```sql
blob_in_string: select typeof(X'0100000300000000000000000000803F000000000000003F0000803F');
```
Result:
```dart
  Selectable<String> blob_in_string() {
    return customSelect(
        'SELECT typeof(\'\'\'0100000300000000000000000000803F000000000000003F0000803F\') AS _c0',
        variables: [],
        readsFrom: {}).map((QueryRow row) => row.read<String>('_c0'));
  }
```

2. This code does not completely create the `Table` class and therefore in `FunctionHandler` I cannot get information that the table is virtual
https://github.com/simolus3/drift/blob/37f120d287057c2cd2ad4b79f02a7584957b9f2e/drift_dev/lib/src/analysis/resolver/drift/sqlparser/mapping.dart#L52-L57

3. About `FunctionHandler`: I need to implement `inferReturnType` so that the correct code is generated - this is ok, but there is `inferArgumentType` in which there is no element index ([workaround](https://github.com/simolus3/drift/blob/37f120d287057c2cd2ad4b79f02a7584957b9f2e/sqlparser/lib/src/engine/module/fts5.dart#L143-L148)) and it is also unclear what the result of this function affects

[...]

PR #2939, opened 2024-04-02
Add Manager

dickermoshe:
### Warning, The following proposed API has changed from the examples shown here, see the docs for an updated representation of the manager APi.

- [x] Docs
- [x] Tests (Lot's of em')

____

This introduces a simpler interface to access the database.  
It makes it easier to create simple, or even some complex quiries and then directly read, watch, update or delete based on 
them. It should also lower the barrier of entry for developers used to simpler database alternatives. 

### Managers
By default every generated database will have a new field `managers`, which exposes a manager for each table in the database.  
Each manager is able to:

* Create new instances: `managers.categories.create((o) => o(name: 'test'))`
* Create new instances in bulk: `managers.categories.bulkCreate((o) => [o(name: 'test1'),o(name: 'test2')])`
* Replace existing instaces: `managers.categories.replace(category.copyWith(name:"Important"))`

#### Filtering & Ordering

However, the real superpower is the packaged filter and ordering builder.
Here is an example schema to show how these filters work.

```dart
@DataClassName('TodoEntry')
class TodoEntries extends Table {
  IntColumn get id => integer().autoIncrement()();
  TextColumn get description => text()();
  @ReferenceName("todos") // Optional: Custom name for the reverse reference 
  IntColumn get category => integer().nullable().references(Categories, #id)();
  DateTimeColumn get dueDate => dateTime().nullable()();
}

@DataClassName('Category')
class Categories extends Table {
  IntColumn get id => integer().autoIncrement()();
  TextColumn get name => text()();
  IntColumn get color => integer().map(const ColorConverter())();
}

class ColorConverter extends TypeConverter<Color, int> {
  const ColorConverter();
  @override
  Color fromSql(int fromDb) => Color(fromDb);
  @override
  int toSql(Color value) => value.value;
}
```

1. Filter on all columns
```dart
// All TodoEntries before today
todoEntries.filter((f) => f.dueDate.isBefore(DateTime.now()))

// Todo Entry that matches id=5
todoEntries.filter((f) => f.id(5))
OR
todoEntries.filter((f) => f.id.equals(5))

// Filter on Custom Types
categories.filter((f) => f.color(Colors.white))

// Combine multiple filters using `&` and `|`
categories.filter((f) => f.color(Colors.white) & f.name("School"))
```

2. Filter on references
```dart
// All Todo Entries whose categories are white
todoEntries.filter((f) => f.category((f) => f.color(Colors.white))

// Combine with local filters
// Todo Entries that are already due or whose categories are white
[...]

PR #2946, opened 2024-04-10
Add support for geopoly Interface 

nikitadol:
Solves https://github.com/simolus3/drift/issues/2892

Questions and suggestions regarding implementation are welcome

More tests and perhaps an example will be added soon


drift_dev/lib/src/analysis/results/types.dart
```diff
@@ -36,7 +36,7 @@ abstract class HasType {
 /// appears where an array is expected or has a type converter applied to it.
 /// [HasType] is the interface for sql-typed elements and is implemented by
 /// columns.
-class ColumnType {
+sealed class ColumnType {
```
simolus3, line 39: I like the distinction between custom and builtin types here :+1: We should check whether we can avoid the special subclass for geopoly types though.

drift/lib/src/runtime/types/mapping.dart
```diff
@@ -81,6 +81,15 @@ final class SqlTypes {
       return dartValue.rawSqlValue;
     }
 
+    if (dartValue is GeopolyPolygon) {
```
simolus3: There are a bunch of cases where I didn't do a good job at that, but I'm trying to move the core `package:drift/drift.dart` library to be more agnostic of the underlying DBMS. So supporting geopoly (a sqlite3-only module) here makes it more complicated to make drift support more databases in the future.

I fully support the analysis additions around geopoly types (because these are always opt-in), but I wonder if we should instead implement the runtime support for polygons in a separate library. It can still be part of the drift package, but perhaps under a different import (say `package:drift/extensions/geopoly.dart`, next to things like json1 support). We can then implement geopoly support with [custom types](https://drift.simonbinder.eu/docs/sql-api/types/) (albeit a custom type implemented in the drift package), which also makes it easier to define tables using polygon columns in Dart.
The generator should still apply that custom type by default if return values from the geopoly extension are involved, but that way we can let the core runtime focus on types supported by all SQL implementations.

simolus3 (commented):
Awesome work with the sqlparser extension! I think we should consider moving the runtime mapping away from the core library to reflect that this is not something all databases support. The rest of this look great!

nikitadol:
Regarding all your comments:
I tried doing this via `CustomSqlType`, which I put in `package:drift/extensions/geopoly.dart`, but there are a few problems:
- The only place where I can specify a different type of function columns and arguments is here:
https://github.com/simolus3/drift/blob/78387a361061c49592b2109e42ba3a7fe33c83ff/drift_dev/lib/src/analysis/resolver/drift/sqlparser/mapping.dart#L153-L155
- It is difficult to create a `DartType` (in this exact place) and then pass it to the `CustomColumnType`

I'm still thinking about a solution, I will be glad if you have any suggestions, since `sqlparser` cannot depend on `drift_dev`

Perhaps a mechanism is needed to allow modules to define types

simolus3:
> The only place where I can specify a different type of function columns and arguments is here:

I'm using the `KnownDriftTypes` utility to solve this. `package:drift/src/drift_dev_helper.dart` exports a bunch of classes which we'd like to resolve in `drift_dev`, we can then pick them up from there and turn them into `InterfaceType`s. That library could also export the geopoly types.

> I'm still thinking about a solution, I will be glad if you have any suggestions, since `sqlparser` cannot depend on `drift_dev`

But we only need to inject the actual custom type in `drift_dev`, right? Once the type hint is there it should be preserved by `sqlparser`.

nikitadol:
I had to slightly change the use of `KnownDriftTypes` so as not to change many synchronous methods to asynchronous

Or is this not an option?

simolus3:
> I had to slightly change the use of `KnownDriftTypes` so as not to change many synchronous methods to asynchronous

I think this shouldn't be a problem. We need the known types for almost everything involving the driver, so we might as well preload it.

The failing test worries me a bit as loading more assets may degrade build performance, but since this is all coming from external packages that would invalidate the entire build either way, it shouldn't be a problem.
You can apply this patch to fix the test

```patch
diff --git a/drift_dev/test/backends/build/build_integration_test.dart b/drift_dev/test/backends/build/build_integration_test.dart
index d81f397f..e1828132 100644
--- a/drift_dev/test/backends/build/build_integration_test.dart
+++ b/drift_dev/test/backends/build/build_integration_test.dart
@@ -583,6 +583,8 @@ class MyDatabase {
       return everyElement(
         anyOf(
           isA<AssetId>().having((e) => e.extension, 'extension', '.json'),
+          // Allow reading SDK or other package assets to set up the analyzer.
+          isA<AssetId>().having((e) => e.package, 'package', isNot('a')),
           other,
         ),
       );
```

docs/pages/docs/SQL API/extensions.md
```diff
@@ -78,3 +78,35 @@ The `bm25`, `highlight` and `snippet` functions from fts5 can also be used in cu
...
+to describe two-dimensional polygons.
+Geopoly includes functions for detecting when one polygon is contained within or overlaps with another, 
+for computing the area enclosed by a polygon, 
+for doing linear transformations of polygons, 
+for rendering polygons as [SVG](https://en.wikipedia.org/wiki/SVG), 
+and other similar operations.
+
+To enable the fts5 extension in drift files and compiled queries, modify the
+[build options]({{ "../Generation options/index.md" | pageUrl }}) to include
+`geopoly` in the `sqlite_module` section.
```
simolus3: Nice :+1: Let's also briefly mention this section here: https://github.com/simolus3/drift/blob/ca0dee4f83234b053039d23b62d9defd007e7686/docs/pages/docs/Generation%20options/index.md?plain=1#L174

simolus3 (approved):
I've added a simple test to `drift` to make sure the type is roughly doing what it's supposed to. This looks good to me now, but I want to merge #2939 first.

nikitadol:
Thanks for helping
I also fixed some tests and added clarification to the documentation, since by default `geopoly` is not in the `sqlite3_flutter_libs` package

simolus3:
LGTM, thanks! :+1:

PR #2951, opened 2024-04-12
docs: minimise the copy/think/paste needed to get the minimal impleme…

ewann:
…ntation working

nice project :)

stepping through this for the first time, I decided to offer this back for consideration.

### DIFF
--- a/docs/pages/docs/setup.md
+++ b/docs/pages/docs/setup.md
@@ -100,7 +100,7 @@ is responsible for telling drift how to open the database. The `schemaVersion` g
 for migrations after changing the database, we can leave it at `1` for now. Update `database.dart`
 so it now looks like this:
 
-<a name="after_generation">
+<a name="open"></a>
 {% include "blocks/snippet" snippets = snippets name = 'after_generation' %}
 
 The Android-specific workarounds are necessary because sqlite3 attempts to use `/tmp` to store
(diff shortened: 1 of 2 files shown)


docs/pages/docs/setup.md
```diff
@@ -97,10 +97,11 @@ After running either command, the `database.g.dart` file containing the generate
...
 You will now see errors related to missing overrides and a missing constructor. The constructor
 is responsible for telling drift how to open the database. The `schemaVersion` getter is relevant
-for migrations after changing the database, we can leave it at `1` for now. The database class
-now looks like this:
-<a name="open">
-{% include "blocks/snippet" snippets = snippets name = 'open' %}
+for migrations after changing the database, we can leave it at `1` for now. Update `database.dart`
+so it now looks like this:
+
+<a name="after_generation">
```
simolus3: This link seems to cause Firefox to parse the HTML in a weird way, it breaks the snippet below for me. I guess we need to add a closing `</a>` immediately afterwards, I don't know how this has ever worked...

Still, the anchor should be named `open` because there are other parts that might link to that. So let's keep make it

```suggestion
<a name="open"></a>
```

simolus3 (commented):
Thanks! I agree that seeing the full snippet again can make it easier for users to just copy something that works.

docs/pages/docs/setup.md
ewann: suggested change applied

simolus3 (approved):
Thanks!

PR #2972, opened 2024-04-26
Improve Manager API design

dickermoshe:
This PR includes the following changes to the manager filter API.

### Remove `Value` Filter for Custom Type
Until now the manage would create 2 filters for a column that has a TypeConverter.
One that accepts the Custom Type and another that uses the raw value.
To avoid polluting the namespace of filterset, the ability to filter on a raw value will be removed.
It is extremely uncommon that a filter on this raw data would be performed, so I think it's worth it.
E.G 
```dart
f.category.status(Status.active)
f.category.statusValue(7) // Removed
```

### References don't use a callback
This will help with readability:
```dart
```
instead of 
```dart
```

### Remove `Id` filter for references
Until now, references columns would create 2 columns, one for the actual column, and one for the reference object.
We will remove the "Id" filter.
Before:
```dart
items.filter((f)=>f.categoryId(5))
```
After:
```dart
items.filter((f)=>f.category.id(5))
```
This looks like we are performing a join for no reason, but we aren't!
Under the hood no join is performed, a filter is applied to the actual category column!

### Reverse References filters wrapped in aggregate class
In the future we may want to add aggregate support for reverse filters.
This prepares for that
E.G
```dart
// Before
f.todoItemsRefs((f) => ...)
// After
f.todoItemsRefs.exist((f) => ...).any()
```

### Count Aggregate
Filter on count of a reverse referenced
```dart
f.todoItemsRefs.count().equals(5)
f.todoItemsRefs.count((f)=>f.name("People")).equals(5)
```

## Other Changes
1. Docs for extensions have been removed until the internal manager API is more consistent. We don't want people writing code at this lower level.
2. Mark lower level classes as internal, but still expose them for the generated code. This will change once the lower level apis are more stable.
3. Disable manager generation on older dart 2 code
4. Add `all` filter which doesn't do any filtering. Will become useful later for creating aggregates that don't use a filter (having)


dickermoshe:
This should be merged before any releases, it contains breaking changes.

I may have done a checkout from the wrong branch, I'll rebase when you finish #2970

simolus3 (approved):
I agree that the API changes make things more consistent, thanks :+1:

dickermoshe:
@simolus3 
What do you reccomend I do?
I would like to keep certain things as "internal" but use them in the generated code.
Can I add `// ignore_for_file: type=lint, invalid_use_of_internal_member` to all generated code?

dickermoshe:
Are you aware of a way to mark things as internal, but still be able to use them in the generated code without lint errors?

dickermoshe:
I've added some things to the PR description:
Take a look.

Otherwise, this PR is done.

dickermoshe:
I came across a bug for this PR while adding aggregate support.
I fixed it and push a fix, but it lowed my confidence in this PR.
I would feel much more confident if tests were added to cover all of these changes.
I think more tests for the manager itself should also be added also.

Marking this as draft until that is complete.

I should be done by Sunday Night

dickermoshe:
Closing in favor of https://github.com/simolus3/drift/pull/2977 which is much more organized.

PR #2977, opened 2024-04-28
Manager v2

dickermoshe:
This PR includes the following changes to the manager filter API.

### Remove `Value` Filter for Custom Type
Until now the manage would create 2 filters for a column that has a TypeConverter.
One that accepts the Custom Type and another that uses the raw value.
To avoid polluting the namespace of filterset, the ability to filter on a raw value will be removed.
It is extremely uncommon that a filter on this raw data would be performed, so I think it's worth it.
E.G 
```dart
f.category.status(Status.active)
f.category.statusValue(7) // Removed
```

### References don't use a callback
This will help with readability:
```dart
```
instead of 
```dart
```

### Remove `Id` filter for references
Until now, references columns would create 2 columns, one for the actual column, and one for the reference object.
We will remove the "Id" filter.
Before:
```dart
items.filter((f)=>f.categoryId(5))
```
After:
```dart
items.filter((f)=>f.category.id(5))
```
This looks like we are performing a join for no reason, but we aren't!
Under the hood no join is performed, a filter is applied to the actual category column!

## Other Changes
1. Docs for extensions have been removed until the internal manager API is more consistent. We don't want people writing code at this lower level.
2. Mark lower level classes as internal, but still expose them for the generated code. This will change once the lower level APIs are more stable.
3. Disable manager generation on older dart 2 code

### DIFF
--- a/docs/lib/snippets/dart_api/manager.dart
+++ b/docs/lib/snippets/dart_api/manager.dart
@@ -1,3 +1,5 @@
+// ignore_for_file: invalid_use_of_internal_member
+
 import 'package:drift/drift.dart';
 
 part 'manager.g.dart';
@@ -209,62 +211,61 @@ extension ManagerExamples on AppDatabase {
 // #enddocregion manager_filter_custom_back_references
 }
 
-// This documentation should added once the internal manager APIs are more stable
-// // #docregion manager_filter_extensions
-// // Extend drifts built-in filters by combining the existing filters to create a new one
-// // or by creating a new filter from scratch
-// extension After2000Filter on ColumnFilters<DateTime> {
-//   // Create a new filter by combining existing filters
-//   ComposableFilter after2000orBefore1900() =>
-//       isAfter(DateTime(2000)) | isBefore(DateTime(1900));
+// #docregion manager_filter_extensions
+// Extend drifts built-in filters by combining the existing filters to create a new one
+// or by creating a new filter from scratch
+extension After2000Filter on ColumnFilters<DateTime> {
+  // Create a new filter by combining existing filters
+  ComposableFilter after2000orBefore1900() =>
+      isAfter(DateTime(2000)) | isBefore(DateTime(1900));
 
-//   // Create a new filter from scratch using the `column` property
-//   ComposableFilter filterOnUnixEpoch(int value) =>
-//       $composableFilter(column.unixepoch.equals(value));
-// }
+  // Create a new filter from scratch using the `column` property
+  ComposableFilter filterOnUnixEpoch(int value) =>
+      $composableFilter(column.unixepoch.equals(value));
+}
 
-// Future filterWithExtension(AppDatabase db) async {
-//   // Use the custom filters on any column that is of type DateTime
-//   db.managers.todoItems.filter((f) => f.createdAt.after2000orBefore1900());
+Future filterWithExtension(AppDatabase db) async {
+  // Use the custom filters on any column that is of type DateTime
+  db.managers.todoItems.filter((f) => f.createdAt.after2000orBefore1900());
 
-//   // Use the custom filter on the `unixepoch` column
-//   db.managers.todoItems.filter((f) => f.createdAt.filterOnUnixEpoch(0));
-// }
-// // #enddocregion manager_filter_extensions
+  // Use the custom filter on the `unixepoch` column
+  db.managers.todoItems.filter((f) => f.createdAt.filterOnUnixEpoch(0));
+}
+// #enddocregion manager_filter_extensions
 
-// // #docregion manager_ordering_extensions
-// // Extend drifts built-in orderings by create a new ordering from scratch
-// extension After2000Ordering on ColumnOrderings<DateTime> {
-//   ComposableOrdering byUnixEpoch() => ColumnOrderings(column.unixepoch).asc();
-// }
+// #docregion manager_ordering_extensions
+// Extend drifts built-in orderings by create a new ordering from scratch
+extension After2000Ordering on ColumnOrderings<DateTime> {
+  ComposableOrdering byUnixEpoch() => ColumnOrderings(column.unixepoch).asc();
+}
 
-// Future orderingWithExtension(AppDatabase db) async {
-//   // Use the custom orderings on any column that is of type DateTime
-//   db.managers.todoItems.orderBy((f) => f.createdAt.byUnixEpoch());
-// }
-// // #enddocregion manager_ordering_extensions
+Future orderingWithExtension(AppDatabase db) async {
+  // Use the custom orderings on any column that is of type DateTime
+  db.managers.todoItems.orderBy((f) => f.createdAt.byUnixEpoch());
+}
+// #enddocregion manager_ordering_extensions
 
-// // #docregion manager_custom_filter
-// // Extend the generated table filter composer to add a custom filter
-// extension NoContentOrBefore2000FilterX on $$TodoItemsTableFilterComposer {
-//   ComposableFilter noContentOrBefore2000() =>
-//       (content.isNull() | createdAt.isBefore(DateTime(2000)));
-// }
+// #docregion manager_custom_filter
+// Extend the generated table filter composer to add a custom filter
+extension NoContentOrBefore2000FilterX on $$TodoItemsTableFilterComposer {
+  ComposableFilter noContentOrBefore2000() =>
+      (content.isNull() | createdAt.isBefore(DateTime(2000)));
+}
 
-// Future customFilter(AppDatabase db) async {
-//   // Use the custom filter on the `TodoItems` table
-//   db.managers.todoItems.filter((f) => f.noContentOrBefore2000());
-// }
-// // #enddocregion manager_custom_filter
+Future customFilter(AppDatabase db) async {
+  // Use the custom filter on the `TodoItems` table
+  db.managers.todoItems.filter((f) => f.noContentOrBefore2000());
+}
+// #enddocregion manager_custom_filter
 
-// // #docregion manager_custom_ordering
-// // Extend the generated table filter composer to add a custom filter
-// extension ContentThenCreationDataX on $$TodoItemsTableOrderingComposer {
-//   ComposableOrdering contentThenCreatedAt() => content.asc() & createdAt.asc();
-// }
+// #docregion manager_custom_ordering
+// Extend the generated table filter composer to add a custom filter
+extension ContentThenCreationDataX on $$TodoItemsTableOrderingComposer {
+  ComposableOrdering contentThenCreatedAt() => content.asc() & createdAt.asc();
+}
 
-// Future customOrdering(AppDatabase db) async {
-//   // Use the custom ordering on the `TodoItems` table
-//   db.managers.todoItems.orderBy((f) => f.contentThenCreatedAt());
-// }
-// // #enddocregion manager_custom_ordering
+Future customOrdering(AppDatabase db) async {
+  // Use the custom ordering on the `TodoItems` table
+  db.managers.todoItems.orderBy((f) => f.contentThenCreatedAt());
+}
+// #enddocregion manager_custom_ordering
--- a/docs/lib/snippets/dart_api/manager.dart
+++ b/docs/lib/snippets/dart_api/manager.dart
@@ -171,14 +171,13 @@ extension ManagerExamples on AppDatabase {
 // #docregion manager_filter_forward_references
   Future relationalFilter() async {
     // Get all items with a category description of "School"
-    managers.todoItems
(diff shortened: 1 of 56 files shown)


dickermoshe:
@simolus3 Adding tests and came across a bug.
The following SQL is completely valid:
```sql
SELECT
  DISTINCT COUNT(*) AS "c0"
FROM
  "book_club"
  LEFT OUTER JOIN "person" "book_club__id__person__club" ON "book_club"."id" = "book_club__id__person__club"."club"
GROUP BY
  "book_club"."_rowid_"
HAVING
  COUNT(*) = ?;
```

However, I'm getting the completly wrong output from row.read:

```dart
/// Count the number of rows that would be returned by the built statement
  Future<int> count() async {
    final count = _tableAsTableInfo.rowId.count();
    final result =
        _buildSelectStatement(targetColumns: [count]) as _JoinedResult;
    print(result.statement.constructQuery().sql); // Added for debugging
    print(result.statement.constructQuery().boundVariables);
    return await result.statement
        .map((row) => row.read(count)!)
        .get()
        .then((value) => value.firstOrNull ?? 0);
  }
```
The above is returning whatever `?` is.
So if:
```
HAVING
  COUNT(*) = 54;
```
Then the above function will return 54. There aren't even 54 clubs!
If seems that `row.read(count)!` is reading the wrong count?!

Just want to ensure that I'm not doing anything wrong before opening a issue on drift

dickermoshe:
@simolus3 I tried debugging this myself, but I'm at a loss.

There are Book Clubs and Members.
I'm filtering book clubs with 3 members.
This SQL query returns 1 club, which is CORRECT!
```sql
SELECT
  "book_club"."id" AS "book_club.id",
  "book_club"."name" AS "book_club.name"
FROM
  "book_club"
  LEFT OUTER JOIN "person" "book_club__id__person__club" ON "book_club"."id" = "book_club__id__person__club"."club"
GROUP BY
  "book_club"."_rowid_"
HAVING
  COUNT(*) = ?;
```

HOWEVER, COUNTING DOESN'T!
```sql
SELECT
  DISTINCT COUNT(*) AS "c0"
FROM
  "book_club"
  LEFT OUTER JOIN "person" "book_club__id__person__club" ON "book_club"."id" = "book_club__id__person__club"."club"
GROUP BY
  "book_club"."_rowid_"
HAVING
  COUNT(*) = ?;
```

This will always return whatever the number filter was, in this case, it's returning 3! There are only 2 clubs in the database!
Even accessing the raw value, gets `{c0: 3}`
This happens even without DISTINCT

This is either a really bad bug in drift or I'm doing something wrong.

This is the test that's failing, It's mind boggling.

https://github.com/dickermoshe/drift/blob/21b7607a77bfeb50b8cb56119043a636d10fa74d/drift/test/manager/manager_referenced_filter_test.dart#L161-L173

dickermoshe:
This is how the django ORM does counts.
Can this be created using drift?
```sql
SELECT
  COUNT(*)
FROM
  (
    SELECT
      "api_product"."sku" AS "col1"
    FROM
      "api_product"
      LEFT OUTER JOIN "api_listing" ON ("api_product"."sku" = "api_listing"."product_id")
    GROUP BY
      1
    HAVING
      COUNT("api_listing"."id") = 3
  )
```

dickermoshe:
@simolus3 
Figured it out!
As usual your awesome documentation made this easy!

dickermoshe:
@simolus3 
Counting can only really be done if there is a rowid or only 1 primary key.
Even ORMs that have been developed for the past 20 years don't support a database with more than 1 primary key.
We can either throw an error if they try to `count` (it's like that now) or ignore such tables when generating manager (makes more sense, less shoot-foot-gun issues).

What should be done?

simolus3:
To support Postgres and MariaDB, where we don't have a `rowid` column, I think we should strive to support compound primary keys as well.

>  Counting can only really be done if there is a rowid or only 1 primary key.

To be sure I'm understanding this correctly, the problem is that we care about the amount of unique rows of one table in a select statement that may have additional joins?
That's not really obvious from the documentation of that method, which just claims to return the amount of rows that would be returned by the whole statement. If that's the number we actually care about, what's wrong with `SELECT count(*) FROM (the actual statement that would otherwise be generated)`?

If we need "amount of unique rows from table x appearing in a joined result set", couldn't we add all the primary key columns to the inner result set and then do a sum + count combination? E.g. if the primary key is the columns `foo`, `bar`, `baz`:

```
SELECT sum(*) FROM (SELECT count(*) FROM (inner query) GROUP BY foo, bar, baz)
```

Then the inner query gives us how many rows there are for each unique combination of primary key columns and the final `sum` gives us the total amount. But I also don't see how that information is relevant in most queries.

dickermoshe:
That's real smart.

On a side note, I was looking at other ORM designs and have decided that aggregates should be done in another method and filtering should focus on filtering.

```dart
// Class where aggregates are predefined, these will be added to the filterset
class CategoryAggregateModel extends ...{
  ...
  final itemCount = ...;
  ...
}

category.annotate(CategoryAggregateModel()).filter((f)=>f.itemCount(5))
```

This will reduce code duplication if we ever decide to add a way to return the aggregated values.

I'm going to revert back to this design for filtering reverse references
```dart
// Category with todo with id of 5
category.filter(f.todos((f)=>f.id(5))) // Remove .exist() or .any()
```

Aggregates will be for a later release, they add complexity to the point where my limited knowledge of SQL is starting to show.

dickermoshe:
I've followed your guidance on counting and have begun adding more and more tests, It's starting to come together nicely.

simolus3:
Nice! Let me know if there's anything ready for me to take a more in-depth look at :)

dickermoshe:
@simolus3 
The refactoring here is quite signifying.
LMK if there is anything I can do to make your reviewing on this PR easier.

This PR is complete

docs/lib/snippets/dart_api/manager.dart
```diff
@@ -210,61 +209,62 @@ extension ManagerExamples on AppDatabase {
...
-extension ContentThenCreationDataX on $$TodoItemsTableOrderingComposer {
-  ComposableOrdering contentThenCreatedAt() => content.asc() & createdAt.asc();
-}
-
-Future customOrdering(AppDatabase db) async {
-  // Use the custom ordering on the `TodoItems` table
-  db.managers.todoItems.orderBy((f) => f.contentThenCreatedAt());
-}
-// #enddocregion manager_custom_ordering
+// This documentation should added once the internal manager APIs are more stable
```
simolus3: IMO we can leave them uncommented in this snippet file, as long as there's no snippet block on the markdown page referencing them they're not visible to users.

drift/lib/drift.dart
```diff
@@ -20,8 +20,8 @@ export 'src/runtime/query_builder/query_builder.dart'
     hide CaseWhenExpressionWithBase, BaseCaseWhenExpression;
 export 'src/runtime/types/converters.dart';
 export 'src/runtime/types/mapping.dart' hide BaseSqlType, UserDefinedSqlType;
-export 'src/runtime/manager/manager.dart'
-    hide JoinBuilder, HasJoinBuilders, Composer, BaseTableManager;
+// ignore: invalid_export_of_internal_element
```
simolus3: If this is necessary, we should either make the affected elements non-`@internal` or indeed hide them.

There are a bunch of classes in drift that are only meant to be used by generated code. I don't have `@internal` annotations on them but instead mention this in their documentation comment.
Adding `@internal` on constructors or methods not used outside of `package:drift` would also work to hide them specifically.

drift/lib/src/runtime/manager/composable.dart
```diff
@@ -50,29 +55,26 @@ class JoinBuilder {
...
-/// An interface for classes that hold join builders
-/// Typically used by classes whose composition requires joins
-/// to be created
+/// An interface for classes which need to hold the information needed to create
+/// orderings or where expressions.
 ///
 /// Example:
 /// ```dart
-/// categories.filter((f) => f.todos((f) => f.dueDate.isBefore(DateTime.now())))
+/// todos.filter((f) => f.category )
```
simolus3: Can this be completed into a working filter (e.g. `(f) => f.category.id(3)`) so that the example makes a little more sense on its own?

drift/lib/src/runtime/manager/filter.dart
```diff
@@ -118,201 +194,150 @@ extension StringFilters<T extends String> on ColumnFilters<String> {
...
-    }
-    return mappedValue;
-  }
-
-  /// Create a filter that checks if the column equals a value.
-  ComposableFilter equals(CustomTypeNonNullable value) {
-    return ComposableFilter(column.equals(_customTypeToSql(value)),
-        inverted: inverted);
-  }
+enum _Opperator { and, or }
```
simolus3: nit, `_Operator` (or perhaps `_BooleanOperator`)

drift_dev/lib/src/backends/build/drift_builder.dart
```diff
@@ -440,7 +440,8 @@ class _DriftBuildRun {
       }
     }
 
-    output.writeln('// ignore_for_file: type=lint');
+    output.writeln(
+        '// ignore_for_file: type=lint, invalid_use_of_internal_member');
```
simolus3: It's a clever solution to the "members only supposed to be used by generated code" problem, but I think I still prefer the approach of mentioning this in their dartdoc without relying on this lint.

drift_dev/lib/src/cli/commands/schema/generate_utils.dart
```diff
@@ -167,5 +167,5 @@ class GenerateUtilsCommand extends Command {
 
   static final _dartfmt = DartFormatter();
   static const _prefix = '// GENERATED CODE, DO NOT EDIT BY HAND.\n'
-      '// ignore_for_file: type=lint';
+      '// ignore_for_file: type=lint, invalid_use_of_internal_member';
```
simolus3: Same here

drift_dev/lib/src/writer/manager_writer.dart
```diff
@@ -88,32 +94,59 @@ class _ReferencedFilterWriter extends _FilterWriter {
   /// E.G `CategoryFilterComposer`
   final String referencedFilterComposer;
 
+  /// Whether this is a reverse reference
+  /// This is used to determine of the filter should
+  /// use a callback or return the related filterset directly
```
simolus3: Let's be more specific here and mention that reverse references get a callback - an explanation for that would also be nice, I guess it's for improved readability in the forward direction whereas we want to be more explicit for backrefs?

extras/integration_tests/drift_testcases/lib/tests.dart
```diff
@@ -1,6 +1,25 @@
...
+        ProcessedTableManager,
+        FilterComposer,
+        OrderingBuilder,
+        Composer,
+        BaseTableManager,
+        ColumnOrderings,
+        ColumnFilters,
+        ColumnWithTypeConverterFilters,
+        TableManagerState,
+        RootTableManager;
```
simolus3: And without the internal marker, re-exporting drift is easier than this :)

drift/lib/src/runtime/manager/composable.dart
```diff
@@ -42,6 +42,11 @@ class JoinBuilder {
     return currentColumn.hashCode ^ referencedColumn.hashCode;
   }
 
+  @override
```
simolus3, line 48: I've missed this in the previous PR. But while we're at it, can we change `hashCode` to use `Object.hash` instead of xors? I think `==` should also not be covariant and instead do a `other is JoinBuilder` check - otherwise it throws instead of returning false which is pretty weird.

simolus3 (commented):
Thanks! The API changes and refactorings look good to me (but I've yet to try them out). I just have some minor comments on the implementation.

drift_dev/lib/src/backends/build/drift_builder.dart
dickermoshe: > but I think I still prefer the approach of mentioning this in their dartdoc without relying on this lint.

What do you mean by this?

I'm confused on how to proceed.

I would like it if changes to this low level API weren't classified as breaking changes.
When `@internal` is removed, test fail due to the `melos analyze` test

I've taken inspiration from Riverpod which uses this lint ignore on their generated code, however if you don't like it that perfectly fine.

Moving the `@internal` to the constructors does solve the ugly `export` -> `hide` issues, but the lint issues are still present.

How should I proceed?

dickermoshe:
Let me know if there is anything else I can do...

simolus3 (approved):
Looks good to me, thanks again for all your hard work on this!

Is this ready to be released or do you have something in mind that should still be added or changed before a public version?

dickermoshe:
Yeah, this is ready for release.

PR #2983, opened 2024-05-02
Fix parameter binding when NULL is provided.

### LINKED ISSUE #2981: [drift_postgres] Cannot insert null value to default column
I'm running into an issue when using the `drift_postgres` package that doesn't occurr when using the `postgres` package.

For the following table:

```sql
CREATE TABLE IF NOT EXISTS public.parent (
  id INTEGER PRIMARY KEY NOT NULL,
  value TEXT,
  other INTEGER DEFAULT 0
);
```

This works using postgres:
```dart
    await db.execute('''
INSERT INTO public.parent (id, value, other) VALUES (3, 'hi', \$1);
''',
      parameters: [null],
    );
```

This doesn't work using drift
```dart
await db.customInsert('''
INSERT INTO public.parent (id, value, other) VALUES (3, 'hi', \$1);
''', variables: [Variable(null)]);
```

The error is 
`Severity.error 42804: column "other" is of type integer but expression is of type text hint: You will need to rewrite or cast the expression.`

davidmartos96:
Fixes #2981 

It turns out we cannot use `Type.text`, because postgres will complain when the type doesn't match, even if NULL is provided.

### DIFF
--- a/extras/drift_postgres/CHANGELOG.md
+++ b/extras/drift_postgres/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.2.2
+
+- Fix parameter binding when NULL is provided.
+
 ## 1.2.1
 
 - Fix `isOpen` so that it properly returns false after the underlying postgres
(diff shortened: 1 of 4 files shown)


simolus3 (approved):
Awesome, thanks for the report and fix!

PR #2984, opened 2024-05-02
Get Refs POC

dickermoshe:
This PR may either be developed to completion or discarded if it doesn't turn out great

@simolus3
Check this out.

Modular generation aren't working but that is ok for now.
The manager generator needs a refactoring either way

### DIFF
--- a/docs/lib/snippets/modular/drift/example.drift.dart
+++ b/docs/lib/snippets/modular/drift/example.drift.dart
@@ -335,6 +335,9 @@ class $TodosProcessedTableManager extends i0.ProcessedTableManager<
     $TodosInsertCompanionBuilder,
     $TodosUpdateCompanionBuilder> {
   const $TodosProcessedTableManager(super.$state);
+  $TodosReferenceReader withReferences() {
+    return $TodosReferenceReader(this);
+  }
 }
 
 typedef $TodosInsertCompanionBuilder = i1.TodosCompanion Function({
@@ -392,6 +395,29 @@ class $TodosTableManager extends i0.RootTableManager<
                   content: content,
                   category: category,
                 )));
+  $TodosReferenceReader withReferences() {
+    return $TodosReferenceReader(this);
+  }
+}
+
+class $TodosReferenceReader<T0>
+    extends i0.ReferenceReader<i1.Todo, ({i1.Todo todo, T0? category})> {
+  $TodosReferenceReader(this.$manager);
+  i0.GeneratedDatabase get _db => $manager.$state.db as i0.GeneratedDatabase;
+  final i0.BaseTableManager $manager;
+  @override
+  Future<({i1.Todo todo, T0? category})> $withReferences(i1.Todo value) async {
+    return (todo: value, category: await _getCategory(value));
+  }
+
+  Future<T0?> _getCategory(i1.Todo value) async {
+    return $getSingleReferenced<i1.Category>(value.category, _db.categories.id)
+        as T0?;
+  }
+
+  $TodosReferenceReader<i1.Category> withCategory() {
+    return $TodosReferenceReader(this.$manager);
+  }
 }
 
 class Categories extends i0.Table with i0.TableInfo<Categories, i1.Category> {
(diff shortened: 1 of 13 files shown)


docs/lib/snippets/modular/drift/example.drift.dart
```diff
@@ -392,6 +395,29 @@ class $TodosTableManager extends i0.RootTableManager<
...
+}
+
+class $TodosReferenceReader<T0>
+    extends i0.ReferenceReader<i1.Todo, ({i1.Todo todo, T0? category})> {
+  $TodosReferenceReader(this.$manager);
+  i0.GeneratedDatabase get _db => $manager.$state.db as i0.GeneratedDatabase;
+  final i0.BaseTableManager $manager;
+  @override
+  Future<({i1.Todo todo, T0? category})> $withReferences(i1.Todo value) async {
+    return (todo: value, category: await _getCategory(value));
```
simolus3, line 410: So this would always resolve all references, but then set the ones we don't care about to `void`? I think we should also keep track of which tables we care about to avoid unnecessary load.

docs/lib/snippets/modular/many_to_many/relational.drift.dart
```diff
@@ -597,4 +600,48 @@ class $$ShoppingCartEntriesTableTableManager extends i0.RootTableManager<
...
+          T0? shoppingCart,
+          T1? item
+        })> {
+  $$ShoppingCartEntriesTableReferenceReader(this.$manager);
+  i0.GeneratedDatabase get _db => $manager.$state.db as i0.GeneratedDatabase;
+  final i0.BaseTableManager $manager;
+  @override
+  Future<({i2.ShoppingCartEntry shoppingCartEntry, T0? shoppingCart, T1? item})>
+      $withReferences(i2.ShoppingCartEntry value) async {
+    return (
```
simolus3, line 622: Does this effectively run two additional queries for each row in the main one? I think we should check whether we can transform the query to be a join so that we get the correct rows right away.

As it stands, this has a potential for race conditions (a query should always return a consistent snapshot between all tables involved). We could solve that by wrapping the whole thing in a transaction, but it's still quite inefficient and I think we should try to use a single statement with joins.

simolus3 (commented):
Thank you for all the work on this! I think the concept looks great from the API perspective, but we should really look at possible performance improvements and avoid the [N+1 problem](https://stackoverflow.com/questions/97197/what-is-the-n1-selects-problem-in-orm-object-relational-mapping).

docs/lib/snippets/modular/drift/example.drift.dart
dickermoshe: Nope, it doesn't actually even query the reference when the generic `T0` is `void`.  

By default all the generics `('T0...T10')` are `void` by default.
Using `with___` sets the associated generic to the actual value.
When $withReferences is eventually called for each object, it uses the generic to decide whether anything should even be queried.

The below code does that check:

https://github.com/dickermoshe/drift/blob/2a8a8f4b41d3fa0f006d17a67cf44d788fe574d6/drift/lib/src/runtime/manager/reference_reader.dart#L24-L33

docs/lib/snippets/modular/many_to_many/relational.drift.dart
dickermoshe: Yeah, this part was done kind of quickly, so it's poorly optimized.  
The solution would be to either do a join, or doing a single 2nd query  `categories.filter((f)=>f.id.isIn([ ... ]))` to return a list of categories, using 1 transaction to avoid race conditions.

I'm still researching ways to have nested work, although I don't think there is way that would be both type-safe & easy to use.

dickermoshe:
I've managed to have this working with joins on another branch. opening another PR in it's place

docs/lib/snippets/modular/drift/example.drift.dart
simolus3: Oh I see! Very clever

PR #2987, opened 2024-05-03
[drift_postgres] Let values go through as is, so that Postgres uses native conversions

### LINKED ISSUE #2986: [drift_postgres] Postgres implicit type casts not working in raw queries
Similar to #2981 this explicit casting can break valid Postgres queries.

For instance, passing a String to insert a INT8 column is valid in the postgres driver, but when forcing it to be always `Type.text` it will complain. 

Should we let the values go through to the `postgres` package with unspecified typing?

davidmartos96:
Fixes #2986

### DIFF
--- a/extras/drift_postgres/lib/src/pg_database.dart
+++ b/extras/drift_postgres/lib/src/pg_database.dart
@@ -168,11 +168,13 @@ class _BoundArguments {
         final value = args[i];
         return switch (value) {
           TypedValue() => value,
-          null => TypedValue(Type.unspecified, null),
-          int() || BigInt() => TypedValue(Type.bigInteger, value),
-          String() => TypedValue(Type.text, value),
-          bool() => TypedValue(Type.boolean, value),
-          double() => TypedValue(Type.double, value),
+          null ||
+          int() ||
+          String() ||
+          bool() ||
+          double() =>
+            TypedValue(Type.unspecified, value),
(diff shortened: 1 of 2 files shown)


davidmartos96:
Just to give you more context, we detected this in electric.
The server sends values that should be inserted as is in the Postgres driver. That means that for bigints for example, it will send a String, and since we are doing the insert through drift_postgres, the explicit type mapping fails.

davidmartos96:
I guess an alternative would be to pass TypedValue with unspecified from inside Electric. 
Let me know if we should close or you want this behavior.

While working on the PR I noticed that BigInts where not working correctly in the develop branch. This PR fixes it. In case of discarding this PR, we might want to fix the BigInt values binding, as the `postgres` package doesn't understand them.

simolus3:
>  Let me know if we should close or you want this behavior.

Given that a workaround exists by passing a `TypedValue`, I'm leaning towards keeping the existing behavior. For `null` values the text type was just broken and definitely wrong, but I think passing a `Variable<String>` from drift should result in something interpreted as a string by the driver. So postgres with drift essentially behaves like sqlite3 would behave if all tables were `STRICT`. I don't think there's any part in drift that relies on weak types, and that sounds like a correct assumption to make for a sound SQL wrapper in Dart.

I've merged your bigint-related changes as 6ba049e5df7f2a3aff50c02ffdd326f596f53452 - thanks!

PR #2988, opened 2024-05-03
Refactor Manager Generator

dickermoshe:
@simolus3 
Now that the manager API is more stable, I've modified the generator to be cleaner and easier to read

This doesn't have any material impact on the generated code.


drift_dev/lib/src/writer/manager/manager_templates.dart
```diff
@@ -0,0 +1,387 @@
...
+/// A class which contains utility functions to generate manager class names
+///
+/// This is used by the [DatabaseManagerWriter] to generate code for the manager classes
+class _ManagerCodeTemplates {
+  _ManagerCodeTemplates(this.__scope);
+
+  /// A Scope class which contains the current scope of the generation
+  ///
+  /// Used to generating names which require import prefixes
+  final Scope __scope;
```
simolus3: One underscore is enough :)

drift_dev/lib/src/writer/manager/manager_templates.dart
```diff
@@ -0,0 +1,387 @@
...
+    return switch (__scope.generationOptions.isModular) {
+      true => leaf.drift("GeneratedDatabase"),
+      false => dbClassName,
+    };
+  }
+
+  /// The code for the database manager getter which will be added to the main database class
+  ///
+  /// E.g. `AppDatabase get managers => AppDatabaseManager(this);`
+  String databaseManagerGetter(String dbClassName) {
```
simolus3: I haven't checked, but this sounds like something we'll only ever use at a single place, right (when generating the main database class)? In that case I think it's easier to define it in the class using it rather than moving it out into a second file (I agree about this refactoring for stuff called in different places though).

This applies to some of the other definitions below writing getters as well - IMO it's easier to have that all in one class if it's only used by one class. The methods returning names/references should be here, but if things aren't used multiple times I think separating them just makes code harder to follow.

drift_dev/lib/src/writer/manager/table_manager_writer.dart
```diff
@@ -0,0 +1,212 @@
...
+
+  final List<DriftTable> otherTables;
+
+  _TableManagerWriter(
+      {required this.table,
+      required this.scope,
+      required this.dbClassName,
+      required this.otherTables})
+      : assert(table.existingRowClass == null,
+            "Manager Writer should ignore tables with custom row classes");
```
simolus3, line 26: I think it's fine for now, but we should investigate supporting custom row classes as well (#2989). I'm happy to take a look at that, do you recall what didn't work with custom classes?

drift_dev/lib/src/writer/manager/manager_templates.dart
```diff
@@ -0,0 +1,387 @@
...
+  String orderingComposerName(DriftTable table) {
+    return '\$${table.entityInfoName}OrderingComposer';
+  }
+
+  /// Name of the typedef for the insert companion builder for a table
+  ///
+  /// This is the name of the typedef of a function that creates new rows in the table
+  String insertCompanionBuilderTypeDef(DriftTable table) {
+    return '\$${table.entityInfoName}InsertCompanionBuilder';
+  }
```
simolus3, line 107: Are these typedefs possibly referenced for other tables? In that case we might have to add a `.generatedElement` variant as well

drift_dev/lib/src/writer/manager/manager_templates.dart
```diff
@@ -0,0 +1,387 @@
...
+    DriftTable table,
+  ) {
+    return '\$${table.entityInfoName}FilterComposer';
+  }
+
+  /// Name of this tables ordering composer class
+  String orderingComposerNameWithPrefix(DriftTable table, TextEmitter leaf) {
+    return leaf
+        .dartCode(leaf.generatedElement(table, orderingComposerName(table)));
+  }
```
simolus3, line 95: If possible, can we only use the `withPrefix` variants and remove the text-based versions? That makes things a bit simpler since it's impossible to forget to use `withPrefix` in modular mode. `generatedElement` will not emit a prefix when not necessary so we can use it for the monolithic builder as well.

simolus3 (commented):
I agree with the typedefs and moving common templates into a separate file. Not so common templates should stay where they are though ;)

drift_dev/lib/src/writer/manager/manager_templates.dart
dickermoshe: I'll move the smaller code templates that are only used once.  

However, the larger code templates, even though they are only used once, pollute the logic of the actual generation  step. I think those should remain in the templates class.

drift_dev/lib/src/writer/manager/manager_templates.dart
dickermoshe: These are only used by their own table managers, no prefix is required

drift_dev/lib/src/writer/manager/manager_templates.dart
dickermoshe: No, even in modular mode you need the name without the prefix when creating the actual class.

```dart
class i0.ClassName { // Invalid
...
}
```

What you're describing happened to me, but any bug like that would throw a static error, which tests find instantaneously

drift_dev/lib/src/writer/manager/manager_templates.dart
dickermoshe: Done

drift_dev/lib/src/writer/manager/manager_templates.dart
```diff
@@ -0,0 +1,371 @@
...
+    return """
+          ${leaf.drift("ColumnWithTypeConverterFilters")}<$converterType,$nonNullableConverterType,$type> get $filterName => \$state.composableBuilder(
+      column: \$state.table.$columnGetter,
+      builder: (column, joinBuilders) => 
+      ${leaf.drift("ColumnWithTypeConverterFilters")}(column, joinBuilders: joinBuilders));
+      """;
+  }
+
+  /// Code for a filter which works over a reference
+  String relatedFitler(
```
simolus3: Missed this typo before, `relatedFilter`

simolus3 (approved):
LGTM - just noticed one typo but apart from that this is good to go

drift_dev/lib/src/writer/manager/manager_templates.dart
dickermoshe: Done

dickermoshe:
Thanks

PR #2994, opened 2024-05-05
add `then` to selectable

dickermoshe:
@simolus3 
This modifies your code

This helper methods add the ability for query A to depend on query B.
This what I would like to use to add prefetching to the manager api.

```dart
$state
.withJoins(joinBuilders)
.managerState
.buildJoinedStatement()
.map((result) {
   // Add what we could get via joins (Regular References)
   final singleReferencedItem = result.readTableOrNull(joinBuilders.referencedTable)
  ...

}).then((results) {
   // This gives us all the items at one, so we can do 1 query to get all the items at once (Reverse References)
 final allIds = results.map((i)=>i.id);
final allrelated = await filter((f)=> f.todos.id.isIn(allIds));
// Now we just need to append the referenced items to a list and return
});
```

### DIFF
--- a/drift/lib/src/runtime/query_builder/statements/query.dart
+++ b/drift/lib/src/runtime/query_builder/statements/query.dart
@@ -1,3 +1,4 @@
+// ignore_for_file: public_member_api_docs, sort_constructors_first
 part of '../query_builder.dart';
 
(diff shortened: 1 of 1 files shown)


dickermoshe:
This may not be needed, investigating steams...

PR #3002, opened 2024-05-09
docs: indicate minimum drift version for manager functionality

ewann:
I believe something like this could be helpful.

Also:
> drift 2.17 introduced a new set of APIs designed to make common queries much easier to write

It could be good to link/clarify the docs regarding that, though I suspect it is basically referencing the remainder of the docs site? Still, it potentially leaves a new consumer unsure if they are referencing the correct info.


simolus3:
Thanks! I have fixed the mistaken version number in aa5314df67d0f8261f82700a0fe235d5e87119a9 - I hope that clarifies things enough. I'm not sure if we need to put the minimum version on the top of the page that prominently, I think it's somewhat expected that the docs always reference the latest version by default.

PR #3004, opened 2024-05-11
Support custom row classes for manager

### LINKED ISSUE #2989: Manager: Support tables with custom row classes
At the moment, no manager classes are generated for tables with custom row classes.

We should investigate whether it's possible to make these APIs available for existing row classes as well. It's always a good principle to have different features work well together, and I don't see why users of custom row classes shouldn't also want the manager APIs for simple queries.
If there are parts of the manager API that are harder to support with existing row classes, perhaps we can start with a subset first and gradually expand it.

simolus3:
This adds support for generating manager classes for tables with existing row classes. At the moment, the only thing I had to change was removing the `DataClass` generic bound in the runtime.
@dickermoshe, can you check whether you need the `extends DataClass` bound for future manager changes? If not I think this sounds like a fairly easy way to port manager code over to more tables.

Closes https://github.com/simolus3/drift/issues/2989.

### DIFF
--- a/drift/lib/src/runtime/manager/manager.dart
+++ b/drift/lib/src/runtime/manager/manager.dart
@@ -9,18 +9,16 @@ part 'filter.dart';
 part 'composable.dart';
 part 'ordering.dart';
 
-sealed class _StatementType<T extends Table, DT extends DataClass> {
+sealed class _StatementType<T extends Table, DT> {
   const _StatementType();
 }
 
-class _SimpleResult<T extends Table, DT extends DataClass>
-    extends _StatementType<T, DT> {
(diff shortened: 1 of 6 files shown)


dickermoshe:
This may make it harder to add features later, but nothing impossible.
Custom Row Classes are needed for supporting other DBs so I guess it's a worthy tradeoff.

dickermoshe (approved):

dickermoshe (changes requested):
Seems like tests are failing

dickermoshe (approved):
Some small assert was causing bugs, LGTM now

simolus3:
Thanks for taking a look and fixing the assert!

PR #3013, opened 2024-05-19
Fix: toColumns mixin generated with wrong name

0biWanKenobi:
In `data_class_writer.dart` we expect a mixin named `${table.entityInfoName}ToColumns`, but then we create one named `${table.baseDartName}ToColumns` in `table_writer.dart`.


simolus3 (commented):
Thanks! To avoid this problem in the future, could you move the name generation into a common getter?

E.g. `String get toColumnsMixin => '${entityInfoName}ToColumns'` in `DriftElementWithResultSet`. Both places could then refer to that getter instead of constructing the name themselves.

0biWanKenobi:
Sounds like a good idea! I'll do it asap 🫡

simolus3 (approved):
Thanks 👍

PR #3019, opened 2024-05-23
Add Name Clash Info To Docs

dickermoshe:


docs/pages/docs/Dart API/manager.md
```diff
@@ -60,6 +60,12 @@ We can now use them in a query like this:
 In this example, had we not specified a custom name for the reference, the code generator would have named both filtersets `userRefs` for both references to the `User` table. This would have caused a conflict. By specifying a custom name, we can avoid this issue.
 
 
+### Name Clashes
```
simolus3: Imo this could be a `####` heading, then it shows neatly under the code snippet using the annotation.

simolus3: Giving this an anchor also gives us a opportunity to reference drift.simonbinder.eu/docs/manager/#name-clashes in the error message.

simolus3 (commented):
That was quick, thanks!!

I think we should also reference the updated docs in the error message to perhaps provide more context.

docs/pages/docs/Dart API/manager.md
dickermoshe: Done :)

simolus3 (approved):
Looks good to me, thanks!

Issue #3015, opened 2024-05-21
Database can no longer be mocked with new manager API


frankvollebregt:
**Describe the bug**
After upgrading drift (currently on 2.18), the `build_runner` outputs the following error when it tries to generate a mocked version of my database:
```
Invalid @GenerateMocks annotation: Mockito cannot generate a valid mock class which implements 'MyDatabase' for the following reasons:
    The property accessor '_$MyDatabase.managers' features a private return type, and cannot be stubbed. Try generating this mock with a MockSpec with 'unsupportedMembers' or a dummy generator (see https://pub.dev/documentation/mockito/latest/annotations/MockSpec-class.html).
```
I tried to follow the advice in the message and add a `MockSpec`, but to no avail for now. Is it necessary/expected that I need to add the custom mock spec, or is this an unintended result of recent changes in drift/drift_dev?

My annotation (I added the `customMocks` after first getting this error):
```dart
@generateMocks(
  [MyDatabase],
  customMocks: [
    MockSpec<MyDatabase>(unsupportedMembers: {#_$MyDatabase.managers}),
  ]
)
```

kuhnroyal:
Had the same problem, but got it working somehow, don't remember if I disabled manager generation.
Try to use `@GenerateNiceMocks` and `#managers` as symbol name. Alternatively disable manager generation in `build.yaml` if you don't use them.

In general this should be changed in drift. The `managers` instance could be some interface type.

frankvollebregt:
Thank you! It now works with
```dart
@GenerateNiceMocks([
  MockSpec<MyDatabase>(unsupportedMembers: {#managers}),
])
```
I'll leave the issue open for now, since this only circumvents the issue, and I indeed suspect this isn't intended behavior

zjcz:
As a workaround to this, the release notes for [2.18.0](https://github.com/simolus3/drift/releases/tag/drift-2.18.0) mention you can turn this feature off in build.yaml - see https://drift.simonbinder.eu/docs/advanced-features/builder_options/ for more info.

Basically, create a file called build.yaml in the root of your Flutter project and add the following.  When you next run build_runner, it won't generate the managers. 

```yaml
targets:
  $default:
    builders:
      drift_dev:
        options:
          generate_manager: false
```

dickermoshe:
@frankvollebregt 
This is my fault. I'll work on this.  

I haven't had much experience with the Mockito library.
What sorta fix do you suppose would solve this issue?

[...]

PR #3025, opened 2024-05-24
Add Flutter Tests - Revert Analyzer Bump - Make Managers Public - Test Mockito

### LINKED ISSUE #3024: Flutter not resolving
```console
mick@MD-LAPTOP-1:~/drift/examples/app$ flutter pub get
Resolving dependencies... 
Note: meta is pinned to version 1.12.0 by flutter_test from the flutter SDK.
See https://dart.dev/go/sdk-version-pinning for details.


Because every version of drift_dev from path depends on analyzer ^6.5.0 and no versions of analyzer match >6.5.0 <7.0.0, every version of drift_dev from path requires analyzer 6.5.0.
And because analyzer 6.5.0 depends on meta ^1.14.0 and every version of flutter_test from sdk depends on meta 1.12.0, drift_dev from path is incompatible with flutter_test from sdk.
So, because app depends on both flutter_test from sdk and drift_dev from path, version solving failed.
```
@simolus3 
I moved back the analyzer and am working on a PR
I'm also adding a flutter test for versions 3.19 and 3.21

dickermoshe:
Fixes https://github.com/simolus3/drift/issues/3024 & https://github.com/simolus3/drift/issues/3015

I don't think we have to limit ourselves to whatever versions flutter limits, but we don't have to here.


drift/test/generated/todos.dart
```diff
@@ -1,7 +1,12 @@
 import 'package:drift/drift.dart';
-// ignore: import_of_legacy_library_into_null_safe
+import 'package:mockito/annotations.dart';
 import 'package:uuid/uuid.dart';
 
+// Generate mocks for drift
+@GenerateNiceMocks([MockSpec<TodoDb>()])
```
simolus3, line 6: Nice :+1: Glad to see that it's working now.

.github/workflows/main.yml
```diff
@@ -204,7 +203,25 @@ jobs:
...
+      - name: Setup Flutter
+        uses: subosito/flutter-action@v2
+        with:
+          # flutter-version: ${{ matrix.version }}
+          channel: stable
+          cache: true
+          cache-key: "flutter-:os:-:channel:-:version:"
+      - name: Install dependencies in drift/example/app
+        run: |
+          cd examples/app
```
simolus3: Can we still apply the `prepare` action here? That one caches a lot of pub dependencies so it might speed up the CI.

.github/workflows/main.yml
```diff
@@ -204,16 +204,16 @@ jobs:
           dart pub upgrade
           dart test
   test_flutter:
-    # strategy:
-    #   matrix:
-    #     version: ["3.16.0", "3.19.0", "3.21.0"]
+    strategy:
+      matrix:
+        version: [3.16.0, 3.19.0, 3.21.0]
```
simolus3: IMO supporting the latest stable version is good enough. I guess we have no choice but to support the older versions of packages pinned by Flutter stable, but going further back sounds annoying for relatively little value. If the latest drift release doesn't support the latest Flutter that's bad, but I'm fine with not supporting older Flutter versions.

dickermoshe: Sure

.github/workflows/main.yml
dickermoshe: This `subosito/flutter-action@v2` caches by itself

dickermoshe: Take a look here https://github.com/subosito/flutter-action?tab=readme-ov-file#caching

This action installs flutter. We could replace the dart setup with flutter setup, but that would mean we could never run any actions on versions of dart that aren't part of a flutter release.

dickermoshe:
For developing bleeding edge features (macros), I think we should use another branch besides develop.

.github/workflows/main.yml
simolus3: I meant package dependencies from pub.dev which are cached from `~/.pub-cache` across runs - I agree with using dart setup for most of the CI and flutter setup here.

dickermoshe: I think those are cached too

https://github.com/subosito/flutter-action/blob/44ac965b96f18d999802d4b807e3256d5a3f9fa1/action.yaml#L103-L111

dickermoshe: Nope, it's not working.
Would you be able to work on this?
I have very little experience with GitHub actions.

simolus3: Sure - is there an actions run with this I could take a look at to see where it went wrong? I couldn't find one from a quick look, but I can also just try it tomorrow.

dickermoshe: > Sure - is there an actions run with this I could take a look at to see where it went wrong?    

What's an "actions run"? I'm not understanding 🤷

The only part here that isn't currently working is the cache for pub.
I'm assuming whatever you did for the main dart cache can be copied for here.
The cache for the entire flutter sdk works great.

dickermoshe: `pub-cache-key` works, but it just seems to sensitive to changes

Tip: Install `act` and `github-act-cache-server`, makes this much easier than committing empty and pushing to github

simolus3: > What's an "actions run"? I'm not understanding 🤷

Sorry, I meant a workflow run of the GitHub action as you've configured it.

>  The cache for the entire flutter sdk works great.

And actually, `flutter-action` also sets up a cache for pub dependencies by itself. So I think we don't need to use our custom cache here, we can leave it as it is.

dickermoshe:
@simolus3 Thanks for finishing this up!

PR #3030, opened 2024-06-04
fix: Update BigInt extension to handle JavaScript bit length difference

gaetschwartz:
This PR fixes the `BigInt` bound check testing that previous used hardcoded min/max values for 64-bit signed ints, but JS only supports up to 53-bit ints in Dart.

This prevented the package from being built on Wb when the extension was used.

https://api.flutter.dev/flutter/dart-core/int-class.html

### DIFF
--- a/extras/drift_postgres/lib/src/pg_database.dart
+++ b/extras/drift_postgres/lib/src/pg_database.dart
@@ -220,8 +220,10 @@ class _PgVersionDelegate extends DynamicVersionDelegate {
 }
 
 extension on BigInt {
-  static final _bigIntMinValue64 = BigInt.from(-9223372036854775808);
-  static final _bigIntMaxValue64 = BigInt.from(9223372036854775807);
+  static const _isJs = identical(1.0, 1);
+  static const _allowedBitLength = _isJs ? 53 : 63;
+  static final _bigIntMinValue64 = BigInt.from(1 << _allowedBitLength);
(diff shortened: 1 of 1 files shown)


simolus3:
I appreciate the contribution, but I'm not sure I understand it fully: This affects `drift_postgres`, a package requiring TCP sockets not available on the web either way, so it's fundamentally incompatible with the web.

Is the problem that you have a shared package using postgres/sqlite3 and picking the right option at runtime depending on where the code is running? I think that could potentially be optimized with conditional imports to avoid importing `drift_postgres` on web code at all, but I'm also fine with this to fix the problem. It would be good to know that that's indeed the problem you have though, as I'll probably forget about the "this package needs to be compilable with dart2js" thing. So I need to come up with a way to test this as well.

gaetschwartz:
Indeed, it is a transitive I'm depending on as a shared package. 

That said, I don't see how that's a problem ? Couldn't one just use WebSocket to communicate to the Postgres database ? (Let's just leave that for another issue maybe ?)

simolus3:
I have fixed this in b499e5588b433ee9cb4627aaef2f7243665bb64c which maintains the old behavior but supports being compiled with dart2js. I think we should keep the behavior since that code can only really run in a Dart VM with `dart:io` support.

> That said, I don't see how that's a problem ? Couldn't one just use WebSocket to communicate to the Postgres database

WebSockets aren't regular TCP connections, since Postgres doesn't support a WebSocket server this is not possible without a third-party proxy solution.

PR #3037, opened 2024-06-07
fix:addColumn occurred An error,  When migrating from a lower version…

linchen00:
When a new table is created during the migration from an earlier version to a higher version, an error occurs when you call addColumn after creating the table with the latest column names

### DIFF
--- a/drift/lib/src/runtime/query_builder/migration.dart
+++ b/drift/lib/src/runtime/query_builder/migration.dart
@@ -429,6 +429,20 @@ class Migrator {
 
   /// Adds the given column to the specified table.
   Future<void> addColumn(TableInfo table, GeneratedColumn column) async {
+    final List<QueryRow> queryRows = await database.customSelect(
+      'SELECT name FROM pragma_table_info(?);',
+      variables: <Variable<Object>>[Variable<String>(table.actualTableName)],
+    ).get();
+
+    final List<String> columnNameList = queryRows
+        .map((QueryRow e) => e.readNullable<String>('name'))
+        .whereType<String>()
+        .toList();
+
+    if (columnNameList.contains(column.name)) {
+      return;
+    }
+
     final context = _createContext();
 
     context.buffer.write(


simolus3:
I don't think the current behavior is broken - it adds a column and fails if the column already exists, that's reasonable.

It's true that not being able to create the table at an earlier snapshot to then apply subsequent migrations to it can make the entire migration script a bit more complicated. If you still have the ability to reconstruct older versions, I strongly recommend using generated [step-by-step migrations](https://drift.simonbinder.eu/docs/migrations/step_by_step/): You can make drift generate snapshots of all your schema versions and then call `create(schema.yourTable)` in one step to create `yourTable` at the version expected at that step (without the new column). Then you call `addColumn(schema.yourTable, schema.yourTable.newColumn)` in a later step to reliably add the column.

Step-by-step migrations are probably the best answer to fix these issues, you can also use related tooling to unit test your migrations which can provide a lot of confidence of not getting these wrong for some combination of `from`, `to` targets. Otherwise you'd have to check the `from` version only to either create the table if you're coming from versions which didn't have it or alternatively create the column.

orestesgaolin:
Just to add my $0.02 - in our real world cases users have been switching between app versions on Testflight which lead to app being stuck on launch, as the migration has been applied the second time. If there was an option to modify table (e.g. add column) in idempotent way, I'd love to have it, e.g. as 

```dart
from1To2: (m, schema) async {
  await m.addColumnIfNotExists(
    schema.table,
    schema.table.column,
  );
},
```

simolus3:
I think idempotent migration APIs alone aren't too helpful in the end when you have more complex migrations consisting of multiple steps (and especially once you start reading data its over, but that's implicitly required for things like `alterTable`).

Serverside migration frameworks usually support going both ways (so you'd define an `_up` and a `_down` step for every migration, allowing them to be rolled back if necessary). That's technically supported in drift too, but not with the step-by-step migration helper.
I don't think it's too helpful for apps though, as older released versions of your app obviously won't know how to roll back the schema from a newer versions they don't know about yet.
I don't have a good answer to the beta test + testflight problem to be honest. I think you should raise your schema version even for versions you only roll out to beta testers (which covers users upgrading from stable release -> testflight). Supporting users downgrading your app is very tricky (e.g. what if a testflight version has deleted a table no longer necessary, you can't just implicitly go back from there). The best approach I can come up with is to inform users that they loose local data after downgrading and delete the database if the schema version in the database is higher than the one the app knows about. Or use a different database file for beta builds.

linchen00:
thanks,I will close it

PR #3040, opened 2024-06-10
Add `exclusively` API

### LINKED ISSUE #3032: Option to run a few queries sequentially
Currently the `DelegatedDatabase` has the notion of `isSequential` in order to support parallel queries and transactions.
On ElectricSQL they recently introduced the option to disable the foreign key checks when replicating from remote, as it has better performance on web. 
In order to do that it's necessary to do that in a safe manner, they do the following in a mutex. This is because foreign_keys cannot be called during a transaction.
```
PRAGMA foreign_keys = OFF;
SQLite transaction block {
	...
}
PRAGMA foreign_keys = ON;
``` 

Given that Drift can already work with sequential databases we were wondering if it would be feasible to add a function that runs a callback inside the same `synchronized` Lock. Something like `runSequentially` or `runExclusively`.

Context:
https://github.com/electric-sql/electric/pull/1281
https://github.com/electric-sql/electric/blob/64a01949376536492fbc72fe6bf0d8539a3552b0/clients/typescript/src/util/transactions.ts#L23

simolus3:
This adds an `exclusively` method to `DatabaseConnectionUser`, which blocks the current scope without explicitly starting a transaction.
This API is useful for things that should run atomically to other concurrent contexts accessing the database while being impossible to do in a transaction, like setting the `foreign_keys` pragma. For instance, this:

```dart
await exclusively(() async {
  await customStatement('pragma foreign_keys = off;');
  await transaction(() async {
    // ...
  });
  await customStatement('pragma foreign_keys = on;');
});
```

ensures that absolutely no statements can run between the transaction and the pragma statements, even if the database is shared across isolates with `DriftIsolate`.

I still need to do some minor work here like adding changelog entries and documentation on the website about this. I don't expect that this will be used often though, `transaction` is likely good enough most of the time.

Closes https://github.com/simolus3/drift/issues/3032. cc @davidmartos96, is this what you had in mind?

### DIFF
--- a/drift/CHANGELOG.md
+++ b/drift/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## 2.19.0-dev
 
+- Add `exclusively` method to database classes, allowing a block to temporarily
+  take exclusive control over a database connection without starting a
+  transaction.
 - Add the `enableMigrations` parameter to `WasmDatabase` to control whether drift
   migrations are enabled on that database.
 - Add `initiallyDeferred` option to `references()` column builder for foreign
--- a/drift/lib/src/remote/client_impl.dart
+++ b/drift/lib/src/remote/client_impl.dart
@@ -132,6 +132,16 @@ abstract class _BaseExecutor extends QueryExecutor {
 
     return result.rows;
   }
+
+  @override
+  QueryExecutor beginExclusive() {
+    return _RemoteExclusiveExecutor(client, _executorId);
+  }
+
+  @override
+  TransactionExecutor beginTransaction() {
+    return _RemoteTransactionExecutor(client, _executorId);
+  }
 }
 
 class _RemoteQueryExecutor extends _BaseExecutor {
@@ -139,11 +149,6 @@ class _RemoteQueryExecutor extends _BaseExecutor {
 
   Completer<void>? _setSchemaVersion;
   Future<bool>? _serverIsOpen;
-
-  @override
-  TransactionExecutor beginTransaction() {
-    return _RemoteTransactionExecutor(client, _executorId);
-  }
 
   @override
   Future<bool> ensureOpen(QueryExecutorUser user) async {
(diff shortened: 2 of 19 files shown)


davidmartos96:
@simolus3 Yes, that was exactly the API I had in mind and using the same lock under the hood.
Thank you!

davidmartos96:
@simolus3 Hello Simon! I'm trying out the new API, but it's not behaving how I expected. Is there something I'm missing?

I expect the following test to pass, but it fails. The test can be added to `integration_tests/database_test.dart` in the drift repo.

```dart
test('exclusively', () async {
    final db = TodoDb(NativeDatabase.memory());
    addTearDown(db.close);

    bool query1Finished = false;

    // Make a slow query and check that it is not interleaved with other queries/transactions
    final fut1 = db.exclusively<int>(() async {
      await Future<void>.delayed(const Duration(milliseconds: 100));
      await db.customSelect('SELECT 1').get();
      query1Finished = true;
      return 7;
    });

    final fut2 = db.transaction<int>(() async {
      expect(query1Finished, true);
      return 5;
    });

    final results = await Future.wait<void>([fut1, fut2]);
    expect(results, [7, 5]);
  });
```

simolus3:
The behavior here isn't outright broken, it's just very weird. In `QueryExecutor`, `beginTransaction` and `beginExclusive` return an executor right away, the lock is only obtained asynchronously when that executor is first used. That is mostly a historical decision, it matches the behavior with the top-level database that can be constructed synchronously even if the database requires async setup work (like with `LazyDatabase`).

In your test, a consequence of this is that the transaction block runs right away, and it's not waiting on the exclusively block since you're not accessing the database in the transaction. Worse, if you add a statement to the transaction before the expectation, the transaction will actually run _before_  the exclusively block since that one will only obtain the lock when running the first statement, after the wait.

One could argue that the current behavior is better because it locks the database for a shorter period of time (and it could be relevant if users are doing weird stuff like a network call in a transaction block before writing a batch), but I think it's rather weird. All of the docs mention that "the block will run in a transaction" when it actually runs in a weird half-state where a transaction is not yet started. I've changed the behavior in 9ba81d53a014fb4cf101fbedb9e431297f611a2f - let's see if we get away with this or if this breaks anyone (in which case we'll only do this for `exclusively`).

PR #3051, opened 2024-06-16
Builtin support for dbstat table

FaFre:
After https://github.com/simolus3/sqlite3.dart/issues/237 landed, I added the dbstat table as a builtin table to the sqlparser.

Open question is, shall the inclusion of the dbstat table be controlled by an option (like its the case for json or fts)?

### DIFF
--- a/sqlparser/lib/src/engine/builtin_tables.dart
+++ b/sqlparser/lib/src/engine/builtin_tables.dart
@@ -32,3 +32,70 @@ Table get sqliteSchema {
     resolvedColumns: sqliteMaster.resolvedColumns,
   );
 }
+
+/// The DBSTAT virtual table is a read-only eponymous virtual table that returns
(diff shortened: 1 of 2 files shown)


simolus3 (commented):
Thanks! But I think this should be guarded behind an option (we're usually pretty cautious with the functionality we assume, we also expect a fairly old sqlite3 version without a builder option telling us. This helps users on platforms not using `sqlite3_flutter_libs`).

I think the best way to expose this is to put dbstat into a module and also refactor `sqlparser` so that modules can add their own default tables. That would also potentially allow us to analyze things like `CREATE VIRTUAL TABLE stat USING dbstat(main)`.

simolus3:
Merged in db75bce0d4ec41fbdf5aba74bd1ada14f4e24efc, I'll make this opt-in with a `dbstat` option you'd have to list in `build.yaml`.

PR #3059, opened 2024-06-23
Add Flutter-specific `drift_flutter` package

simolus3:
This adds `drift_flutter`, providing a `driftDatabase` method returning a `QueryExecutor` and an implementation following best practices currently described in the documentation.
This makes drift much easier to set up in Flutter apps, as this code can be bundled in this package and doesn't have to be copied into every app.

Remaining todos:

- adding an example in the package
- adding flutter integration tests
- rewriting the docs to use this package (but also keeping the existing bits somewhere so that CLI/backend devs using drift can easily find them)


dickermoshe:
Will `drift_flutter` re-export `drift`?
Will `sqlite3_flutter_libs` be included automatically?

It would be awesome if we didn't require people install so many separate dependencies directly.

I'm so excited for this

PR #3062, opened 2024-06-26
Refactor Manager

dickermoshe:
This PR does the following:

* Cherry Pick the Flutter CI fixes from the `drift_flutter` branch
* Add a option for combining successive filter calls with an OR instead of a hardcoded AND.
* Rename Generics in the Manager API to more descriptive names (`$Table` instead of `T`)
* Make Manager classes non-const for now
* Rename some fields on `ManagerState`
* Remove unnecessary code from Manager API (`_getChildManagerBuilder`)
* Add distinct option for some queries

I will create the next PR once this is merged.

NOTE: I "fixed" a "bug" and then reverted it. 
Lmk if I should go back and undo the commits so we should have a cleaner git history.

I want to make this as painless for you as possible.
The next feature (references) will introduce another level of complexity and I don't want to be the only person with a strong understanding of it.  

I'll be documenting the next PR very clearly.
I'm looking for constructive criticism if you ever have any for me.

### DIFF
--- a/drift/lib/src/runtime/manager/filter.dart
+++ b/drift/lib/src/runtime/manager/filter.dart
@@ -285,7 +285,7 @@ extension DateFilters<T extends DateTime> on ColumnFilters<T> {
 }
 
 ///  Enum of the possible boolean operators
-enum BooleanOperator {
+enum _BooleanOperator {
   /// Combine the existing filters to the new filter with an AND
   and,
 
@@ -311,22 +311,22 @@ class ComposableFilter extends _Composable {
 
   /// Combine two filters with an AND
   ComposableFilter operator &(ComposableFilter other) =>
-      _combineFilter(BooleanOperator.and, other);
+      _combineFilter(_BooleanOperator.and, other);
 
   /// Combine two filters with an OR
   ComposableFilter operator |(ComposableFilter other) =>
-      _combineFilter(BooleanOperator.or, other);
+      _combineFilter(_BooleanOperator.or, other);
 
   /// A helper function to combine two filters
   ComposableFilter _combineFilter(
-      BooleanOperator opperator, ComposableFilter otherFilter) {
+      _BooleanOperator opperator, ComposableFilter otherFilter) {
     final combinedExpression = switch ((expression, otherFilter.expression)) {
       (null, null) => null,
       (null, var expression) => expression,
       (var expression, null) => expression,
       (_, _) => switch (opperator) {
-          BooleanOperator.and => expression! & otherFilter.expression!,
-          BooleanOperator.or => expression! | otherFilter.expression!,
+          _BooleanOperator.and => expression! & otherFilter.expression!,
+          _BooleanOperator.or => expression! | otherFilter.expression!,
         },
     };
     return ComposableFilter._(
--- a/drift/lib/src/runtime/manager/filter.dart
+++ b/drift/lib/src/runtime/manager/filter.dart
@@ -31,7 +31,7 @@ abstract class _BaseColumnFilters<T extends Object> {
   /// This helper method is used internally to create a new [ComposableFilter]s
   /// that respects the inverted state of the current filter
   ComposableFilter $composableFilter(Expression<bool>? expression) {
-    return ComposableFilter.$_(
+    return ComposableFilter._(
         inverted ? expression?.not() : expression, joinBuilders);
   }
 
@@ -307,7 +307,7 @@ class ComposableFilter extends _Composable {
   Expression<bool>? expression;
 
   /// Create a new [ComposableFilter] for a column with joins
-  ComposableFilter.$_(this.expression, this.joinBuilders);
+  ComposableFilter._(this.expression, this.joinBuilders);
 
   /// Combine two filters with an AND
   ComposableFilter operator &(ComposableFilter other) =>
@@ -329,7 +329,7 @@ class ComposableFilter extends _Composable {
           BooleanOperator.or => expression! | otherFilter.expression!,
         },
     };
-    return ComposableFilter.$_(
+    return ComposableFilter._(
       combinedExpression,
       joinBuilders.union(otherFilter.joinBuilders),
     );
--- a/drift/lib/src/runtime/manager/filter.dart
+++ b/drift/lib/src/runtime/manager/filter.dart
@@ -31,7 +31,7 @@ abstract class _BaseColumnFilters<T extends Object> {
   /// This helper method is used internally to create a new [ComposableFilter]s
   /// that respects the inverted state of the current filter
   ComposableFilter $composableFilter(Expression<bool>? expression) {
-    return ComposableFilter._(
+    return ComposableFilter.$_(
         inverted ? expression?.not() : expression, joinBuilders);
   }
 
@@ -284,7 +284,14 @@ extension DateFilters<T extends DateTime> on ColumnFilters<T> {
       $composableFilter(column.isBetweenValues(lower, higher));
 }
 
-enum _BooleanOperator { and, or }
+///  Enum of the possible boolean operators
+enum BooleanOperator {
+  /// Combine the existing filters to the new filter with an AND
+  and,
+
+  /// Combine the existing filters to the new filter with an OR
+  or;
+}
 
 /// This class is used to compose filters together
 ///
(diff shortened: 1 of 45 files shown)


dickermoshe:
@simolus3 
This PR is done.
We are ready for review

drift/lib/src/runtime/manager/filter.dart
```diff
@@ -300,29 +307,29 @@ class ComposableFilter extends _Composable {
   Expression<bool>? expression;
 
   /// Create a new [ComposableFilter] for a column with joins
-  ComposableFilter._(this.expression, this.joinBuilders);
+  ComposableFilter.$_(this.expression, this.joinBuilders);
```
simolus3: Is the rename because this is now meant to be called from generated code directly? In that case we should expand the doc comment to mention that this is not meant to be called by user code (or maybe just mark it as `@internal` since generated code ignores lints).

drift/lib/src/runtime/manager/manager.dart
```diff
@@ -9,45 +9,27 @@ part 'filter.dart';
...
-    C extends ProcessedTableManager<DB, T, DT, FS, OS, C, CI, CU>,
-    CI extends Function,
-    CU extends Function> {
+    $Database extends GeneratedDatabase,
+    $Table extends Table,
+    $Dataclass,
+    $FilterComposer extends FilterComposer<$Database, $Table>,
+    $OrderingComposer extends OrderingComposer<$Database, $Table>,
+    $CreateCompanionCallback extends Function,
+    $UpdateCompanionCallback extends Function> {
```
simolus3, line 27: Nice idea! This is much better, thanks.

drift/lib/src/runtime/manager/manager.dart
```diff
@@ -389,8 +391,8 @@ abstract class BaseTableManager<
   ///
   /// Uses the distinct flag to ensure that only distinct rows are returned
```
simolus3: Let's update the comment as well, e.g. with

```suggestion
  /// The [distinct] parameter (enabled by default) controls whether to generate
  /// a `SELECT DISTINCT` query, removing duplicates from the result.
```

drift/lib/src/runtime/manager/manager.dart
```diff
@@ -450,50 +456,49 @@ abstract class BaseTableManager<
...
+  Stream<$Dataclass?> watchSingleOrNull({bool distinct = true}) => $state
+      .copyWith(distinct: distinct)
       .buildSelectStatement()
       .watchSingleOrNull();
 }
 
 /// A table manager that exposes methods to a table manager that already has filters/orderings/limit applied
 //  As of now this is identical to [BaseTableManager] but it's kept seperate for future extensibility
+@immutable
 class ProcessedTableManager<
```
simolus3, line 501: We're currently generating typedefs to this class, but these aren't being referenced by generated code elsewhere.

So I'm wondering if we even need to have this class at the moment? If we have it to add new features in the future like the documentation suggest, shouldn't it be referenced in code? I understand if we explicitly reference `RootTableManager` now and then realize that we should have been referencing a more general manager class instead, we have a problem - and that justifies having `ProcessedTableManager` around.
But as far as I can see, we're not currently using this class outside of deaf typedef. So what's the point of having it? Given that it's not currently used, couldn't we just introduce separate classes when we need them and only start referencing them then? Or am I overlooking something here?

drift/lib/src/runtime/manager/manager.dart
```diff
@@ -292,53 +275,71 @@ class TableManagerState<
...
         joinBuilders: $state.joinBuilders.union(orderings.joinBuilders)));
   }
 
   /// Add a filter to the statement
-  C filter(ComposableFilter Function(FS f) f) {
+  ///
+  /// The [combineWith] parameter can be used to combine the new filter with any existing filters
+  ProcessedTableManager<$Database, $Table, $Dataclass, $FilterComposer,
+          $OrderingComposer, $CreateCompanionCallback, $UpdateCompanionCallback>
+      filter(ComposableFilter Function($FilterComposer f) f,
```
simolus3: I wonder if having two methods instead of the `combineWith` parameter may be easier?

```dart
        db.managers.tableWithEveryColumnType
            .filter((f) => f.aText("person"))
            .or((f) => f.aReal(5.0))
```

But then we kind of start mirroring the filter API in the general manager API which is also not too nice... Maybe we can also keep it as it is, what you do think?

drift/lib/src/runtime/manager/manager.dart
simolus3: Ah I see we're using it as a return value - that's fair enough, but can we remove the typedef in generated code?

simolus3 (commented):
Thanks! I agree that the full names for the type arguments make the manager code much easier to grasp (and the new companion typedef names are also clearer). I don't think we need to rebase/clean up the history.

Revisiting the code, I'm not quite sure what justifies having `ProcessedTableManager` as a separate class _right now_, is this required for references or another feature you have in mind that would be breaking without it?

drift/lib/src/runtime/manager/filter.dart
dickermoshe: Turns out, that this should've been reverted along with https://github.com/simolus3/drift/pull/3062/commits/63ba8fec47ac6915ac1fff87096464f2cc1b920c .
I'll reverse this too.

drift/lib/src/runtime/manager/manager.dart
dickermoshe: Done

drift/lib/src/runtime/manager/manager.dart
dickermoshe: This is done in preparation for the references pull request.

There will need to combine managers, each with their own query, into 1 manager which has all the filters combined with an OR statement.

However, being that this is only needed internally, I'll make it private.

drift/lib/src/runtime/manager/manager.dart
dickermoshe: @simolus3 
I will remove the unnecessary typedef.

However:

In a future breaking change I will rearrange where certain method are placed.

### Future Plans

Method are broken up into 3 categories.

Top level methods:
```dart
users.create(...)
users.bulkCreate(...)
users.replace(...)
users.bulkReplace(...)
```

Row Specific methods:
```dart
users.update(...)
users.exists(...)
users.delete(...)
users.get(...),users.getSingle(...),users.getSingleOrNull(...)
users.watch(...),users.watchSingle(...),users.watchSingleOrNull(...)
```

Manager Modification methods: (changes the manager in some way)
```dart
users.filter(...)
users.order(...)
```
---

When someone does any "Manager Modification" method, it shows that he wants to perform some "Row Specific" method.
```dart
users.filter(...).get();
``` 

However, there is no reason for a "Manager Modification" method and a "Top level" method to be applied to the same filter.
```dart
users.filter(...).create(...);
``` 
Therefore, we only put the "Top level" methods on a root manager.
This is already implemented right now.

--- 

We also want to name things more appropriately on the root level manager.
For example right now `users.delete(...) ` would delete ALL users, this should be named, `users.deleteAll()` on the root manager.

Therefore, certain "Row Specific" methods should only be on managers that have has a "Manager Modification" performed.

```dart
users.filter(...).delete()
users.filter(...).get()
users.all().get()
```
and some just on the root manager:
```dart
users.deleteAll()()
```
some should exist everywhere:
```dart
users.exist(...)
```

So therefore, we will eventually need 3 classes, a `BaseClass`, a `ProccessedClass` and a `RootClass`.

We could remove it for now, but I rather not have to shoehorn it back in later.

Thoughts?

dickermoshe:
@simolus3 
Made changes and ready for supplemental review

simolus3:
Thanks for the quick follow-up and the detailed explanation! I'll check them tomorrow, I wanted to get the docs update for `drift_flutter` done today and that took more time than expected.

drift/lib/src/runtime/manager/manager.dart
simolus3: That makes total sense, thanks for the detailed explanation!

I have added a brief section about the `filter` + `create` incompatibility as a motivating example to the `ProcessedTableManager` docs.

simolus3 (approved):
Thanks, this looks good to me now!

dickermoshe:
@simolus3 

We have 2 ways to proceed for a `UserWithReferences` class.

Option 1: Wrap the Class
```dart
// Implementation
class UserWithReferences {
  UserWithReferences(this.user)
  final User user;
  final GroupManager get group => GroupManager(...).filter(...);
}

// Usage

final usersWithRefs = await users.withReferences(...).get()
for (var userWithRefs in usersWithRefs){
  final user = userWithRefs.user;
  final group = await userWithRefs.group.getSingle()
}
```
### Pros
* The `UserWithReferences` class is small

##### Cons
* Usage of this class is bloated

### Option 2: Extend the Class
```dart
// Implementation
class UserWithReferences extends User{
  UserWithReferences(
    super.name,
    super.age,
    super.gender,
   /// ... and all the other fields
)
  final GroupManager get groupManager => GroupManager(...).filter(...);
}

// Usage
final users = await users.withReferences(...).get()
for (var user in users ){
  //  final user = userWithRefs.user; NOT NEEDED
  final group = await user.groupManager.getSingle()
}

```
##### Pros
* Using the results is much cleaner.
* If one migrated to using a reference at any point, they would not need to refactor.

##### Cons
* The generated code would be much longer
* We couldn't use the term `group`, because it would clash with the group id field on the original model. We will need to use `groupManager` instead. However, this isn't so bad. Because a field named `group` that doesnt return an instance of `Group`, but rather a `GroupManager`, doesn't seem that great either.


What are your thoughts?
 How could this work for custom classes?

simolus3:
I think I'm in favor of option 1 here, mostly for the reasons you've also mentioned (smaller code size, field name conflicts impossible). I also don't see how option 2 would work with custom classes (because sometimes there isn't even a class to extend, it could also be a record typedef).

I agree that the API is more convenient for users in option 2 - I wonder if we could use records to perhaps make this a bit easier? E.g.

```dart
final usersWithRefs = await users.withReferences(...).get()
for (var (user, refs) in usersWithRefs){
  final user = userWithRefs.user;
  final group = await refs.group.getSingle()
}
```

Or perhaps even:

```dart

final (users, refs) = await users.withReferences(...).get()
for (var user in users){
  final group = await refs.group(user).getSingle()
}
```

That approach would still be possible with little generated code and it's compatible with whatever the shape of the data class/type is. It may be a little too clever, but if we establish this pattern of destructuring as how it's done, I don't think it's confusing.

PR #3070, opened 2024-06-27
feat: Read References

dickermoshe:
This PR adds the ability to do `withReferences` to get a manager that will return the row, along with pre-filtered managers for the related values.

This is how this feature work.

Every Manager has a Generic for holding the Dataclass for a row. 
It's called `$Dataclass`, and it refers to the original data class that Drift uses when working with the table. E.G `$User`

We now create another class called `<name>WithReferences` which contains the original dataclass, along with all the references. This new `$UserWithReferences` class is now ALSO a generic on the manager called `$MappedDataclass`

However, we need to keep track which data class will be returned.
For this we have a 3rd generic `$ActiveDataclass`, this is set to `$Dataclass` by default, but when `users.withReferences` is called, we return a Manager with `$MappedDataclass` as the `$ActiveDataclass`.

To actually perform the mapping, we have a callback on the manager that applies a mapping to each item in the database.

It's all getting very confusing. Let me know how If can help clarify anything.

---

This potentially has a breaking change if people have filters that utilize references with Type Converters that aren't symmetrical.
This causes a bug where needless joins are being created, so it's not a breaking change.

People may get warnings when re-generating, warnings that they've never seen before, which is jarring to them. I wonder what we should do alleviate that.

@simolus3 Ready for review

### DIFF
--- a/docs/lib/snippets/_shared/todo_tables.drift.dart
+++ b/docs/lib/snippets/_shared/todo_tables.drift.dart
@@ -335,7 +335,9 @@ class $$TodoItemsTableTableManager extends i0.RootTableManager<
     i1.$$TodoItemsTableFilterComposer,
     i1.$$TodoItemsTableOrderingComposer,
     $$TodoItemsTableCreateCompanionBuilder,
-    $$TodoItemsTableUpdateCompanionBuilder> {
+    $$TodoItemsTableUpdateCompanionBuilder,
+    $$TodoItemsTableWithReferences,
+    i1.TodoItem> {
   $$TodoItemsTableTableManager(
       i0.GeneratedDatabase db, i1.$TodoItemsTable table)
       : super(i0.TableManagerState(
@@ -345,6 +347,8 @@ class $$TodoItemsTableTableManager extends i0.RootTableManager<
               i1.$$TodoItemsTableFilterComposer(i0.ComposerState(db, table)),
           orderingComposer:
               i1.$$TodoItemsTableOrderingComposer(i0.ComposerState(db, table)),
+          dataclassMapper: (p0) async =>
+              p0.map((e) => $$TodoItemsTableWithReferences(db, e)).toList(),
           updateCompanionCallback: ({
             i0.Value<int> id = const i0.Value.absent(),
             i0.Value<String> title = const i0.Value.absent(),
@@ -375,6 +379,17 @@ class $$TodoItemsTableTableManager extends i0.RootTableManager<
           ),
         ));
 }
+
+typedef $$TodoItemsTableProcessedTableManager = i0.ProcessedTableManager<
+    i0.GeneratedDatabase,
+    i1.$TodoItemsTable,
+    i1.TodoItem,
+    i1.$$TodoItemsTableFilterComposer,
+    i1.$$TodoItemsTableOrderingComposer,
+    $$TodoItemsTableCreateCompanionBuilder,
+    $$TodoItemsTableUpdateCompanionBuilder,
+    $$TodoItemsTableWithReferences,
+    i1.TodoItem>;
 
 class $$TodoItemsTableFilterComposer
     extends i0.FilterComposer<i0.GeneratedDatabase, i1.$TodoItemsTable> {
@@ -457,6 +472,13 @@ class $$TodoItemsTableOrderingComposer
                     parentComposers)));
     return composer;
   }
+}
+
+class $$TodoItemsTableWithReferences {
+  // ignore: unused_field
+  final i0.GeneratedDatabase _db;
+  final i1.TodoItem i1TodoItem;
+  $$TodoItemsTableWithReferences(this._db, this.i1TodoItem);
 }
 
 class $CategoriesTable extends i2.Categories
@@ -655,7 +677,9 @@ class $$CategoriesTableTableManager extends i0.RootTableManager<
     i1.$$CategoriesTableFilterComposer,
     i1.$$CategoriesTableOrderingComposer,
     $$CategoriesTableCreateCompanionBuilder,
-    $$CategoriesTableUpdateCompanionBuilder> {
+    $$CategoriesTableUpdateCompanionBuilder,
+    $$CategoriesTableWithReferences,
+    i1.Category> {
   $$CategoriesTableTableManager(
       i0.GeneratedDatabase db, i1.$CategoriesTable table)
       : super(i0.TableManagerState(
@@ -665,6 +689,8 @@ class $$CategoriesTableTableManager extends i0.RootTableManager<
               i1.$$CategoriesTableFilterComposer(i0.ComposerState(db, table)),
           orderingComposer:
               i1.$$CategoriesTableOrderingComposer(i0.ComposerState(db, table)),
+          dataclassMapper: (p0) async =>
+              p0.map((e) => $$CategoriesTableWithReferences(db, e)).toList(),
           updateCompanionCallback: ({
             i0.Value<int> id = const i0.Value.absent(),
             i0.Value<String> name = const i0.Value.absent(),
@@ -683,6 +709,17 @@ class $$CategoriesTableTableManager extends i0.RootTableManager<
           ),
         ));
 }
+
+typedef $$CategoriesTableProcessedTableManager = i0.ProcessedTableManager<
+    i0.GeneratedDatabase,
+    i1.$CategoriesTable,
+    i1.Category,
+    i1.$$CategoriesTableFilterComposer,
+    i1.$$CategoriesTableOrderingComposer,
+    $$CategoriesTableCreateCompanionBuilder,
+    $$CategoriesTableUpdateCompanionBuilder,
+    $$CategoriesTableWithReferences,
+    i1.Category>;
 
 class $$CategoriesTableFilterComposer
     extends i0.FilterComposer<i0.GeneratedDatabase, i1.$CategoriesTable> {
@@ -710,6 +747,13 @@ class $$CategoriesTableOrderingComposer
       column: $state.table.name,
       builder: (column, joinBuilders) =>
           i0.ColumnOrderings(column, joinBuilders: joinBuilders));
+}
+
+class $$CategoriesTableWithReferences {
+  // ignore: unused_field
+  final i0.GeneratedDatabase _db;
+  final i1.Category i1Category;
+  $$CategoriesTableWithReferences(this._db, this.i1Category);
 }
 
 class $UsersTable extends i2.Users with i0.TableInfo<$UsersTable, i1.User> {
@@ -908,7 +952,9 @@ class $$UsersTableTableManager extends i0.RootTableManager<
     i1.$$UsersTableFilterComposer,
     i1.$$UsersTableOrderingComposer,
     $$UsersTableCreateCompanionBuilder,
-    $$UsersTableUpdateCompanionBuilder> {
+    $$UsersTableUpdateCompanionBuilder,
+    $$UsersTableWithReferences,
+    i1.User> {
   $$UsersTableTableManager(i0.GeneratedDatabase db, i1.$UsersTable table)
       : super(i0.TableManagerState(
           db: db,
@@ -917,6 +963,8 @@ class $$UsersTableTableManager extends i0.RootTableManager<
               i1.$$UsersTableFilterComposer(i0.ComposerState(db, table)),
           orderingComposer:
               i1.$$UsersTableOrderingComposer(i0.ComposerState(db, table)),
+          dataclassMapper: (p0) async =>
+              p0.map((e) => $$UsersTableWithReferences(db, e)).toList(),
           updateCompanionCallback: ({
             i0.Value<int> id = const i0.Value.absent(),
             i0.Value<DateTime> birthDate = const i0.Value.absent(),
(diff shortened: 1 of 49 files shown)


dickermoshe:
@simolus3 
I came across something very strange.
Why is `S` nullable here?

https://github.com/simolus3/drift/blob/06dbcc7ee70e2180648125581ea94f9aa1352c6f/drift/lib/src/runtime/query_builder/schema/column_impl.dart#L285-L309

I came across this when adding type converter support.
I found it very strange that I could create a `RowId` from `null` 

image

Turns out that we allow all types to be created from null.
Is this a bug?

dickermoshe:
Removing support for type converters would not be a breaking change.
It would be a bug fix.

There are elements of the current api that don't work when references don't use the same type.
It's a fairly rare occurrence, although, I think removing it is justified.

dickermoshe:
@simolus3 
Ready for review.
Once this is merged, I'll start working on prefetch.

This will be done in two step.
Some references can be gotten with a JOIN and others require another query.
I'll be doing the latter first.

simolus3:
> Why is `S` nullable here?

It should be possible for converters to map non-null Dart values to a null SQL value. This is only rarely done, but something that type converters generally support. In that case the column would be non-nullable in Dart but nullable in SQL.
I think doing `fromSql` like you did here is correct. Most of the time the converter will not have a nullable SQL type, but in that case the Dart runtime will insert the null check (and `drift_dev` will issue a warning if a converter that doesn't support null SQL values is used on a nullable column or wrap it into a `NullableTypeConverter` as applicable).

simolus3:
> Removing support for type converters would not be a breaking change.
It would be a bug fix.

Would we also remove type converter support in existing manager APIs (haven't looked at the code yet, will do that later today)? I think we should generally aim for orthogonality (if we support feature x and y, it's always better to have these features working together too).
Having a reference with the same underlying SQL type but different Dart types due to converters is super weird and I agree that we don't have to support that in the manager API which has to use the mapped Dart values in lots of places to be convenient. But generally type converters are pretty useful and we should support them in the manager API where possible.

drift/lib/src/runtime/manager/manager.dart
```diff
@@ -11,9 +12,14 @@ part 'ordering.dart';
 
 /// Defines a class that holds the state for a [BaseTableManager]
 ///
-/// It holds the state for manager of [$Table] table in [$Database] database, used to return [$Dataclass] data classes/rows.
+/// It holds the state for manager of [$Table] table in [$Database] database.
 /// It holds the [$FilterComposer] Filters and [$OrderingComposer] Orderings for the manager.
 ///
+/// There are 3 Dataclass generics:
+///   - [$Dataclass] is the dataclass that is used to interact with the table
+///   - [$MappedDataclass] refrences the type that will be returned if we chose to apply some mapping/wrapping to the data
```
simolus3: Will we ever apply a mapping that is not a "+ references"? If not, let's rename this to be something like `$DataclassWithRefs`. But even if we need a generic name, I think the docs should mention that we're primarily using this to represent a reference wrapper at the moment.

drift/lib/src/runtime/manager/manager.dart
```diff
@@ -69,6 +77,22 @@ class TableManagerState<
   /// for updating data in the table
   final $UpdateCompanionCallback _updateCompanionCallback;
 
+  /// This function will map every [$Dataclass] instance returned by the query to a [$MappedDataclass] instance
+  /// This method will be called by the [applyDataclassMapper] method. This method should never be called directly.
+  /// This is typically used to wrap the dataclass in a class that provides getters for the items references
+  final DataclassMapper<$Dataclass, $MappedDataclass> _dataclassMapper;
+
+  /// This function will apply the dataclass mapper to the given list of dataclasses,
+  /// but only if the generic types [$MappedDataclass] and [$ActiveDataclass] are the same
```
simolus3: Comments don't have to explain _what_ the code is doing, but rather _why_. In this case, maybe something like "applies the configured dataclass mapper if necessary (that is, a mapping is active)" may be more helpful.

drift/lib/src/runtime/manager/manager.dart
```diff
@@ -123,6 +152,35 @@ class TableManagerState<
...
+      $Database,
+      $Table,
+      $Dataclass,
+      $FilterComposer,
+      $OrderingComposer,
+      $CreateCompanionCallback,
+      $UpdateCompanionCallback,
+      $MappedDataclass,
+      $NewActiveDataclass> copyWithNewActiveDataclass<$NewActiveDataclass>() {
+    return TableManagerState(
```
simolus3, line 168: Would it be possible for this method to take the mapper (and us then simply applying the mapper if the field is non-null instead of introducing additional type parameters)?
If it's not easy to pull of it's probably not worth it, but I think it may be more intuitive than always having the field and then using type arguments as the selector.

drift/test/manager/manager_selectable_test.dart
```diff
@@ -0,0 +1,217 @@
...
+  late TodoDb db;
+
+  setUp(() {
+    db = TodoDb(testInMemoryDatabase());
+  });
+
+  tearDown(() => db.close());
+
+  test('manager - selectable tests', () async {
+    final stores = [
```
simolus3: Can we simplify the initialization here somewhat (perhaps by using non-manager batches or iterating over arrays)? That code makes up like 80% of this test.

drift/test/manager/manager_selectable_test.dart
```diff
@@ -0,0 +1,217 @@
...
+    final getAllStoresWithFilterAndOrderingWithReferences = db.managers.store
+        .filter((f) => f.id.not(10))
+        .orderBy((f) => f.id.asc())
+        .withReferences();
+
+    void testManager<T, M>(
+        BaseTableManager<dynamic, dynamic, T, dynamic, dynamic, dynamic,
+                dynamic, M, T>
+            selectable) {
+      expect(selectable.get().then((v) => v.length), completion(3));
```
simolus3: I think it would be easier to make `testManager` async and then do `expect((await selectable.get()).length, 3)`. Or, even better `expect(await selectable.get(), hasLengh(3))`

drift_dev/lib/src/writer/manager/manager_templates.dart
```diff
@@ -210,6 +236,7 @@ class _ManagerCodeTemplates {
         table: table,
         filteringComposer: ${filterComposerNameWithPrefix(table, leaf)}(${leaf.drift("ComposerState")}(db, table)),
         orderingComposer: ${orderingComposerNameWithPrefix(table, leaf)}(${leaf.drift("ComposerState")}(db, table)),
+        dataclassMapper: (p0)  async => p0.map((e) => ${rowWithReferencesClassName(table)}(db,e)).toList() ,
```
simolus3: Does this have to be `async`?

drift_dev/lib/src/writer/manager/manager_templates.dart
```diff
@@ -345,4 +372,43 @@ class _ManagerCodeTemplates {
         ))
               );""";
   }
+
+  /// Code for a row class which contains references to other tables
+  String? rowClassWithReferences(
```
simolus3, line 390: As I've mentioned in the other PR we should investigate whether we want to perhaps refactor this into records to generate less code and allow destructuring which could enable a nicer API. But I'm fine with this approach as well.

simolus3 (commented):
I agree with the approach and implementation for the most part, I think the biggest thing is perhaps shuffling the generated API around to make resolving references easier. But if that turns out to not be more ergonomic, the generated code changes here look good to me.

drift/test/manager/manager_selectable_test.dart
dickermoshe: Done

drift_dev/lib/src/writer/manager/manager_templates.dart
dickermoshe: Nope!

drift/lib/src/runtime/manager/manager.dart
dickermoshe: Sure

drift/lib/src/runtime/manager/manager.dart
dickermoshe: There would still be a need for 3 generics. 
The parent state still needs to know 3 things:

1. The actual Dataclass of the Table
2. **IF** we run `withReferences`, what we then be returned.
3. What will `.get` return if we ran it now.

We could use a field to be the deciding factor, but it would not reduce any generics.

dickermoshe:
The issue with type converters is that they create a completely different column type under the hood.
Parts of the Filtering API rely on the fact that the types of reference be the same.

This was done so that filtering looked better.
Instead of having 2 fields for filtering on a relation
```dart
users.filter((f)=>f.groupId(5))
users.filter((f)=>f.group.name(5))
```
We have 1:
```dart
users.filter((f)=>f.group.id(5))
users.filter((f)=>f.group.name(5))
```
Where the Filtering API knows that `f.group.id(5)` does NOT require a join, but `f.group.name(5)` does.

This hasn't come up until now because we've only been testing filtering on a type converter in one direction
```
id:    GeneratedColumnWithTypeConverter
group: GeneratedColumn
users.filter((f)=>f.group.id(5))
```
However in the reverse
```
group:    GeneratedColumnWithTypeConverter
id: GeneratedColumn
users.filter((f)=>f.group.id(5)) // 
```
this throws an error. This is because `GeneratedColumnWithTypeConverter` extends `GeneratedColumn`, but not vice versa.


I've added a regression test for this.

drift_dev/lib/src/writer/manager/manager_templates.dart
dickermoshe: Let me take a look

dickermoshe: Done.
Not sure If like it better or not.

dickermoshe:
> Would we also remove type converter support in existing manager APIs (haven't looked at the code yet, will do that later today)?

It would remove filters and ordering for relations that don't align.
E.G
```dart
users.filter((f)=>f.group.id(5))
```
would no longer be generated until the user put a type converter on `User.group` too

This is also what happens when there are 2 relation whose name clash.

drift/lib/src/runtime/manager/manager.dart
simolus3: In that case we can keep it as is.

drift_dev/lib/src/writer/manager/manager_templates.dart
simolus3: I agree it's not that nice with the tests using `.then` but it might be cleaner when doing an assignment in an async/await function?

docs/lib/snippets/_shared/todo_tables.drift.dart
```diff
@@ -459,6 +475,14 @@ class $$TodoItemsTableOrderingComposer
   }
 }
 
+class $$TodoItemsTableWithReferences {
+  // ignore: unused_field
```
simolus3: Do we really still need this with the `ignore_for_file: type=lint` at the top?

docs/lib/snippets/modular/custom_types/table.drift.dart
```diff
@@ -319,3 +336,11 @@ class $$PeriodicRemindersTableOrderingComposer extends i0
       builder: (column, joinBuilders) =>
           i0.ColumnOrderings(column, joinBuilders: joinBuilders));
 }
+
+class $$PeriodicRemindersTableWithReferences {
```
simolus3: Instead of generating empty classes here, is it possible to define a generic class in `package:drift` and use that as a type argument to the manager for tables that don't have references? Then we don't have to generate what is essentially dead code.

simolus3 (approved):
This looks good me apart from the two minor nits. Thanks a lot for the contribution, awesome work!

docs/lib/snippets/_shared/todo_tables.drift.dart
dickermoshe: Done

docs/lib/snippets/modular/custom_types/table.drift.dart
dickermoshe: Done

dickermoshe:
I'll start working on prefetching.

If you think it's complex now, it's going to get much worse, with functions returning functions which returns more function which eventually return data.

I'm going to take the time to ensure that everything is named appropriately and that the doccoments explain how it works.

Please ask me before you do any sort of release. There is a possibility that there may be breaking changes from this PR.

The basic idea is that `withReferences((o)=> o(prefetchUser: true)).get()` will get all the users before and put them in a cache.

dickermoshe:
This is how prefetching will work under the hood

### Step 1 -  The Callback

The user decides what he would like prefetch by using an optional call back in the `withReferences` method.
```dart
withReferences((o)=> o(prefetchUsers: true)).get()
```
This `o` function is a field on the model.
Running the `o` function, will return a new function.
What does this function do? What does it return?

Well, the return value from `o` is itself an async function.
```dart
o(prefetchUsers: true) // Future<Prefetches> Function()
await o(prefetchUsers: true)() // Prefetches
```
Calling it will return a `Prefetches` object.

### Step 2 -  Prefetches
Well, what is a prefetches?
Relax, I'm getting to it.
Each table has a prefetches object associated with it. (yes, even more code 😟)
This holds the prefetched data. So running `await o(prefetchUsers: true)()` will return a class `GroupsPrefetches` which holds all the users for the group.

### Step 3 - OK, When are we running this?
Well, we don't run `await o(prefetchUsers: true)()` just yet. When we run `withReferences((o)=> o(prefetchUsers: true))`, we don't get the prefetches yet.
1. The race conditions could be even worse
2. `withReferences` would need to be async.

So when do we run them?
We run them when we actually do the `get()`

So calling `withReferences(...)` saves the entire callback (`(o)=>o(prefetchUsers: true)`) for later. This is stored on another field of the model.

### OK, I want my references now
Sure, when you finally run `get()`, we will run `await o(prefetchUsers: true)()` we will get the prefetches for this class.
We will then insert this prefetches object into every `<TableName>WithReferences` object.

So when you finally run
```dart
final (group,refs)  = await groups.withReferences((o)=> o(prefetchUsers: true)).getSingle()
final users = await refs.users.get();
```
The refs object will insert the appropriate users into the cache of the newly created table manager (remember `users` is a table manager).
Running `.get()` or `.watch` will return from this cache, instead of making a query.


If you don't understand this, it's ok. It will make sense once you see the code (I hope)

PR #3086, opened 2024-07-10
References with Prefetching

dickermoshe:
After hundreds of man--hours finding the most optimal way to add support for prefetching, I've come up with this.

Let me know if you need more info. This is a massive PR, and I'm happy to explain any part of it.

### References

This function with return a new manager which will return each item in the database with its references

The references are returned as a prefiltered manager, which will only return the items which are related to the item

For example:
```dart
for (final (group,refs) in await groups.withReferences().get()) {
  final usersInGroup = await refs.users.get();
  /// Is identical to:
  final usersInGroup = await users.filter((f) => f.group.id(group.id)).get();
}
 ```
### Prefetching

The keen among you may notice that the above code is extremly inefficient, as it will run a query for each group to get the users in that group.
This could mean hundreds of queries for a single page of data, grinding your application to a halt.

The solution to this is to use prefetching, which will run a single query to get all the data you need.

For example:
```dart
for (final (group,refs) in await groups.withReferences((prefetch) => prefetch(users: true)).get()) {
   final usersInGroup = refs.users.prefetchedData;
}
```

Note that `prefetchedData` will be null if the reference was not prefetched.

### Symmetrical Type Converters
This PR fixes a bug where referenced filters which didn't share the exact same type would not act correctly.
The Manager API now ignores references whose type converters arrent the same. The Tests were updated to support this.

### Internals

The is much going on, if we go through this step by step, it should come clear

#### References

When `withReferences`, we create a new manager which has it's `ActiveDataclass` generic updated to the `DataclassWithReferences` generic. The ActiveDataclass is the class which is returned by any of the `get`/`watch` functions.
When any of these `get`/`watch` method are called, we apply the `_withReferenceMapper` if the  `ActiveDataclass` is a `DataclassWithReferences`.

What is returned from a withReferences call is a record, with the item, along with an object which has getters for prefiltered managers.

```dart
for (final (group,refs) in await groups.withReferences().get()) {
    group; // Original Item
    refs; // Prefiltered Managers
    final usersInGroupManager = refs.users  // Prefiltered Manager
    final usersInGroupManager = refs.users  // Prefiltered Manager
}
```

#### Prefetching

When `withReferences` is called with a prefetcher, we create a class called `PrefetchHooks` and add it to the manager. This class will be different depening on what was prefetched.

```dart
groups.withReferences((prefetch) => prefetch(users: false, admin: true))
```
Would create a `PrefetchHooks` which was configured to only prefetch `admin`.

This `PrefetchHooks` class has 2 jobs:
 1. Add joins to the query before it is executed
 2. Execute a 2nd query to get reverse references and store it's result in each `TypedResult` object.

When a referenced manager is created that has prefetch data in it's `TypedResult` object, it will be loaded into the `prefetchedData` field of the manager.

```dart
$$TodosTableTableProcessedTableManager get todos {
    final manager = $$TodosTableTableTableManager($_db, $_db.todosTable)
        .filter((f) => f.category.id($_item.id));

    final cache = $_typedResult.readTableOrNull(_todosTable($_db)); // Loads the prefetched data from the TypedResult
    return ProcessedTableManager(
        manager.$state.copyWith(prefetchedData: cache));
  }
```

#### Hacks
Loading in multiple items into a single `TypedResult` object wasn't currently supported.
Reading a table from a `TypedResult` expects a single item, so we had to hack around this by creating a new `MultiTypedResultKey`. This item is only meant to act as a key for the `TypedResult` object, and is never actually used to read the data. Although, nothing bad would happen if it was. 

####  More work:

- [x] Docs
- [ ] Tests

### DIFF
--- a/docs/lib/snippets/dart_api/manager.dart
+++ b/docs/lib/snippets/dart_api/manager.dart
@@ -1,4 +1,4 @@
-// ignore_for_file: invalid_use_of_internal_member
+// ignore_for_file: invalid_use_of_internal_member, unused_local_variable
 
 import 'package:drift/drift.dart';
 
@@ -209,6 +209,66 @@ extension ManagerExamples on AppDatabase {
     );
   }
 // #enddocregion manager_filter_custom_back_references
+
+// #docregion manager_references
+  Future references() async {
+    /// Get each todo, along with a its categories
+    final categoriesWithReferences =
+        await managers.todoItems.withReferences().get();
+    for (final (todo, refs) in categoriesWithReferences) {
+      final category = await refs.category?.getSingle();
+    }
+
+    /// This also works in the reverse
+    final todosWithRefs = await managers.todoCategory.withReferences().get();
+    for (final (category, refs) in todosWithRefs) {
+      final todos = await refs.todoItemsRefs.get();
+    }
+  }
+
+// #enddocregion manager_references
+// #docregion manager_prefetch_references
+  Future referencesPrefetch() async {
+    /// Get each todo, along with a its categories
+    final categoriesWithReferences = await managers.todoItems
+        .withReferences(
+          (prefetch) => prefetch(category: true),
+        )
+        .get();
+    for (final (todo, refs) in categoriesWithReferences) {
+      final category = refs.category?.prefetchedData?.firstOrNull;
+      // No longer needed
+      // final category = await refs.category?.getSingle();
+    }
+
+    /// This also works in the reverse
+    final todosWithRefs = await managers.todoCategory
+        .withReferences((prefetch) => prefetch(todoItemsRefs: true))
+        .get();
+    for (final (category, refs) in todosWithRefs) {
+      final todos = refs.todoItemsRefs.prefetchedData;
+      // No longer needed
+      //final todos = await refs.todoItemsRefs.get();
+    }
+  }
+// #enddocregion manager_prefetch_references
+
+  Future referencesPrefetchStream() async {
+// #docregion manager_prefetch_references_stream
+    /// Get each todo, along with a its categories
+    managers.todoItems
+        .withReferences((prefetch) => prefetch(category: true))
+        .watch()
+        .listen(
+      (todoWithRefs) {
+        for (final (todo, refs) in todoWithRefs) {
+          // Updates to this category won't trigger an update
+          final category = refs.category?.prefetchedData?.firstOrNull;
+        }
+      },
+    );
+// #enddocregion manager_prefetch_references_stream
+  }
 }
 
 // #docregion manager_filter_extensions
--- a/docs/lib/snippets/dart_api/manager.dart
+++ b/docs/lib/snippets/dart_api/manager.dart
@@ -16,10 +16,10 @@ class TodoItems extends Table {
 class TodoCategory extends Table {
   IntColumn get id => integer().autoIncrement()();
   TextColumn get description => text()();
+  IntColumn get user => integer().nullable().references(Users, #id)();
 }
 
 // #docregion user_group_tables
-
 class Users extends Table {
   IntColumn get id => integer().autoIncrement()();
   TextColumn get name => text()();
@@ -256,14 +256,17 @@ extension ManagerExamples on AppDatabase {
   Future referencesPrefetchStream() async {
 // #docregion manager_prefetch_references_stream
     /// Get each todo, along with a its categories
-    managers.todoItems
-        .withReferences((prefetch) => prefetch(category: true))
+    managers.todoCategory
+        .withReferences((prefetch) => prefetch(todoItemsRefs: true, user: true))
         .watch()
         .listen(
-      (todoWithRefs) {
-        for (final (todo, refs) in todoWithRefs) {
-          // Updates to this category won't trigger an update
-          final category = refs.category?.prefetchedData?.firstOrNull;
+      (catWithRefs) {
+        for (final (cat, refs) in catWithRefs) {
+          // Updates to the user table will trigger a query
+          final users = refs.user?.prefetchedData;
+
+          // However, updates to the TodoItems table will not trigger a query
+          final todos = refs.todoItemsRefs.prefetchedData;
         }
       },
     );
--- a/docs/lib/snippets/dart_api/manager.dart
+++ b/docs/lib/snippets/dart_api/manager.dart
@@ -130,7 +130,7 @@ extension ManagerExamples on AppDatabase {
   // #enddocregion manager_filter
 
   // #docregion manager_type_specific_filter
-  Future filterWithType() async {
+  Future<void> filterWithType() async {
     // Filter all items created since 7 days ago
     managers.todoItems.filter(
         (f) => f.createdAt.isAfter(DateTime.now().subtract(Duration(days: 7))));
@@ -141,7 +141,7 @@ extension ManagerExamples on AppDatabase {
 // #enddocregion manager_type_specific_filter
 
 // #docregion manager_ordering
-  Future orderWithType() async {
+  Future<void> orderWithType() async {
     // Order all items by their creation date in ascending order
     managers.todoItems.orderBy((o) => o.createdAt.asc());
 
@@ -151,7 +151,7 @@ extension ManagerExamples on AppDatabase {
 // #enddocregion manager_ordering
 
 // #docregion manager_count
-  Future count() async {
+  Future<void> count() async {
     // Count all items
     await managers.todoItems.count();
 
@@ -161,7 +161,7 @@ extension ManagerExamples on AppDatabase {
 // #enddocregion manager_count
 
 // #docregion manager_exists
-  Future exists() async {
+  Future<void> exists() async {
     // Check if any items exist
     await managers.todoItems.exists();
 
@@ -171,7 +171,7 @@ extension ManagerExamples on AppDatabase {
 // #enddocregion manager_exists
 
 // #docregion manager_filter_forward_references
-  Future relationalFilter() async {
+  Future<void> relationalFilter() async {
     // Get all items with a category description of "School"
     managers.todoItems.filter((f) => f.category.description("School"));
 
@@ -186,7 +186,7 @@ extension ManagerExamples on AppDatabase {
 // #enddocregion manager_filter_forward_references
 
 // #docregion manager_filter_back_references
-  Future reverseRelationalFilter() async {
+  Future<void> reverseRelationalFilter() async {
     // Get the category that has a todo item with an id of 1
     managers.todoCategory.filter((f) => f.todoItemsRefs((f) => f.id(1)));
 
@@ -199,7 +199,7 @@ extension ManagerExamples on AppDatabase {
 // #enddocregion manager_filter_back_references
 
 // #docregion manager_filter_custom_back_references
-  Future reverseNamedRelationalFilter() async {
+  Future<void> reverseNamedRelationalFilter() async {
     // Get all users who are administrators of a group with a name containing "Business"
     // or who own a group with an id of 1, 2, 4, or 5
     managers.users.filter(
@@ -211,24 +211,24 @@ extension ManagerExamples on AppDatabase {
 // #enddocregion manager_filter_custom_back_references
 
 // #docregion manager_references
-  Future references() async {
+  Future<void> references() async {
     /// Get each todo, along with a its categories
-    final categoriesWithReferences =
-        await managers.todoItems.withReferences().get();
-    for (final (todo, refs) in categoriesWithReferences) {
+    final todosWithRefs = await managers.todoItems.withReferences().get();
+    for (final (todo, refs) in todosWithRefs) {
       final category = await refs.category?.getSingle();
     }
 
     /// This also works in the reverse
-    final todosWithRefs = await managers.todoCategory.withReferences().get();
-    for (final (category, refs) in todosWithRefs) {
+    final categoriesWithRefs =
+        await managers.todoCategory.withReferences().get();
+    for (final (category, refs) in categoriesWithRefs) {
       final todos = await refs.todoItemsRefs.get();
     }
   }
 
 // #enddocregion manager_references
 // #docregion manager_prefetch_references
-  Future referencesPrefetch() async {
+  Future<void> referencesPrefetch() async {
     /// Get each todo, along with a its categories
     final categoriesWithReferences = await managers.todoItems
         .withReferences(
@@ -253,7 +253,7 @@ extension ManagerExamples on AppDatabase {
   }
 // #enddocregion manager_prefetch_references
 
-  Future referencesPrefetchStream() async {
+  Future<void> referencesPrefetchStream() async {
 // #docregion manager_prefetch_references_stream
     /// Get each todo, along with a its categories
     managers.todoCategory
@@ -287,7 +287,7 @@ extension After2000Filter on ColumnFilters<DateTime> {
       $composableFilter(column.unixepoch.equals(value));
 }
 
-Future filterWithExtension(AppDatabase db) async {
+Future<void> filterWithExtension(AppDatabase db) async {
   // Use the custom filters on any column that is of type DateTime
   db.managers.todoItems.filter((f) => f.createdAt.after2000orBefore1900());
 
@@ -302,7 +302,7 @@ extension After2000Ordering on ColumnOrderings<DateTime> {
   ComposableOrdering byUnixEpoch() => ColumnOrderings(column.unixepoch).asc();
 }
 
-Future orderingWithExtension(AppDatabase db) async {
+Future<void> orderingWithExtension(AppDatabase db) async {
   // Use the custom orderings on any column that is of type DateTime
   db.managers.todoItems.orderBy((f) => f.createdAt.byUnixEpoch());
 }
@@ -315,7 +315,7 @@ extension NoContentOrBefore2000FilterX on $$TodoItemsTableFilterComposer {
       (content.isNull() | createdAt.isBefore(DateTime(2000)));
 }
 
-Future customFilter(AppDatabase db) async {
+Future<void> customFilter(AppDatabase db) async {
   // Use the custom filter on the `TodoItems` table
   db.managers.todoItems.filter((f) => f.noContentOrBefore2000());
 }
@@ -327,7 +327,7 @@ extension ContentThenCreationDataX on $$TodoItemsTableOrderingComposer {
   ComposableOrdering contentThenCreatedAt() => content.asc() & createdAt.asc();
 }
 
-Future customOrdering(AppDatabase db) async {
+Future<void> customOrdering(AppDatabase db) async {
   // Use the custom ordering on the `TodoItems` table
   db.managers.todoItems.orderBy((f) => f.contentThenCreatedAt());
 }
--- a/docs/pages/docs/Dart API/manager.md
+++ b/docs/pages/docs/Dart API/manager.md
@@ -49,9 +49,12 @@ This will return a record with the entity and a `refs` object which contains the
 The problem with the above approach is that it will issue a separate query for each row in the result set. This can be very inefficient if you have a large number of rows. If there were 1000 todos, this would issue 1000 queries to fetch the category for each todo.
 
 {% block "blocks/alert" title="Watching with Prefetches" color="info" %}
-Prefetching references works when streaming data, however updates wont be triggered when the referenced data changes.
+If you are using `watch` you can use the `withPrefetches` method to prefetch references in a single query.
+However changes to the referenced table will not trigger a re-query of the parent table.
 
 {% include "blocks/snippet" snippets = snippets name = 'manager_prefetch_references_stream' %}
+
+This is because that when single references (e.g a category has a single user) are prefetched, they are included in the same query as the parent table. Therefore, when the user table changes, the query will be re-run and the parent table will be updated. However, when multiple references are prefetched, they are included in a separate query. Therefore, changes to the referenced table will not trigger a re-query of the parent table.
 
 {% endblock %}
 
--- a/docs/pages/docs/Dart API/manager.md
+++ b/docs/pages/docs/Dart API/manager.md
@@ -46,17 +46,8 @@ This will return a record with the entity and a `refs` object which contains the
 
 {% include "blocks/snippet" snippets = snippets name = 'manager_references' %}
 
-The problem with the above approach is that it will issue a separate query for each row in the result set. This can be very inefficient if you have a large number of rows. If there were 1000 todos, this would issue 1000 queries to fetch the category for each todo.
-
-{% block "blocks/alert" title="Watching with Prefetches" color="info" %}
-If you are using `watch` you can use the `withPrefetches` method to prefetch references in a single query.
-However changes to the referenced table will not trigger a re-query of the parent table.
-
-{% include "blocks/snippet" snippets = snippets name = 'manager_prefetch_references_stream' %}
-
-This is because that when single references (e.g a category has a single user) are prefetched, they are included in the same query as the parent table. Therefore, when the user table changes, the query will be re-run and the parent table will be updated. However, when multiple references are prefetched, they are included in a separate query. Therefore, changes to the referenced table will not trigger a re-query of the parent table.
-
-{% endblock %}
+The problem with the above approach is that it will issue a separate query for each row in the result set. This can be very inefficient if you have a large number of rows.  
+If there were 1000 todos, this would issue 1000 queries to fetch the category for each todo.
 
 #### Prefetching references
 
--- a/docs/pages/docs/Dart API/manager.md
+++ b/docs/pages/docs/Dart API/manager.md
@@ -39,6 +39,28 @@ Type specific filters for `int`, `double`, `Int64`, `DateTime` and `String` are 
 {% include "blocks/snippet" snippets = snippets name = 'manager_type_specific_filter' %}
 
 
+### Referencing other tables
+
+The manager also makes it easy to query an entities referenced fields by using the `withReferences` method.
+This will return a record with the entity and a `refs` object which contains the referenced fields.
+
+{% include "blocks/snippet" snippets = snippets name = 'manager_references' %}
+
+The problem with the above approach is that it will issue a separate query for each row in the result set. This can be very inefficient if you have a large number of rows. If there were 1000 todos, this would issue 1000 queries to fetch the category for each todo.
+
+{% block "blocks/alert" title="Watching with Prefetches" color="info" %}
+Prefetching references works when streaming data, however updates wont be triggered when the referenced data changes.
+
+{% include "blocks/snippet" snippets = snippets name = 'manager_prefetch_references_stream' %}
+
+{% endblock %}
+
+#### Prefetching references
+
+Drift provides a way to prefetch references in a single query to avoid inefficient queries. This is done by using the callback in the `withReferences` method. The referenced item will then be available in the referenced managers `prefetchedData` field.
+
+{% include "blocks/snippet" snippets = snippets name = 'manager_prefetch_references' %}
+
 ### Filtering across tables
 You can filter across references to other tables by using the generated reference filters. You can nest these as deep as you'd like and the manager will take care of adding the aliased joins behind the scenes.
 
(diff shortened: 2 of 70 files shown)


dickermoshe:
@simolus3 Ready for review!

dickermoshe:
Let me know if I should break up this PR into parts to make it digestible.

drift/lib/src/runtime/manager/references.dart
```diff
@@ -0,0 +1,224 @@
+part of 'manager.dart';
+
+/// A simple function for generating aliases for referenced columns
+///
+/// This function is used internally by generated code and should not be used directly.
+// ignore: non_constant_identifier_names
+String $_aliasNameGenerator(
```
simolus3, line 7: Is this actually used outside of the library? The only use in generated code is to `MultiTypedResultKey`, so couldn't that constructor also take the involved columns and call this private method internally?

drift/lib/src/runtime/manager/references.dart
```diff
@@ -0,0 +1,224 @@
...
+/// final (group,refs) = await groups.filter((f) => f.id(5)).withReferences().getSingle()
+/// final usersInGroup = refs.users.get() // filter((f) => f.group(group.id)) is already applied
+/// ```
+/// is short for:
+/// ```dart
+/// final group = await groups.filter((f) => f.id(5)).getSingle()
+/// final usersInGroup = await users.filter((f) => f.group(group.id)).get()
+/// ```
+/// {@macro manager_internal_use_only}
+class BaseReferences<$Database extends GeneratedDatabase, $Table extends Table,
```
simolus3: Let's make this `base` and generated reference classes `final`.

drift/lib/src/runtime/manager/references.dart
```diff
@@ -0,0 +1,224 @@
...
+/// Type definition for a function that transforms the state of a manager
+typedef StateTransformer = T Function<
+    T extends TableManagerState<dynamic, dynamic, dynamic, dynamic, dynamic,
+        dynamic, dynamic, dynamic, dynamic, dynamic>>(T $state);
+
+T _defaultStateTransformer<
+    T extends TableManagerState<dynamic, dynamic, dynamic, dynamic, dynamic,
+        dynamic, dynamic, dynamic, dynamic, dynamic>>(T $state) {
+  return $state;
+}
```
simolus3, line 288: It might be too late for that, and in this case I think it's fine, but since managers are mostly internal and short-lived builders, I think it would be much easier to just make them mutable.

We need all this ceremony with the type parameters just to indicate that this function turns specific manager instances into the something of the same type - with mutable managers it could just be `void Function(TableManagerState)`.

drift/lib/src/runtime/manager/manager.dart
```diff
@@ -505,12 +754,35 @@ class ProcessedTableManager<
...
+        $UpdateCompanionCallback,
+        $DataclassWithReferences,
+        $ActiveDataclass,
+        $CreatePrefetchHooksCallback> {
   /// Create a new [ProcessedTableManager] instance
   @internal
   ProcessedTableManager(super.$state);
+
+  /// The prefetched data for this manager
+  /// This will be null if the manager was not created with prefetched data
```
simolus3: Maybe it's simpler and more precise to say "Prefetched data, if references with prefetching enabled were added to this manager".

drift/lib/src/runtime/manager/manager.dart
```diff
@@ -463,10 +705,19 @@ abstract class BaseTableManager<
...
+    final iterator = list.iterator;
+
+    if (!iterator.moveNext()) {
+      return null;
+    }
+    final element = iterator.current;
+    if (iterator.moveNext()) {
+      throw StateError('Expected exactly one result, but found more than one!');
+    }
+    return element;
```
simolus3: ~I think we should mention the changed behavior of throwing if there are multiple rows~ (nevermind, updated to be consistent with the rest of drift). Instead of using the iterator manually we can also do

```dart
if (list.isNotEmpty) {
  return list.single;
} else {
  return null;
}
```

drift/lib/src/runtime/manager/manager.dart
```diff
@@ -69,26 +87,88 @@ class TableManagerState<
...
+  List<$ActiveDataclass> toActiveDataclass(List<TypedResult> items) {
+    if ($DataclassWithReferences == $ActiveDataclass) {
+      return _withReferenceMapper(items) as List<$ActiveDataclass>;
+    } else {
+      return items.map((e) => e.readTable(_tableAsTableInfo)).toList()
+          as List<$ActiveDataclass>;
+    }
+  }
+
+  /// If this field has references, this field will contain the function that the user will use to create a [PrefetchHooks]
```
simolus3: "If this field" should be "If this table" here right?

simolus3 (commented):
Thanks! I've had a quick look at the changes in `drift` and some of the changes to generated code, but not at the implementation in `drift_dev` yet.

I'll need to take a deeper look to fully understand everything. This is fine as a single PR, but having a complete dartdoc on some of the classes or fields giving an overview of how fetching references works end-to-end would be useful.
It should also mention what role the generated code plays (e.g. how the prefetch hook callback looks like with different kinds of relationships between tables). Then the transformations contained in a `PrefetchHook`, how they are invoked when building the query. And also how the joins are used (having a concrete example in the docs like "prefetching here generates this sql statement" would be super helpful) and how the result is grouped into lists.
This is a complex feature (not that it's easy to simplify, this necessarily has to bridge between lots of runtime APIs, query generation and generated code to hold the structure), but it should be obvious to understand how everything is set up in code. I think a central comment somewhere explaining the main idea and then linking to relevant classes and methods would help a lot. 

For tests it would also be good to use the mocked executor to validate the generated SQL just so we can be sure what it looks like.

drift/lib/src/runtime/manager/references.dart
dickermoshe: It's used on a static field, so it can't be a private value on the class

```dart
class BaseFoo {
  func();
}

class Foo extends BaseFoo {
  static const name = func(); // Doesnt work
}
```

dickermoshe:
Good point. I'll add samples of generated code to the doccoments.
You can pause review until I add them. It's pretty confusing without any context

drift/lib/src/runtime/manager/references.dart
dickermoshe: I agree, but IMHO it's a worthy tradeoff.

There are some places where it would be simplified, but this manager api is quite complex, and I don't trust myself enough not to have bugs.

drift/lib/src/runtime/manager/manager.dart
dickermoshe: I copy pasted this from the original implementation.

Should I update that too?

dickermoshe:
Good point on a centralized comment explaining everything.
Il put it on BaseReferences

dickermoshe:
Take a look at the doccomment

drift/lib/src/runtime/manager/manager.dart
simolus3: Hm it looks like we have multiple copies of that around then, I've checked `singleElementsOrNull` which uses the `isNotEmpty` check. It might be worth extracting them all into a common method then.

drift/lib/src/runtime/manager/references.dart
```diff
@@ -0,0 +1,491 @@
...
+///
+/// ### Solution
+///
+/// There are two methods we use together to solve this.
+/// Joins and Prefetches.
+///
+/// ##### Join
+///
+/// For fields which are foreign keys to other tables, we can just use a join to select the related field and get it all in one query
+/// There are many examples how to do this in the drift documentation
```
simolus3, line 52: Perhaps obvious, but mention that the generated code + manager runtime now adds the joins automatically. That's something not done in other places of drift.

drift/lib/src/runtime/manager/references.dart
```diff
@@ -0,0 +1,491 @@
...
+/// }
+/// ```
+/// The issue with the above code is that we could be setting ourselves up to do tons of queries.
+///
+/// ### Solution
+///
+/// There are two methods we use together to solve this.
+/// Joins and Prefetches.
+///
+/// ##### Join
```
simolus3: This and Prefetch should be h4 instead of h5?

drift/lib/src/runtime/manager/references.dart
```diff
@@ -0,0 +1,491 @@
...
+/// ```
+///
+/// We get all the users ahead of time, and then manually return each group with it's users using a `where(...)` filter.
+///
+/// We will call both of these "prefetch" throughout.
+///
+///
+/// ### Manager API
+///
+/// This is quite verbose, and the manager api seeks to solve this for you.
```
simolus3: This section is aimed at contributors, so

```suggestion
/// This is quite verbose, and the manager api seeks to solve this for drift users.
```

drift/lib/src/runtime/manager/references.dart
```diff
@@ -0,0 +1,491 @@
...
+///
+/// ### Implementation
+///
+/// We only need to do 2 things for this to work:
+///
+/// 1) Reference class
+///
+/// This class is kinda cool. It holds the original result of the query as a `TypedResult`, which is what drift uses to store the raw query.
+/// But along with that, it has getters for prebuilt managers.
+/// When `withReferences` is called, we return the dataclass, together with this.
```
simolus3, line 102: Perhaps also mention that this is what the three type arguments on the manager classes are for.

drift/lib/src/runtime/manager/references.dart
```diff
@@ -0,0 +1,491 @@
...
+///
+/// To keep things simple for now, this is a reference class which has it's prefetching abilities removed
+/// ```dart
+/// final class $$ProductTableReferences
```
simolus3, line 106: I think it would be helpful to show a consistent example throughout this explainer - so probably the users / groups relation?

simolus3: Or updating the initial examples to use the products table might be easier than updating the larger examples at the end :D

drift/lib/src/runtime/manager/references.dart
```diff
@@ -0,0 +1,491 @@
...
+/// final (group,refs) = await groups.filter((f) => f.id(5)).withReferences().getSingle()
+/// final usersInGroup = refs.users.get() // filter((f) => f.group(group.id)) is already applied
+/// ```
+/// is short for:
+/// ```dart
+/// final group = await groups.filter((f) => f.id(5)).getSingle()
+/// final usersInGroup = await users.filter((f) => f.group(group.id)).get()
+/// ```
+///
+/// {@macro manager_internal_use_only}
```
simolus3, line 27: We should either remove this macro invocation or the one at the very end

simolus3 (commented):
That documentation is really helpful, thanks! It makes the implementation choices much clearer. I've left some minor comments on it, but I should be able to review the reminder of this tomorrow :)

dickermoshe:
### Problem: Streaming + Prefetching
When we stream a table together with reverse references, we aren't using a join. We are doing a 2nd query and doing the JOIN in dart instead of SQL.

However this 2nd query is just a plain `get()` query, meaning that changes to that 2nd table won't trigger a new event.

```dart
group.withReferences((prefetch)=>prefetch(users: true)).watch().listen((groupWithRefs){
  final (group,refs) = groupWithRefs;
  final users =  refs.users.prefetchedData;
})
```
Changes to `group` table will trigger a new event, but changes to the `user` table won't.

3 Options:

1) Leave as is. Put a notice in the docs "Changes to Reverse References won't trigger a new event"
  Con: Massive Foot Gun
  Pro: Easy
2) Inject this table into the query, so that the ENTIRE query gets ran again when the users table is edited.
  Pro: Easy
  Con: Inefficient, If there we no changes to the groups table, we should not rerun that query, we should only rerun the 2nd query and join the data in dart again.
3) Do more magic to make it work. (asyncExpand & combineStream)
  Con: It's hard, but I will take some time to experiment
  Pro: It's perfect
  



Right now the docs are poorly written for option `#1`.

dickermoshe:
Maybe we could make a version of the watch function which allows us to pass in additional tables to watch. (to trigger new event when a reverse references table has an updates [which is not prefetched using a join, but another query])
Won't be very efficient, because it will rerun all the queries, even though it only needs to re-run the 2nd one

dickermoshe:
Either way my docs aren't clear. I'm need a fresh set of eyes on these new streaming docs

dickermoshe:
@simolus3 

Update:

Adding support for streams is not as complex as I thought.
I won't have time to work on it today, but should be done over the next couple of days.
Doesn't add any significant complexity.

Hold off on merging

simolus3:
Yeah I'm in favor of option 3 of course :) We're recommending [something similar](https://drift.simonbinder.eu/docs/examples/relationships/#selecting-a-cart) for a current approach shared on the documentation website.

drift/lib/src/runtime/manager/references.dart
```diff
@@ -0,0 +1,492 @@
...
+    required Iterable<$ReferencedDataclass> Function(
+            $CurrentDataclass item, List<$ReferencedDataclass> referencedItems)
+        referencedItemsForCurrentItem}) async {
+  if (!doPrefetch || typedResults.isEmpty) {
+    return typedResults;
+  } else {
+    final managers = typedResults.map(managerFromTypedResult);
+    // Combine all the referenced managers into 1 large query which will return all the
+    // referenced items in one go.
+    final manager = managers.reduce((value, element) => value._filter(
```
simolus3, line 506: Does this work if the referenced items are from multiple different tables or would `typedResultsWithPrefetched` be called multiple times in that case?

drift/lib/src/runtime/manager/references.dart
```diff
@@ -0,0 +1,492 @@
...
+          {required String aliasName}) {
+    return MultiTypedResultKey<$Table, List<$Dataclass>>._(
+      $columns: table.$columns,
+      asDslTable: table.asDslTable,
+      attachedDatabase: table.attachedDatabase,
+      columnsByName: table.columnsByName,
+      entityName: aliasName,
+
+      /// This table is used a key for a map and should never be used to read data
+      /// However, in case it is, we will map a single item to a list of items
```
simolus3: Supporting things we don't want to support doesn't seem right to me. Either we officially support this, or we override the relevant methods to throw an `UnsupportedError`.

drift_dev/lib/src/writer/manager/table_manager_writer.dart
```diff
@@ -93,6 +85,29 @@ class _TableManagerWriter {
       return true;
     }).toList();
 
+    // Remove any relations where the type isnt exactly the same (num and int)
+    // This is caused by using different type converters
```
simolus3, line 89: Also mention why we're doing this (we'd generate invalid code because we're comparing loaded fields from data classes in the manager API, right?)

drift/test/manager/manager_prefetch_test.dart
```diff
@@ -0,0 +1,331 @@
...
+        Value<RowId> category,
+        String content,
+        Value<TodoStatus> status,
+        Value<DateTime> targetDate,
+        Value<String> title
+      })> todoData;
+
+  setUp(() async {
+    db = TodoDb(testInMemoryDatabase());
+    stores = [
```
simolus3: Can we share this data between tests to avoid all this lengthy setup stuff here? Having too many rows can also make the tests harder to understand because it makes it non-trivial to figure out what the right response should be.

drift/test/manager/manager_referenced_filter_test.dart
```diff
@@ -516,4 +274,478 @@ void main() {
...
+    expect(
+        db.managers.product
+            .withReferences()
+            .get(distinct: true)
+            .then((value) =>
+                value.first.$2.department?.getSingle() ?? Future.value(null))
+            .then(
+              (value) => value?.id,
+            ),
+        completion(departments[0].id));
```
simolus3: I know we're also doing this for the other tests, but if we don't need to test concurrent invocations explicitly, `expect(await future, assertion)` is generally more readable than `expect(future, completion(assertion))`.

simolus3 (commented):
Had another look - I agree with the approach and will wait to see the implementation for streams (for which we should also have tests then :)).

drift/lib/src/runtime/manager/references.dart
dickermoshe: It's called once per prefetch

dickermoshe:
Also need to work on doing everything in a single transaction to for race conditions.

It may seem like I am remembering new things last minute, but I'm just going through my TODOs, which have all of this laid out.

dickermoshe:
@simolus3 Can I add rxdart to dependencies?

dickermoshe:
@simolus3 Things got slightly more complex, but I think it's still pretty clear

Lmk if anything isn't clear and i'll update the doccomment.

I've got no idea on how to use a transaction with a stream, so I'm going to need your help on that...

dickermoshe:
The reference tests need a complete revamp. That's the next step once you are happy with how the rest of it looks

drift/lib/src/runtime/manager/references.dart
```diff
@@ -288,25 +285,70 @@ T _defaultStateTransformer<
...
+  Stream<List<TypedResult>> _addPrefetchedDataToStream(
+      Stream<List<TypedResult>> itemStream,
+      {required bool watch}) {
+    /// If this table contains no reverse references, we can just return the stream as is.
+    if (prefetchedDataStreamsCallback == null) {
+      return itemStream;
+    }
+
+    /// Otherwise we add the prefetched data to the stream
+    return itemStream.asyncMap((rows) {
```
simolus3: This is probably super tricky, but at the moment the selects for the streams are not running in a transaction right?

I wonder if we should just register a stream depending on all the tables (including possible prefetches) and then run that in a transaction at once instead of combining independent query streams here?
Not being consistent with `get` in that we don't use a transaction here is not great, but we also don't have the fundamentals in place to get this right at the moment. I'll take a look at adding those first, it's just something remember.

simolus3:
Apart from relatively minor things (not in these sense that we shouldn't fix them, just that they won't impact the overall implementation much), this looks good to me. So I agree with the API, design and the general implementation.

dickermoshe:
The problem with marking the original query to watch all referenced tables
is that ALL the queries will run again if anything changes.
Now only the reference table query is ran again.

Do you think you’ll be able to find a way to get transactions on streams
working?

drift/lib/src/runtime/manager/references.dart
dickermoshe: I don't have the slightest idea on how to get this working with a transaction.

How would having the other operations as a `get` help?

dickermoshe:
I was going through the possibilities of what a race cond could be:

Example `groups.withReferences((prefetch)=> prefetch(users: true)).watch()`

1) A groups is added in between queries:
    Result: The group won't be in the final result, no error will occur.
2) A group is deleted in between queries,
    Result: No users will be returned for this now-non-existent group.

There will be a single tick where info is out of data, but no errors will occur

Adding tests for this won't be trivial.

dickermoshe:
@simolus3 I'm going to make this a bool on `prefetch(inTransaction: false)`
Complicates the code further, but I think this is valid solution to this issue.

dickermoshe:
@simolus3 If you're good with this I'm going to proceed

simolus3:
> Adding tests for this won't be trivial.

I think it could be possible with a mocked executor or when manually invalidating stream queries. Definitely not easy though, let me know if you need more ideas on that.

>  Result: The group won't be in the final result, no error will occur.

I'm still not sure I like that - stream queries should not emit stale data. If we know that data is going to be out-of-date while a query is still running, we should ignore the results and run the query again (which is also what we're doing for single queries). We should find a way to provide these guarantees for stream queries as well (e.g. by not emitting a completed snapshot if the upstream groups table has changed while we're processing an earlier change by re-running the users query).

> There will be a single tick where info is out of data, but no errors will occur

You left out the case where a group was updated, couldn't that emit an inconsistent snapshot if columns from the groups table are referenced in users? The principle of never emitting stale data should still apply though.

dickermoshe:
@simolus3 
Is this impossible to get right?
This is where my experience with Drifts internals and Streams in general is starting to show.
Awaiting your input on how to proceed...

simolus3:
I think the only fully sound solution is to run each fetch of the stream in a transaction. That can be inefficient, and we should be transparent about this caveat when mentioning streams with pre-fetches, but it's the only way I can come up with that provides consistency guarantees.
Even if we don't know about an update, or learn about an update that happened between the main and a prefetch query too late, we'd emit inconsistent rows.

In 58aa118fea323b9dd9b76c4f533af0d772c09fcc, I've explored a relatively simple change for this. You would create a `StreamQueryFetcher` that has a `null` key and `fetchData: get`. `readsFrom` must be all tables then. This makes sure that, if an update happens while we're fetching data, the (read-only) transaction is thrown away and we'd restart.

dickermoshe:
~~I'm confused. I was under the impression that we are already doing everything in a transaction:~~

~~I've added a private top level function called `_watchWithAdditionalTables` and am using a transaction for the referenced prefetches, I'm not sure why anything else needs to be added?~~

Seems like I missread the docs

dickermoshe:
Have we thought about the following?
1. Cache the previous stream result, so we can emit the old result if we detect something has changed (This is possible with signals, but I wonder if it can be done with streams)
2. How do we detect if something has changed? We have a UUID for each table, which updates when any mutating action is performed on it. We store this UUID on the db instance, along with each result each stream.

```dart
class db {
  final tableIds = {"table_name":"y37423483729..."}
}
/// The final combined stream which may contains inconsistent data
stream.listen((v){
  final typedResult, previousTypedResult ,tableUuid,tableName = v;
  if (db.tableIds[tableName] != tableUuid){
    return previousTypedResult;
  } else {
    return typedResult;
  }
});

/// Create a custom stream query executer which will return the UUID at the time of fetching
/// And that also caches the previous result
/// TODO

class UUIDIterceptor extends QueryInterceptor {
  //... Listen for mutating queries and update db.tableIds
}

```

We can then use an CombineStream, where we have the best performance.

The above implementation should work, however It doesn't look super neat.
@simolus3 , what are your thoughts?
I'm not super familiar with the current Stream Query code, could something like the above be shoehorned in?

simolus3:
> The above implementation should work, however It doesn't look super neat.

Wouldn't we still run the query then and only decide that we can cache the interpretation of results afterwards? 
I also think this would be pretty hacky and unreliable. We cannot be 100% accurate about when a stream must update (a second connection to the database could also be made outside of drift for example). But I strongly feel that the actual results emitted by a stream must always be a consistent snapshot, regardless of what other tools may also be accessing the database. And the only way to ensure that when multiple queries are involved is to use a transaction and invalidate everything when any involved query changes. By the way, I don't think this would be unacceptably slow either. We're doing the same for single queries with joins for instance, so I don't think users would be surprised that everything making up the query is re-run when an input table changes.

dickermoshe:
I was expecting such a response.
I completely agree.

What are our next steps to finish this PR (besides tests)?

simolus3:
> What are our next steps to finish this PR (besides tests)?

So implementation wise we still need to change the watch implementation to migrate from `asyncExpand` (which thus runs prefetch queries independently) to something actually ensuring each emit happens in a transaction. Given the small size you can probably cherry pick 58aa118fea323b9dd9b76c4f533af0d772c09fcc into this PR (it's not on `develop` yet since I wanted your feedback on whether that approach would work for you).

Given that the prefetch hooks already expose the list of additional tables, I think an implementation could simply be:

```dart
@override
Stream<List<$ActiveDataclass>> watch(
    {bool distinct = false, int? limit, int? offset}) {
  // Build the initial query once so that we know which tables it selects from
  // (todo if there's a better way to get the tables we should use it)
  var baseSelect =
      $state.prefetchHooks.withJoins($state).buildSelectStatement();
  var context = GenerationContext.fromDb($state.db);
  baseSelect.writeInto(context);

  return $state.db.createStream(QueryStreamFetcher<List<$ActiveDataclass>>(
    readsFrom: TableUpdateQuery.onAllTables([
      ...context.watchedTables,
      ...$state.prefetchHooks.explicitlyWatchedTables
    ]),
    fetchData: () async {
      return get(distinct: distinct, limit: limit, offset: offset);
    },
  ));
}
```

dickermoshe:
Thanks!
Now comes the fun part, TESTS!

dickermoshe:
@simolus3 
Hmmm, something isn't quite working with the new CustomQuery feature.
Just using it plainly is throwing an error I can't seem to nail down.
Would you be able to take a look at this please?

Pull the branch and run this test:
```bash
dart test ./test/manager/manager_prefetch_test.dart  --chain-stack-traces  --plain-name "watch"
```

dickermoshe:
~~Seems to be an issue with a couple of fields on `QueryStream` which are not updated to use the `Rows` generic.
I don't understand much of this code, so I don't feel confident fixing this...~~

I've pushed a fix, but please review it with extra care

drift/lib/src/runtime/cancellation_zone.dart
```diff
@@ -50,8 +50,7 @@ class CancellationToken<T> {
 
 /// Extensions that can be used on cancellable operations if they return a non-
 /// nullable value.
-extension NonNullableCancellationExtension<T extends Object>
-    on CancellationToken<T> {
+extension NonNullableCancellationExtension<T> on CancellationToken<T> {
```
simolus3: This should still have `extends Object` as bounds, otherwise the result of `resultOrNullIfCancelled` would be ambiguous.

The rest of this fix looks good to me, I've pushed a fix making everything non-null. I'll do a final review tomorrow and merge if it looks good.

docs/lib/snippets/dart_api/manager.dart
```diff
@@ -209,6 +209,69 @@ extension ManagerExamples on AppDatabase {
     );
   }
 // #enddocregion manager_filter_custom_back_references
+
+// #docregion manager_references
+  Future references() async {
```
simolus3: These should be `Future<void>`

docs/lib/snippets/dart_api/manager.dart
```diff
@@ -209,6 +209,69 @@ extension ManagerExamples on AppDatabase {
...
+    /// Get each todo, along with a its categories
+    final categoriesWithReferences =
```
simolus3: I think `todosWithReferences` or `todosWithRefs` may be a more appropriate name? And the one below should be `categoriesWithRefs` as that's the one selecting from categories as a main table?

docs/pages/docs/Dart API/manager.md
```diff
@@ -39,6 +39,22 @@ Type specific filters for `int`, `double`, `Int64`, `DateTime` and `String` are
 {% include "blocks/snippet" snippets = snippets name = 'manager_type_specific_filter' %}
 
 
+### Referencing other tables
```
simolus3, line 42: We don't need to necessarily do this in this PR, but I think we should consider [updating this example](https://drift.simonbinder.eu/docs/examples/relationships/) to show how the manager APIs can be used to resolve complex queries. I think this feature would make that a lot easier, so it's a good place to show case it.

drift/lib/src/runtime/query_builder/statements/select/select_with_join.dart
```diff
@@ -384,3 +393,11 @@ class _ResultStructure {
 
   _ResultStructure({required this.columnAliases, required this.queriedTables});
 }
+
+@internal
+extension JoinedSelectStatementAdditionalTablesX on JoinedSelectStatement {
```
simolus3: Drop the `X`,

```suggestion
extension JoinedSelectStatementAdditionalTables on JoinedSelectStatement {
```

simolus3 (approved):
LGTM apart from those small nits :+1:

docs/pages/docs/Dart API/manager.md
dickermoshe: IMHO the M2M docs should be featured more prominently, and that it should be shown there.
I'll add this to the TODOs for the docs rewrite

docs/lib/snippets/dart_api/manager.dart
dickermoshe: Good Point!

simolus3:
Thanks for the contribution!! :tada: :heart:

Issue #3097, opened 2024-07-18
Schema dump fails in flutter with drift_dev 2.19


fmorato:
**Describe the bug**

I'm just learning to use drift, and looking at how to dump schemas, but it's failing. 
Is this a bug, or how am I supposed to use it?

This is what I did

```
$ git clone https://github.com/simolus3/drift.git
$ cd drift/examples/app
```
<details><summary>dart pub get</summary>
<p>

```
$ dart pub get
Resolving dependencies... (1.2s)
Downloading packages... (4.1s)
+ _fe_analyzer_shared 67.0.0 (73.0.0 available)
+ analyzer 6.4.1 (6.8.0 available)
+ analyzer_plugin 0.11.3
+ archive 3.6.1
+ args 2.5.0
+ async 2.11.0
+ bazel_worker 1.1.1
+ boolean_selector 2.1.1
+ build 2.4.1
+ build_config 1.1.1
+ build_daemon 4.0.2
+ build_modules 5.0.8 (5.0.9 available)
+ build_resolvers 2.4.2
+ build_runner 2.4.11
+ build_runner_core 7.3.1
+ build_web_compilers 4.0.10 (4.0.11 available)
+ built_collection 5.1.1
+ built_value 8.9.2
+ characters 1.3.0
+ charcode 1.3.1
+ checked_yaml 2.0.3
+ cli_util 0.4.1
+ clock 1.1.1
+ code_builder 4.10.0
+ collection 1.18.0 (1.19.0 available)
+ convert 3.1.1
+ cross_file 0.3.4+1
+ crypto 3.0.3
+ dart_style 2.3.6
+ drift 2.19.1
+ drift_dev 2.19.0
+ drift_flutter 0.1.0
+ fake_async 1.3.1
+ ffi 2.1.2
+ file 7.0.0
+ file_picker 8.0.6
+ fixnum 1.1.0
+ flutter 0.0.0 from sdk flutter
+ flutter_colorpicker 1.1.0
+ flutter_lints 4.0.0
+ flutter_plugin_android_lifecycle 2.0.20
+ flutter_riverpod 2.5.1
+ flutter_test 0.0.0 from sdk flutter
+ flutter_web_plugins 0.0.0 from sdk flutter
+ frontend_server_client 4.0.0
+ glob 2.1.2
+ go_router 14.2.1
+ graphs 2.3.1
+ http_multi_server 3.2.1
+ http_parser 4.0.2 (4.1.0 available)
[...]

PR #3103, opened 2024-07-19
Test Code Generation with Flutter Version

dickermoshe:
Flutter is not always on the latest version of dart.
We wan't to test that the version of dart that flutter is using is works with drift_dev.
I've also changed the `sqlite3` constraints on the `drift_flutter` package to be compatible with older versions of dart.

I've also added testing for Flutter 3.19. 3.22 changed lots of stuff. IMHO I think we should support the 2 latest version of Flutter.

EDIT:
I've also set better lower version constraints for come packages.
All packages should still work if `pub downgrade` is ran.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -209,6 +209,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
+        # We want to support the two latest stable Flutter versions
         flutter: [3.19, 3.22]
 
     steps:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -228,6 +228,12 @@ jobs:
       - name: Generate code
         run: dart run build_runner build --delete-conflicting-outputs
         working-directory: examples/app
+      - name: Test CLI
+        run: |
+          dart run drift_dev schema dump lib/database/database.dart drift_schemas/
+          dart run drift_dev schema generate drift_schemas/ test/generated_migrations/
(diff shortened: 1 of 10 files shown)


.github/workflows/main.yml
```diff
@@ -207,11 +207,16 @@ jobs:
           dart test
   test_flutter:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        flutter: [3.19, 3.22]
```
simolus3, line 213: Add a comment explaining why these, and when they should be updated.

```suggestion
        # We want to support the two latest stable Flutter versions
        flutter: [3.19, 3.22]
```

simolus3 (commented):
Thanks! I agree that having these checks should make releases a lot less surprising and catch a bunch of Flutter incompatibilities during development.

Looks like we only need to change `drift_flutter` to have `^3.3.0` in its `environment.sdk` constraints?

Another thing we may want to test is the drift_dev CLI. Can we run `dart run drift_dev schema dump lib/database/database.dart drift_schemas/`, `dart run drift_dev schema generate drift_schemas/ test/generated_migrations/` and `dart run drift_dev schema steps drift_schemas/ lib/database/schema_versions.dart` in the example app before we run `flutter test`? That would serve as a regression test for https://github.com/simolus3/drift/issues/3097.

dickermoshe:
Chore:
Run `dart pub downgrade` and then pin the lower requirements until all the errors go away.

dickermoshe:
Ready for review

simolus3 (approved):
LGTM, thanks!

PR #3116, opened 2024-07-26
Migrate to Mkdocs

dickermoshe:
- Remove zap demo to simplify docs for now
- Remove dartdocs
- Replace docsy with mkdocs

I normaly develop on my own dickermoshe/drift fork, but I would like to be able to preview docs.
I'm hoping a PR from  simolus/drft will allow netlify to work properly


simolus3:
Just keeping track of the things that I've noticed - it's probably easier to track here than in the diff. Feel free to tick boxes that are done.
- [ ] Add pub.dev link
- [ ] The manager docs under "Queries" are really good. We should still document the manual way used for more complex queries though, and perhaps link those docs from the manager pages (e.g. `bulkCreate` could link to batches?)
- [x] The cookie consent popup needs to be removed. It's annoying, and we're not collecting any information that would require consent.
- [ ] Related to the first point, and I haven't checked it yet so it may be done already. We need to self-host fonts, CSS and scripts - the site shouldn't include any assets from third-party sites.
- [x] I prefer we remove the "switch to dark theme" button and just unconditionally mirror the OS settings (via `prefers-color-scheme: dark`, MkDocs is probably already doing this by default if we hide the button).
- [ ] [Cool URIs don't change](https://www.w3.org/Provider/Style/URI) - existing links should continue to work. Previously we have used `alias` for this to keep generating the page at locations it used to be at previously. We can also use Netlify redirects as an alternative, but it's important that existing links keep working.
~~- [ ] If possible, can we have the entire navigation stack (across tabs) shown on the left-side navbar?~~
- [ ] Because if we can, I think we may benefit from more tabs (e.g. "Setup" (basic stuff + manager), "Verified SQL", "Advanced features", "Examples").
~~- [ ] IMO the home tab can be removed since the drift name in the navbar is already linking to it.~~ This will be instead of the old splash screen.
- [x] Let's not use sticky navigation tabs.
- On the new tables page:
  - [x] I think "By default, Drift will name this reference after the table it references with Refs appended" is a little confusing without introducing the manager first? What if the structure was to first introduce references (as we have it right now), then show a simple manager query resolving the todo references, then use that to mention that references can be named for those queries?
  - [x] Maybe we want to move the references explanation to the columns page or onto a whole new page that is just very briefly mentioned + linked on both the columns/table pages? Especially with the features the manager API gives us around references, they deserve a good introduction that may be out of scope for the other pages (and I would expect both the columns and table pages to mention them a bit, both use them in snippets so there should be a short explanation).
  - [x] "By default, drift uses the snake_case name of the Dart getter in the database. ". Perhaps also link to the builder options page (and in particular `case_from_dart_to_sql`) and point that out as another way to override this globally.
  - [x] "Without them, you'd have to scan the whole database for each search, which is slow. Indexes make reads much faster, but slightly slow down adding new data" is not quite correct. Let's perhaps say "Without them, the database may have to scan the entire table for each lookup. Indexes store a bit of additional data to make lookups more efficient." (I wouldn't mention the insert thing, it's often negligible). To me, the "When to use indexes" thing also reads like it suggests premature optimizations. It should at least mention that indexes are most useful for tables expected to store a large amount of rows. Giving candidates that aren't specific to what queries a user might run is also not good advice. We should also mention that primary keys, unique keys and columns being referenced by another table are indexed automatically and don't need a manual index.
  I think most users won't need to use indexes often, so the suggestion should be more focused on the "large table + columns frequently used in a filter" part.
  - [x] "However, queries that filter on both columns can't use the index.", well, not as efficiently but the DBMS can still do two index lookups and then compute the intersection.
  - [ ] The index part should also mention unique indexes. That IMO also makes the two indexes vs. on index over two columns distinction more obvious.
- On the new columns page:
 - [x] The supported types section: "Additional types can be stored using type converters" should perhaps say "Other Dart types that can be converted into these primitive types can be stored using type converters." and also mention that "To support database engines with additional column types, custom types can be used - this feature is primarily aimed towards users porting drift to a different database system".
- On the new dataclass page:
 - [x] The "custom data classes" section should mention that it's possible to use records as custom data classes, either with a custom shape (`typedef User = ({int id, String name}); @UseRowClass(User) class Users extends Table`) or with a shape generated by drift `@UseRowClass(Record)`.
- On the new references page:
 - [ ] I think the warning about foreign keys not being enabled by default should be higher up.
- On the new dataclass page:
 - [ ] Don't mention `toCompanion` before introducing the companion section. Actually maybe don't mention it at all, data classes aren't supposed to be used for updates and inserts.
 - [ ] I think the json section should have a warning box saying that drift mainly offers json serialization for backwards compatibility and that it is not as customizable as proper json packages. Then point towards json serializable and built value + existing data classes as a better workaround.

dickermoshe:
> If possible, can we have the entire navigation stack (across tabs) shown on the left-side navbar?

Do you mean "Remove the Tab Bar and put everything on the left hand side?"

simolus3:
No, I was wondering if we can list every page on the left-side navbar regardless of which tab is selected. E.g. in this screenshot all the articles from the references section aren't shown because I'm in the examples tab. I wonder if we can show everything on every page to make it easier to discover, but it's not a big problem if MkDocs doesn't support that.

grafik

dickermoshe:
> No, I was wondering if we can list every page on the left-side navbar regardless of which tab is selected. E.g. in this screenshot all the articles from the references section aren't shown because I'm in the examples tab. I wonder if we can show everything on every page to make it easier to discover, but it's not a big problem if MkDocs doesn't support that.

MkDocs does not support that


> I prefer we remove the "switch to dark theme" button and just unconditionally mirror the OS settings (via prefers-color-scheme: dark, MkDocs is probably already doing this by default if we hide the button).

I'll can use css to hide the button, but there is no way to have an adaptive theme without the button being in the html

dickermoshe:
> [Cool URIs don't change](https://www.w3.org/Provider/Style/URI) - existing links should continue to work. Previously we have used alias for this to keep generating the page at locations it used to be at previously. We can also use Netlify redirects as an alternative, but it's important that existing links keep working.

There is a plugin for mkdocs that does this!
https://github.com/mkdocs/mkdocs-redirects
https://github.com/mkdocs/mkdocs-redirects

dickermoshe:
>  Related to the first point, and I haven't checked it yet so it may be done already. We need to self-host fonts, CSS and scripts - the site shouldn't include any assets from third-party sites.

Flutter uses gstatic, is this ok?

dickermoshe:
> The index part should also mention unique indexes. That IMO also makes the two indexes vs. on index over two columns distinction more obvious.

How?

simolus3:
> Flutter uses gstatic, is this ok?

I'm not 100% sure what the privacy implications of this are, let's just self-host the assets to be safe.

simolus3:
> > The index part should also mention unique indexes. That IMO also makes the two indexes vs. on index over two columns distinction more obvious.
> 
> How?

It might give us a more concrete example, e.g. we could say that for two columns `a` and `b`, `(a=10, b=5), (a=10, b=3)` is a conflict when using two independent indexes but not when using an index covering both columns (with column names that closer relate to the examples on the page of course).

simolus3:
It appears like mirroring the OS settings on dark mode doesn't appear to be working - the page always loads in dark mode for me. You can use this option in the Chrome Developer tools to manually toggle the settings. MkDocs claims to support this as  "Automatic light / dark mode", I think that's the only mode we should be using.

grafik

PR #3133, opened 2024-08-06
Change type of `UpsertMultiple.clauses` from `List<DoUpdate>` to `List<UpsertClause>`

wnowak-vialytics:
SQLite supports mixing `DO UPDATE` and `DO NOTHING` conflict actions (see https://www.sqlite.org/lang_upsert.html), while the `UpsertMultiple` class allows the use of only the former. This change aims to fix that using a generic type extended by the `DoUpdate` and `DoNothing` classes.

The `InsertStatement._writeOnConflict` is already compatible with the change and correctly handles nested `UpsertMultiple` objects.

### DIFF
--- a/drift/lib/src/runtime/query_builder/statements/insert.dart
+++ b/drift/lib/src/runtime/query_builder/statements/insert.dart
@@ -556,12 +556,14 @@ class DoUpdate<T extends Table, D> extends UpsertClause<T, D> {
 ///
 /// The first [DoUpdate.target] matched by this upsert will be run.
 class UpsertMultiple<T extends Table, D> extends UpsertClause<T, D> {
-  /// All [DoUpdate] clauses that are part of this upsert.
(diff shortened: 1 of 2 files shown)


simolus3:
Thanks! I agree with the changes, could you also add a simple test for them? We have existing tests here: https://github.com/simolus3/drift/blob/24905dc6c23a375999c1771a1ab089f5918b4f66/drift/test/database/statements/insert_test.dart#L282-L341

I think you can adapt one of them or add a new one mixing `DoUpdate` and `DoNothing` to validate that the generated SQL is what we'd expect.

wnowak-vialytics:
Sure, I can also add something for nested `UpsertMultiple`. I'll try to squeeze it in today.

wnowak-merix:
@simolus3 Done!

simolus3 (approved):
Awesome, thanks!

simolus3:
The CI failures appear to be due to changes in Dart 3.5.0 - I'll take a look.

PR #3153, opened 2024-08-15
Table discovery to ignore private classes

### LINKED ISSUE #3152: Generator table discovery should exclude private classes
The generator does not currently exclude private classes (ones whose names start with an underscore) from the drift file analysis process.

_To clarify - it's specifically elements discovered during `.drift` file traversal, that exhibit this issue. Private dart-file  classes do not exhibit this issue, UNLESS that dart file can also be reached via a drift file - then it also becomes problematic._

This can result in the final generated dart code creating derivatives of classes that are not legally accessible, i.e. - uncompilable code.

For example

```
// public definition of table, note the fact it derives a private class
class MyTable extends _$MyTable {
  TextColumn get customField => text()();
}

// this is a private base-class whose purpose is to define many of the fields. The public/private split of classes is a convention we follow 
class _$MyTable extends Table {
  TextColumn get description => text()();
}
```

following is what drift will generate
```
// this is the generated table for the public "MyTable" class
class $MyTableTable extends MyTable with TableInfo<$MyTableTable, MyTableData> {
// ...
}

// this is also generated, for the PRIVATE "_$MyTable" class - this should not be here - it won't compile because "_$MyTable" cannot be resolved
class $_$MyTableTable extends _$MyTable
    with TableInfo<$_$MyTableTable, _$MyTableData> {
// ...
}
```

I've supplied a test case, and proposed fix PR, below.

Adam-Langley:
Closes #3152 

Rudimentary modification to the discovery process, to exclude Table classes whose names begin with an underscore.


dickermoshe:
Adding tests for this should be quite simple.
Just add a table that starts with an underscore to https://github.com/simolus3/drift/blob/develop/drift/test/generated/todos.dart and add it to the database.
And rerun the build in `drift/drift`.

Adam-Langley:
@dickermoshe looking at the failing test - this entire "fix" may not be as simple as I've treated it.

The failing test is based on the specific scenario where by a private Table derivative class "_SomeTable" is defined in the same file as the db - hence the generated code DOES actually have access to the private class.
I wasn't even considering that someone would do this - so as I've written it, it's a breaking change.

I think I need to enhance it, so that it doesnt just blindly exclude tables starting with an underscore, but more carefully checks if the generated code can legally use the private class.

I have to figure out how to do this.

simolus3:
What if we discover these elements, but add a property to `DriftElementId` denoting whether the found element is private?

This snippet is responsible for determining the tables added under an `includes`/`import` rule:

https://github.com/simolus3/drift/blob/c49bd4975f1251004f3294434f1c3e9b74c50f52/drift_dev/lib/src/analysis/resolver/file_analysis.dart#L66-L76

What if we change line 68 here to only include analyzed entries that aren't private? I think that would solve the problem.

>  Just add a table that starts with an underscore to https://github.com/simolus3/drift/blob/develop/drift/test/generated/todos.dart and add it to the database.

If I understand this correctly, the intention is for that table to not be recognized in that case - that's hard to test. It might be better to use a simulated build in `drift_dev/test/backends/build/build_integration.test` as those are able to assert failing builds.

Adam-Langley:
@simolus3 - thanks for the direction.

Isn't the true test, whether the user defined Table derivative is inside the same _library_ as the user defined Db derivative class?
Because private table classes (as per the failing integration test) are entirely legal - they just need to exist in the same library as the db class (well, more specifically, where all the drift generated code ends up).

From this, it seems like we need to track the originating library of each table, so that when it comes time to generate all the code, we can exclude any private classes that aren't in the same destination library.generated.

simolus3:
Good point!

> it seems like we need to track the originating library of each table

This is already tracked as the URI component of `DriftElementId`.

Adam-Langley:
I've decided to set up a build environment, to debug a solution. I've followed the instructions and run `dart run build_runner generate-build-script` in the drift folder, on the `develop` branch, and get a dump of errors:

`Because drift_dev <2.9.0 depends on sqlite3 >=0.1.6 <2.0.0 and drift_dev >=2.8.0 <2.12.0 depends on analyzer ^5.12.0, drift_dev <2.12.0 requires sqlite3 >=0.1.6 <2.0.0 or analyzer ^5.12.0.
And because drift_dev >=2.12.0 <2.13.0 depends on drift >=2.12.0 <2.13.0, drift_dev <2.13.0 requires sqlite3 >=0.1.6 <2.0.0 or analyzer ^5.12.0 or drift >=2.12.0 <2.13.0.
And because drift_dev >=2.13.0 <2.14.0 depends on drift >=2.13.0 <2.14.0 and drift_dev >=2.14.0 <2.15.0 depends on drift >=2.14.0 <2.15.0, drift_dev <2.15.0 requires sqlite3 >=0.1.6 <2.0.0 or analyzer ^5.12.0 or drift >=2.12.0
  <2.13.0 or >=2.13.0 <2.14.0 or >=2.14.0 <2.15.0.
And because drift_dev >=2.15.0 <2.16.0 depends on drift >=2.15.0 <2.16.0 and drift_dev >=2.16.0 <2.17.0 depends on drift >=2.16.0 <2.17.0, drift_dev <2.17.0 requires sqlite3 >=0.1.6 <2.0.0 or analyzer ^5.12.0 or drift >=2.12.0
  <2.13.0 or >=2.13.0 <2.14.0 or >=2.14.0 <2.15.0 or >=2.15.0 <2.16.0 or >=2.16.0 <2.17.0.
And because drift_dev >=2.17.0 <2.18.0 depends on drift >=2.17.0 <2.18.0 and drift_dev >=2.18.0 <2.19.0 depends on drift >=2.18.0 <2.19.0, drift_dev <2.19.0 requires drift >=2.12.0 <2.13.0 or >=2.13.0 <2.14.0 or >=2.14.0 <2.15.0
  or >=2.15.0 <2.16.0 or >=2.16.0 <2.17.0 or >=2.17.0 <2.18.0 or >=2.18.0 <2.19.0 or sqlite3 >=0.1.6 <2.0.0 or analyzer ^5.12.0.
And because drift_dev >=2.19.0 <2.20.0 depends on drift >=2.19.0 <2.20.0 and drift_dev >=2.20.0 depends on drift >=2.20.0 <2.21.0, every version of drift_dev requires drift >=2.12.0 <2.13.0 or >=2.13.0 <2.14.0 or >=2.14.0
  <2.15.0 or >=2.15.0 <2.16.0 or >=2.16.0 <2.17.0 or >=2.17.0 <2.18.0 or >=2.18.0 <2.19.0 or >=2.19.0 <2.20.0 or >=2.20.0 <2.21.0 or sqlite3 >=0.1.6 <2.0.0 or analyzer ^5.12.0.
And because drift depends on both sqlite3 ^2.4.3 and analyzer ^6.4.1, every version of drift_dev requires drift >=2.12.0 <2.13.0 or >=2.13.0 <2.14.0 or >=2.14.0 <2.15.0 or >=2.15.0 <2.16.0 or >=2.16.0 <2.17.0 or >=2.17.0 <2.18.0
  or >=2.18.0 <2.19.0 or >=2.19.0 <2.20.0 or >=2.20.0 <2.21.0.
So, because drift depends on drift_dev any and drift is 2.20.0-dev, version solving failed.`

Any ideas?
I'm just working off the latest dev branch - nothing changed. Latest flutter install.

Adam-Langley:
Hi guys.

After setting up my dev environment for drift, and investigating a corrected implementation based somewhat on your suggested, I nailed down some more specifics about the issue.

In my test project I had attached to the issue report, the public/private table classes were linked via a .drift file import. You need this specific scenario to trigger the issue.
If the table is *only* linked via the `tables: []` annotation field - drift will NOT generate a table for the private superclass. Which is great...

Anyway, I thought this was an interesting find.
I couldn't use the 'id' property to detect "private scope" as suggested, as its the name of the dart class, not the table name, thats the issue. Hence, please check my revised code, I've just filtered the availableElements collection (as suggested).
I determine accessibility by testing the `baseDartName` of DriftTable elements.
I then check that the elements libraryUri will be different from the database elements libraryUri, before excluding it.

I've added the previously failing scenario (public table class deriving from private table class, linked from a separate dart class, reached through a drift file only) - to the "custom_tables" test.

And I've confirmed that `regressions_2166` no longer fails.

drift_dev/lib/src/analysis/resolver/file_analysis.dart
```diff
@@ -73,6 +73,14 @@ class FileAnalyzer {
               .whereType<DriftElement>()
               .followedBy(availableByDefault)
               .transitiveClosureUnderReferences()
+              .where((e){
+                if (element is DriftDatabase) {
```
simolus3: Why would this only apply to databases here?

drift_dev/lib/src/analysis/resolver/file_analysis.dart
```diff
@@ -73,6 +73,14 @@ class FileAnalyzer {
...
+                  if ((e is DriftTable) && e.baseDartName.startsWith('_')) {
+                    return e.id.libraryUri == element.id.libraryUri;
+                  }
```
simolus3: Add a comment explaining the intention of only including public elements of other libraries, allowing private table classes to be defined without breaking the build.

drift_dev/lib/src/analysis/resolver/file_analysis.dart
```diff
@@ -73,6 +73,14 @@ class FileAnalyzer {
               .whereType<DriftElement>()
               .followedBy(availableByDefault)
               .transitiveClosureUnderReferences()
+              .where((e){
```
simolus3: This where should be in the one from line 70, the private elements shouldn't even reach `transitiveClosureUnderReferences`.

drift_dev/lib/src/analysis/resolver/file_analysis.dart
Adam-Langley: My understanding of what's going in the overall code is likely to blame here.

I needed to grab the instance of DriftDatabase so that I could access its `id.libraryUri` for comparison against each element that I wanted to filter out.
Maybe this file is entirely the wrong location for the logic?

I used line 96 as a template for line 76, to get a reference to the db.

drift_dev/lib/src/analysis/resolver/file_analysis.dart
Adam-Langley: Absolutely, done.

drift_dev/lib/src/analysis/resolver/file_analysis.dart
Adam-Langley: What I found, using the example case in this comment (and the project file in the issue report), is that the private classes are sneaking in via '.followedBy(availableByDefault)' - It can be moved above `transitiveClosureUnderReferences ` move it up one, but any further and it wont work anymore.

simolus3: > followedBy(availableByDefault)

Wait really? Do you have an example for that? `availableByDefault` only includes tables and views explicitly added to the `@DriftDatabase` annotation, it should be impossible to reference private tables with that right?

drift_dev/lib/src/analysis/resolver/file_analysis.dart
simolus3: This might be flow typing not applying in nested functions, but since this is all wrapped in an `if (element is BaseDriftAccessor)` you can safely use `element!.id` instead.

My comment was that we may have a similar problem with DAOs, as those are not `DriftDatabase`s.

Adam-Langley:
Sorry for the short response, just on my phone but wanted to get you a
response back asap.

Yes all the examples I have provided so far exercise exactly this issue -
it all relates back to the derivation of a private class.

Here’s a description of the walk-path.

1. @DriftDatabase “includes” a “.drift” file.
2. The drift file references a public, user-defined class A.
3. The class A derives a private, user-defined class _B.
4. Class _B derives Table.

All of these classes come through in ‘availableByDefault’.
Debug the code in this issue, or the PR, you’ll see it.

Thanks

simolus3:
I can't reproduce what you're saying with this test:

```dart
import 'package:drift_dev/src/analysis/results/database.dart';
import 'package:drift_dev/src/analysis/results/element.dart';

import '../test_utils.dart';

void main() {
  test('does not include private superclasses through import', () async {
    final backend = await TestBackend.inTest({
      'a|lib/database.dart': '''
import 'tables.dart';

@DriftDatabase(includes: {'test.drift'}, tables: [PartOfDatabase])
class Database {}
''',
      'a|lib/tables.dart': '''

class PartOfDatabase extends _NotPartOfDatabase {

}

class _NotPartOfDatabase extends Table {
  IntColumn get id => integer()();
}

''',
      'a|lib/test.drift': '''
import 'tables.dart';
''',
    });

    final state = await backend.analyze('package:a/database.dart');
    backend.expectNoErrors();

    final database = state.analyzedElements.whereType<DriftDatabase>().single;
    final availableByDefault = <DriftSchemaElement>{
      ...database.declaredTables,
      ...database.declaredViews,
    };
    expect(availableByDefault, hasLength(1));
  });
}
```

`availableByDefault` here copies the definition in `FileAnalysis`, but I also set a breakpoint there and it did not include the private table as expected. The private tables are coming in through `import.expand((reachable) => ...)` only.

Adam-Langley:
Thanks for the test case @simolus3 - let me try to figure out why your case behaves differently from my case (in the PR).

UPDATE: Ok, I've figured some things out. Apologies - I've put you wrong - but also your test case above is a little broken.

### Firstly: availableByDefault.
Apologies, it wasn't MY scenario that was populating this with private classes. I saw this scenario happen in `regress_2166_test.dart`. My crude assumption when I saw `_SomeTable` appear in this collection, was that I needed to put my code AFTER this point, so my predicate would be aware of all processed classes before making any decisions. That was before I was aware of where `_SomeTable` was coming from, and that there was a broken integration test as a result of my original, rudimentary fix.

However, now I understand the drift code better, I see it's in this collection because that private table is explicitly referenced in `@DriftDatabase(tables: [])` .
image

So, yes, you're right, I don't need to be this late in the fluent chain to fix the specific test case you've created above.

### Second: Small typo in your test case
`includes: ` needs to be `include:` Hence it's not processing the drift file at all.
image

**Off-topic**: Should perhaps the drift file processor be throwing when it detects superfluous attributes? That would mitigate the risk of these test-case issues. Of course, actual runtime code has the benefit of the dart static analyzer.

**Final Question**: I'll include your test case in the PR - however it should be actually testing the detected tables, right?
I'll rename it slightly and add the assert - please comment if not suitable.
Thanks @simolus3

drift_dev/lib/src/analysis/resolver/file_analysis.dart
Adam-Langley: I've added a new comment further down which addresses everything, including this (my mistake)

drift_dev/lib/src/analysis/resolver/file_analysis.dart
Adam-Langley: Thanks @simolus3  - new comment at the end which encapsulates all my learnings.

drift_dev/lib/src/analysis/resolver/file_analysis.dart
```diff
@@ -71,6 +71,18 @@ class FileAnalyzer {
...
               })
               .whereType<DriftElement>()
+              .where((e){
+                // Exclude any private tables (name prefixed with "_") that do not
+                // reside in the same library as the DriftDatabase.
+                // Failure to exclude these, can generate dart code which references
+                // classes that cannot be legally accessed - and will not compile.
+                // Private classes in residing in the same library are allowed, as
+                // per dart language accessibility rules.                
+                if ((e is DriftTable) && e.baseDartName.startsWith('_')) {
```
simolus3: > I couldn't use the 'id' property to detect "private scope" as suggested, as its the name of the dart class, not the table name, thats the issue

How is this an issue again? The id should point to the definition of the class, and we want to exclude it if it's private, right?

The `e is DriftTable` condition is a bit unfortunate here because it means that this won't apply to views. So I'm wondering if we can use the id or `e.entityInfoName` as an alternative.

Adam-Langley: Hi, apologies, I may not have explained what I found very well the first time.

I couldn't use 'id.name' to detect "privateness", because id.name appeared to be the DataClassName, not the name of the table class.

For example, if I change my fix, to compare 'e.id.name' instead of 'e.baseDartName', then the following will fail to be picked up as a private table class (notice how @DataClassName does not have the underscore, this is what goes into id.name).
I did consider - why would someone have a private table, but a public dataclass? But I really don't know, but if they do - it will just break, which I didn't like.

```
@DataClassName('NotPartOfDatabase')
class _NotPartOfDatabase extends Table {
  IntColumn get id => integer()();
}
```

Your suggestion of `entityInfoName` looks viable, that will cover both tables and views, and I just need to look for `$_` instead of `_`.

The predicate then becomes:
```
if (e is DriftElementWithResultSet && e.entityInfoName.startsWith('\$_')) {
  return e.id.libraryUri == element.id.libraryUri;
}
```

Do you think that's the way to go?
I've committed it with some enhancements to the integration test for your consideration.

Thanks again Simon.

Adam-Langley: Just an aside @simolus3 - thanks for your patience. I'm doing my best to understand the drift_dev internals with the time I have available to me, to produce something befitting the level of quality you've achieved in drift. I appreciate that this is probably pulling your away from your day-job too.

simolus3 (approved):
LGTM, thanks!

PR #3170, opened 2024-08-23
Review for docs to talk about known issue for same name classes

FMorschel:
Review for docs to talk about a known issue for same-name classes.

As discussed on Discord with simolus3, he suggested I use the modular code generation in my case, I thought it would be good to list this in the docs as well.

Let me know what you think! Also, not sure if the links are right but copied from the setup page.


simolus3 (approved):
Nice, thank you!

PR #3176, opened 2024-08-26
Adding documentation about under-development DevTools extension

FMorschel:
I could not find any docs on the DevTools Extension so I figures at least adding that to the FAQ on that part would be good.


simolus3 (approved):
Thanks again!

PR #3178, opened 2024-08-27
Adding more instructions to dev the DevTools extension

FMorschel:
Only a few minor instructions to help anyone that wants to see how the DevTools extension works or wants to help develop it.

### DIFF
--- a/extras/drift_devtools_extension/README.md
+++ b/extras/drift_devtools_extension/README.md
@@ -20,22 +20,28 @@ flutter run --debug
 Once the example app launches, it will print out something like:
 
 ```console
-Connecting to VM Service at ws://127.0.0.1:33333/7j9YZqoiWe8=/ws
+A Dart VM Service on Linux is available at: http://127.0.0.1:42439/XZyfqXFUYlg=/  
+The Flutter DevTools debugger and profiler on Linux is available at:  
+http://127.0.0.1:9102?uri=http://127.0.0.1:42439/XZyfqXFUYlg=/  
 ```
 
-or
+You can simply copy the first url it shows inside the `Dart VM Service Connection` that appears
+on the right side of the simulated environment.
 
-```console
-This app is linked to the debug service: ws://127.0.0.1:33333/7j9YZqoiWe8=/ws
-```
-
-or even
-
-```console
-Debug service listening on ws://127.0.0.1:33333/7j9YZqoiWe8=/ws
-```
-
-What you will need to do is to copy everything from `//127.0.0.1` to `=/` (no `ws`) and
+> P.S.: If you are running it (example app/other) from an IDE it will print
+> something like:
+>
+> ```console
+> Connecting to VM Service at ws://127.0.0.1:33333/7j9YZqoiWe8=/ws
+> [...]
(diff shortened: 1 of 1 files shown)


extras/drift_devtools_extension/README.md
```diff
@@ -5,6 +5,37 @@ based on the [drift_db_viewer](https://pub.dev/packages/drift_db_viewer) package
...
+
+You'd also need to run an app with a drift database for you to be able to see it running.
+You can use our example app for that.
+
+```console
+cd ../../examples/app/
+flutter run --debug
+```
+
+Once the example app launches, it will print out something like:
```
simolus3, line 20: For me it doesn't print a ws URI, I wonder if that's OS specific? Mine says:

```
A Dart VM Service on Linux is available at: http://127.0.0.1:42439/XZyfqXFUYlg=/
The Flutter DevTools debugger and profiler on Linux is available at:
http://127.0.0.1:9102?uri=http://127.0.0.1:42439/XZyfqXFUYlg=/
```

I can paste the first link into the simulated environment without any modifications.

FMorschel: I think you are running from the CLI, right? I remember it being different somehow. In vsCode at least it prints as I wrote. I could test it on Android Studio as well to see if this is a debugger/LSP thing or what. Anyway, I can rewrite this section when I do that.

FMorschel: I asked DanTup (developer of the Dart-Code extension) on Discord and I got the answer:

> I'm not sure exactly who is printing each of those lines, but the first line in the first example is printed by the debug adapter - it prints the VM Service URL that it is connecting to.
> 
> In the second example, the URL is essentially the same, but it looks like it is really the old "Observatory URL", which was a web app used for debugging. You'd access that directly in a browser, to it uses HTTP as the protocol, and doesn't have the "/ws" websocket suffix.
> 
> Many Dart tools will accept the second format (a HTTP "observatory URL") and just change the protocol and add /ws automatically.
> 
> The final URL in the second example is that dart run has automatically started a DevTools server and is providing the URL. When running from VS Code, Dart-Code will do this internally and doesn't print the URL because you can access DevTools using the built-in commands instead (whereas from the CLI, printing the URL is the only way to give you it).
> 
> (and finally, the reason the port numbers and the strings after the port number are different between each run is just that they're random... a random port is used to avoid clashing with other instances that might be running and the other string is a random authentication token to ensure no other processes that can connect to your localhost can scan for VM Services and use them to evaluate code)

FMorschel: I'll rewrite this part and commit it here.

simolus3: > I think you are running from the CLI, right?

Well your previous snippet suggested running the app on the CLI :D Perhaps we should mention how the output looks like for both?

simolus3 (approved):
LGTM :+1:

extras/drift_devtools_extension/README.md
FMorschel: I realized that hahaha. Did mentioned the output for both.

PR #3186, opened 2024-08-30
Fix the links of the comments on update and delete methods

luxun1910:
This is a trivial fix of comments.
I couldn't find the details of delete and update methods from the links of the comments. That's why I fixed the links.


simolus3 (approved):
Good catch, sorry for breaking the URLs and thanks for the fix!

PR #3190, opened 2024-08-31
feat(drift_dev): Support view triggers in drift SQL

dnys1:
Adds support for triggers on views created by `CREATE TRIGGER` expressions.

https://www.sqlite.org/lang_createtrigger.html#instead_of_triggers

### DIFF
--- a/drift_dev/lib/src/analysis/resolver/resolver.dart
+++ b/drift_dev/lib/src/analysis/resolver/resolver.dart
@@ -48,7 +48,7 @@ class DriftResolver {
         return await _deserializer.readDriftElement(element);
       }
     } on CouldNotDeserializeException catch (e, s) {
-      driver.backend.log.fine('Could not deserialize $element', e, s);
+      driver.backend.log.warning('Could not deserialize $element', e, s);
     }
 
     // We can't resolve the element from cache, so we need to resolve it.
--- a/drift_dev/lib/src/analysis/results/query.dart
+++ b/drift_dev/lib/src/analysis/results/query.dart
@@ -398,11 +398,9 @@ class UpdatingQuery extends SqlQuery {
   @override
   final AstNode root;
 
-  bool get isOnlyDelete =>
-      updates.isNotEmpty && updates.every((w) => w.kind == UpdateKind.delete);
+  bool get isOnlyDelete => updates.every((w) => w.kind == UpdateKind.delete);
 
-  bool get isOnlyUpdate =>
-      updates.isNotEmpty && updates.every((w) => w.kind == UpdateKind.update);
+  bool get isOnlyUpdate => updates.every((w) => w.kind == UpdateKind.update);
 
   UpdatingQuery(
     String name,
--- a/drift_dev/lib/src/analysis/results/query.dart
+++ b/drift_dev/lib/src/analysis/results/query.dart
@@ -398,9 +398,11 @@ class UpdatingQuery extends SqlQuery {
   @override
   final AstNode root;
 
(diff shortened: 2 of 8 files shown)


simolus3 (commented):
Thanks! I think this will need some minor additions before it fully works though: In most builds, drift serializes analysis results as intermediate files to speed things up. There's a cast in `ElementDeserializer._readDriftElement` that also affects triggers in line 632.

In addition to the `TestBackend`-based tests which only exercise drift's analyzer, a simple e2e should give confidence that nothing breaks in later steps. Something like this would work with a simple assertion checking that the generated trigger uses the right view getter in generated code:

https://github.com/simolus3/drift/blob/9d41c821eeaa28054f13c7e706b4feb669f98374/drift_dev/test/backends/build/build_integration_test.dart#L404-L430

drift_dev/lib/src/analysis/resolver/resolver.dart
```diff
@@ -48,7 +48,7 @@ class DriftResolver {
         return await _deserializer.readDriftElement(element);
       }
     } on CouldNotDeserializeException catch (e, s) {
-      driver.backend.log.fine('Could not deserialize $element', e, s);
+      driver.backend.log.warning('Could not deserialize $element', e, s);
```
dnys1, line 51: In order to have the test fail for deserialization exceptions, I needed to change this to a warning.

drift_dev/lib/src/analysis/results/query.dart
```diff
@@ -398,9 +398,11 @@ class UpdatingQuery extends SqlQuery {
   @override
   final AstNode root;
 
-  bool get isOnlyDelete => updates.every((w) => w.kind == UpdateKind.delete);
+  bool get isOnlyDelete =>
+      updates.isNotEmpty && updates.every((w) => w.kind == UpdateKind.delete);
```
dnys1: Since `Iterable.every` returns `true` for empty iterables, this was always true for views leading to generated code with `updateKind: UpdateKind.delete` even for `INSERT` and `UPDATE` statements. My sense was that this was not intentional, but let me know if it was.

dnys1:
Good catch! I've added integration tests to verify the output and updated a couple smaller things as noted.

drift_dev/lib/src/analysis/resolver/resolver.dart
simolus3: Which is appropriate, considering that this shouldn't fail :+1:

drift_dev/lib/src/analysis/results/query.dart
simolus3: I agree that this is kind of weird, but can we move the `isNotEmpty` check to `QueryWriter._writeUpdateKind`?

The intended behavior is that for writes we don't recognize, we don't generate an `updateKind` parameter at all. So I think `_writeUpdateKind` should just check for the list being empty and not emit anything in that case.

I have also improved the situation somewhat in ed72f6fa19d6ef844030c0c87c29f2d107e1483e by including views in `updates`. With this PR landing then, we should be able to recognize these patterns for stream queries as well (an insert on the view in `.drift` files propagates to the table thanks to the trigger, and that then updates everything listening on that).

simolus3 (commented):
Thanks for the tests (and for catching the `isOnlyUpdate`/ `isOnlyDelete` issue). I've added support for recognizing writes to views on `develop` so this should be less of an issue now.

dnys1:
@simolus3 Makes sense, thanks! Updated 👍

simolus3 (approved):
Thanks for all your contributions!

dnys1:
My pleasure! Thanks for drift 💪😎

PR #3191, opened 2024-08-31
fix(drift_dev): Strip leading numerics in modular generation

dnys1:
For drift modules of the form `\d+_foo.drift` which is a common naming convention for migration files, strip the leading numerics when generating module accessors.

Currently, module accessors result in classes of the form `class 001Foo extends ModuleAccessor` which is invalid Dart syntax. Escaping with `$` also solves the problem, but simply stripping the numerics leads to cleaner code.


simolus3 (approved):
Thanks!

PR #3192, opened 2024-08-31
fix(drift_dev): Infer `json_extract` type from context

dnys1:
Currently, inference of all `json_extract` expressions results in an unknown type for inference. However, there are cases where the type can be known/coerced, for example with `INSERT INTO SELECT` statements when the column types are inferred from the written table.

In those cases, we use the expectation type to resolve `json_extract` expressions.


simolus3:
Nice, thanks!

PR #3193, opened 2024-08-31
Recreating examples projects to migrate android

### LINKED ISSUE #3189: Migrate Android app plugin loader - examples
**Describe the bug**

Running the examples/app on Android gives the following message: 

```console
You are applying Flutter's app_plugin_loader Gradle plugin imperatively using the apply script method, which is deprecated and will be removed in a future release. Migrate to applying Gradle plugins with the declarative plugins block: https://flutter.dev/to/flutter-gradle-plugin-apply
```

Perhaps we could simply delete the /android folder for all examples and whatever else needs this and run `flutter create .`? Or we can manually migrate everything. I volunteer to do either.

FMorschel:
Fixes https://github.com/simolus3/drift/issues/3189.

I believe these are the only two that needed this.

- examples/app
- examples/encryption


simolus3 (approved):
Thanks!

PR #3204, opened 2024-09-06
Migrate to MkDocs

dickermoshe:
This PR does the following:

1. Moves to MkDocs
2. New syntax highlighting
3. New Flutter Example
4. Remove Zap

TODO:
- [x] Old Links Should Work
- [x] Docs added since this PR was started need to be added back in
- [x] Update Favicon
- [x] Confirm that code blocks are correct
- [x] Web compatibility check
- [x] Code Blocks in Tabs
- [x] Edit edit_uri
- [x] The `Change cookie settings` link from the footer should be removed since we're not setting any cookies.

Commits are broken up pretty neatly


dickermoshe:
@simolus3 
https://66db487c5363bf1fd5267305--moor.netlify.app
Thoughts?

simolus3:
Nice, thanks!

This looks good, but since most of the content is under "Reference" at the moment, I'd rename the tabs a bit. I think "Documentation & Guides" should be called "Home" (or something like that ?) and "Reference" should probably just be "Documentation".

> Only Dark Mode

Is there a reason we support only dark mode? If possible, I'd prefer both, ideally using the  [prefers-color-scheme](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) CSS query to determine whether to use dark/light mode.

dickermoshe:
The syntax highlighting injects the colors using inline css.
Refactoring to use a single style sheet would be very time consuming.
Take a look at the code, most of it was copied from Google/ServerPod

dickermoshe:
@simolus3 
Got some free time. Checkout the new [deployment](https://deploy-preview-3204--moor.netlify.app/setup/). Uses dark/light mode based on device and has corrected section names

dickermoshe:
@simolus3 Are you OK with just the not hitting a 404:

```
/docs/something -> /docs/something-else          (No 404)
/docs/something#header -> /docs/something-else   (Won't go to header)
```
Or do you need the header (permalinks) to be the same?
```
/docs/something#header-name -> /docs/something-else#other-header-name   (Will go to header)
```

The are tons and tons of permalinks and mapping all of them would be hours of brain numbing work.

dickermoshe:
@simolus3 I will be using this to create the redirects:
https://docs.netlify.com/routing/redirects/

dickermoshe:
@simolus3 
It's all good to go!
Let me know If there is something I've missed

dickermoshe:
@simolus3 
Flutter broke `webdev` with the latest release of devtools_shared: https://github.com/dart-lang/webdev/issues/2492

dickermoshe:
Aaaand now I'm done.

simolus3:
Thank you very much, this looks great! I'll take a deeper look at this soon, just some quick thoughts:

1. Do we depend on Flutter for building the docs now? If so, I think we can remove the stubs packages and just use the proper dependencies?
2. We're still using a custom highlighter, right? Is it possible to inject something into the CSS generated by Material for MkDocs (e.g. is there some hooks sass file that we can override)? That way we might be able to use predefined classes instead of inline styles. Don't worry about it though, I'm just wondering if it's possible and I can do it myself since I've already caused so much back-and-forth steps with the syntax highlighting.

dickermoshe:
1. Sure
3. We are not using inline css for the syntax highlighting. We are using classes. (I figured out how to do it) The css for it can be found in `docs/docs/css/syntax_highlight.css`. You can add more css in that file. You can also create a brand new css file, just make sure to add an entry for it in `extra_css:` in `mkdocs.yml`. We can do SaSS too by compiling it to css before we build the site.

### @simolus3 See the updated `docs/Readme.md` for info on how syntax highlighting works

simolus3:
Ah one thing I've forgotten - the deployment previews we generate for branches and PRs must have a `noindex, nofollow` meta tag for search engines so that they can't show up in favor of the main website. The older docs were able to do that with different builder configurations. Is there a way to do the same with MkDocs (or do we have the option to build a dev variant of the site that has those tags)

docs/docs.sh
```diff
@@ -0,0 +1,142 @@
...
+
+# Get the 1st argument passed to the script and convert it to lowercase
+arg1=$(echo $1 | tr '[:upper:]' '[:lower:]')
+arg2=$(echo $2 | tr '[:upper:]' '[:lower:]')
+
+
+# Build the MkDocs container which will be used to build/serve the MkDocs project
+build_container () {
+    echo "Building the container..."
+    docker build -t mkdocs:latest ./mkdocs/
```
simolus3: Let's not use `mkdocs` as a tag, it might conflict with others. Perhaps `drift_docs_builder`?

docs/docs.sh
```diff
@@ -0,0 +1,142 @@
...
+
+    # The below command will compile the dart code in `/web` to js & run build_runner
+    webdev build -o web:build/web -- --delete-conflicting-outputs
+    if [ $? -ne 0 ]; then
+        echo "Failed to build the project"
+        exit 1
+    fi
+
+    # Run the build_runner command to generate files in the `test` directory
+    dart run build_runner build --delete-conflicting-outputs
```
simolus3: `webdev build` has an optimization to not run a full build since we only care about `web`. Since we're running a full build afterwards, let's run `build_runner` first and then use `webdev` to emit the files.

Also, let's add a dev dependency on `webdev` and use `dart run webdev` instead of relying on global dependencies.  And to be sure, pass `--release` to both `build_runner build` and `webdev build`.

docs/docs.sh
```diff
@@ -0,0 +1,142 @@
...
+        exit 1
+    fi
+    # Run build_runner to generate files in the `test` directory
+    dart run build_runner build --delete-conflicting-outputs
+    if [ $? -ne 0 ]; then
+        echo "Failed to build the example project"
+        exit 1
+    fi
+
+    flutter build web --base-href "/examples/app/" --no-web-resources-cdn
```
simolus3: Nice 👍 Do we also have a way to disable the service worker?

dickermoshe: I don't understand what you mean

docs/docs.sh
dickermoshe: Done

docs/docs.sh
dickermoshe: Done

dickermoshe:
> Ah one thing I've forgotten - the deployment previews we generate for branches and PRs must have a...

Added a robots.txt for these sites > https://66e4b8c256c80c93df7ec8e7--moor.netlify.app/robots.txt

docs/docs.sh
simolus3: Flutter generates a service worker that caches assets for apps so that they load faster on subsequent visits. The worker doesn't properly realize when assets have changed though, or wants us to include a version number everywhere. So we should disable the service worker in the example app, no worker should run on the documentation website.

dickermoshe: Done

dickermoshe:
Is there anything I missed which is holding this PR up?

I've started working on the rewrite and would love to get this done sooner than later.

simolus3:
Apologies on the delays reviewing this. I've taken another look, and only two remaining things stood out to me. Once the first one is addressed we can merge this:

- When I run `./docs.sh serve`, generated build assets like `compatibility.dart.js` are not served. Is this intentional? Even `--with-build-runner` does not include any output argument, so the compiled files are not visible. I get that we can't rebuild the Flutter example on every edit, but is it possible to bring `build` and `serve` closer together and very clearly document the remaining differences?
- (we can do this one after the merge) If possible, can we avoid spaces in the URI and use custom paths (e.g. `/Generation options` -> `/generation_options`) instead? Browsers should copy them with `%20`, but these still look a bit ugly and might be harder to share because some messengers don't recognize the entire URL.

dickermoshe:
1. Will do.
2. This is because the docs folders have spaces in their names. I'll fix this during the rewrite.

dickermoshe:
1. The web build and the Flutter example project don't change often, so I don't think the serve command **needs** to include them.

    We can't `watch` the `webdev` build because the generated files need to be organized after each build, which is something we can't do in a bash script. Although we should at least build it once. 

    I'll update the script to do that.
    I've also updated the docs in /docs/Readme.md to explain all this


2. I've renamed the folders so that there are no spaces & added clearer `docs.sh serve` docs to the ReadMe.md

simolus3:
Building once is sufficient, thanks! I just noticed this because the interactive feature detection on `/web` was broken since I've used `serve` without ever running `build`.

simolus3 (approved):
Thank you so much!!

dickermoshe:
@simolus3 How does the production site get released?
https://drift.simonbinder.eu/docs/ still has the old site

simolus3:
It's released when I push to the `latest-release` branch, which I do after a drift release so that newer APIs mentioned in the docs don't confuse users.

simolus3:
Finally released :rocket:

dickermoshe:
Look'n good!

PR #3209, opened 2024-09-09
Delete unreachable `default` clauses.

stereotype441:
The Dart analyzer will soon be changed so that if the `default` clause of a `switch` statement is determined to be unreachable by the exhaustiveness checker, a new warning of type
`unreachable_switch_default` will be issued. This parallels the behavior of the existing `unreachable_switch_case` warning, which is issued whenever a `case` clause of a `switch` statement is determined to be unreachable. For details see
https://github.com/dart-lang/sdk/issues/54575.

This PR deletes unreachable `default` clauses from `drift` now, to avoid spurious warnings when the analyzer change lands.

### DIFF
--- a/drift/lib/src/runtime/executor/helpers/engines.dart
+++ b/drift/lib/src/runtime/executor/helpers/engines.dart
@@ -501,8 +501,6 @@ class DelegatedDatabase extends _BaseExecutor {
             this, context, noTransactionDelegate);
       case SupportedTransactionDelegate supported:
         return _WrappingTransactionExecutor(this, supported);
-      default:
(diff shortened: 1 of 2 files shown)


simolus3 (approved):
Thanks :+1: I assume you need this rolled into g3 as well to land the analyzer changes? IIRC that's mirroring the `latest-release` branch, but I can release these changes and push if needed.

stereotype441:
> Thanks 👍 I assume you need this rolled into g3 as well to land the analyzer changes? IIRC that's mirroring the `latest-release` branch, but I can release these changes and push if needed.

Yes, that would be great, thank you!

simolus3:
Done :+1:

PR #3222, opened 2024-09-16
feat: add support for `late final` fields

### LINKED ISSUE #3093: feat: Declare column on table with `late final`
Now we have to do:
```dart
class Foo extends Table{
  IntColumn get name => text();
}
```
but this looks nicer
```dart
class Foo extends Table{
  late final name => text();
}
```

dickermoshe:
Writing fields with getters is hard. You are effectively writing the column twice with `IntColumn` and again with `integer()`.

This PR will allow you to write a table like:

```dart
class Users extends Table {
  late final id = integer()();
}
```

- [x] Some more `drift_dev` tests just to be 100% certain there is no bugs

Closes https://github.com/simolus3/drift/issues/3093

### DIFF
--- a/drift/test/generated/todos.dart
+++ b/drift/test/generated/todos.dart
@@ -14,7 +14,7 @@ extension type RowId._(int id) {
 }
 
 mixin AutoIncrement on Table {
-  IntColumn get id => integer()
+  late final id = integer()
       .autoIncrement()
       .map(TypeConverter.extensionType<RowId, int>())();
 }
@@ -24,17 +24,17 @@ class TodosTable extends Table with AutoIncrement {
   @override
   String get tableName => 'todos';
 
-  TextColumn get title => text().withLength(min: 4, max: 16).nullable()();
-  TextColumn get content => text()();
+  late final title = text().withLength(min: 4, max: 16).nullable()();
+  late final content = text()();
   @JsonKey('target_date')
-  DateTimeColumn get targetDate => dateTime().nullable().unique()();
+  late final targetDate = dateTime().nullable().unique()();
   @ReferenceName("todos")
-  IntColumn get category => integer()
+  late final category = integer()
       .references(Categories, #id, initiallyDeferred: true)
       .map(TypeConverter.extensionType<RowId, int>())
       .nullable()();
 
-  TextColumn get status => textEnum<TodoStatus>().nullable()();
+  late final status = textEnum<TodoStatus>().nullable()();
 
   @override
   List<Set<Column>>? get uniqueKeys => [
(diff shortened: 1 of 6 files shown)


drift/test/generated/todos.dart
```diff
@@ -159,7 +157,7 @@ class CustomRowClass {
 
 class PureDefaults extends Table {
   // name after keyword to ensure it's escaped properly
-  TextColumn get txt =>
```
simolus3: We also have the example and so on, but IMO we should leave a few of these unchanged to make sure the existing syntax isn't negatively affected.

simolus3 (commented):
Implementation looks good to me :+1: Let's also add a changelog entry to `drift_dev` though.

drift/test/generated/todos.dart
dickermoshe: Done

dickermoshe:
@simolus3 I wrote a test for this in `drift_dev` .

However, while `drift_dev` works when generating code elsewhere, i does not work when running in tests.
```dart    
final state = await TestBackend.inTest({
      'a|lib/a.dart': '''

class Students extends Table {
  @JsonKey('group_id')
  @ReferenceName('students')
  IntColumn get myGroup => integer().references(Groups,#id)();
}

class Teachers extends Table {
  @JsonKey('group_id')
  @ReferenceName('teachers')
  late final myGroup = integer().references(Groups,#id);
}

class Groups extends Table {
  late final id = integer().autoIncrement()();
}

@DriftDatabase(tables: [Groups, Students, Teachers])
class Database {}
''',
    });
```

The generator is ignoring the `late final myGroup = integer().references(Groups,#id);` column. 

Seems that the test builder is broken 🤷‍♀️.

Could you please take a look?

```
dart test test/analysis/resolver/dart/column_test.dart -p vm --plain-name 'columns by getter and declaration'
```

dickermoshe:
Added to the changelog and added a test

dickermoshe:
I was running into a bug where a column wasn't showing up.
Turns out it was because I forgot the extra `()()`

Using getters didnt have this issue
image

So to avoid handing users this massive footgun
I've added a warning in the generator that when we come across a `late final` Column Builder it print the following.

```console
line 29, column 27 of asset:drift/test/generated/todos.dart: 
It seems that you forgot to initialize the `stuff` column on the `TodosTable` table.
Add an extra pair of parentheses at the end of the column declaration as follows: `stuff = text()()`.
   ╷
29 │   late final stuff = text();
   │                           ^
   ╵
[SEVERE] drift_dev on test/generated/todos.dart:

Drift emitted warnings and the `fatal_warnings` build option is enabled.
```

@simolus3  This PR is now ready for review

dickermoshe:
Maybe we should also create a `drift_lint` package 🤔

drift_dev/lib/src/analysis/resolver/dart/table.dart
```diff
@@ -260,16 +260,46 @@ class DartTableResolver extends LocalElementResolver<DiscoveredDartTable> {
...
+
+    // Returns true if the given field is a column defined as a late final variable declaration
+    Future<bool> isLateFinalColumn(FieldElement e) async {
+      final isLateFinalField = e.isLate && e.isFinal && e.getter != null;
+      if (!isLateFinalField) return false;
+
+      if (isColumn(e.type)) {
+        return true;
+      } else {
+        if (isColumnBuilder(e.type)) {
```
simolus3, line 276: Nice :+1:

simolus3 (approved):
Thanks!

PR #3234, opened 2024-09-22
`annotate` for manager 

dickermoshe:
`annotate` is basically a way to add columns and then to order/filter and read them.
I'm imagining something like this:

```dart
// New class which will handle all the joins needed
final todoCount = db.managers.categories.annotate((a)=> a.todos.count());

// Add the annotation to the manager
final todosWithRefs = await db.managers.categories.withAnnotations([todoCount]).get()

// You can also read the annotation
for (final (todo,ref) in todosWithRefs){
  final count = refs.read(todoCount);
}

```

You can also filter and order using the added columns

```dart
final todosWithRefs = await db.managers.categories.withAnnotation([todoCount]).filter((f)=>todoCount.moreThan(10)).get()
```

@simolus3 What are your thoughts on this?

- [x] small refactor on composer (big simplification = big win)
- [x] fix doccoments after composer refactor
- [x] create the feature
- [x] write tests
- [x] write docs
- [x] add doccoments
- [x] Tests for filters on aggregates


dickermoshe:
Most of this feature is reusing the same code that ordering and filters use.
Everything is falling into place nicely

dickermoshe:
@simolus3 You can checkout this branch and look at `drift/test/manager/manager_annotation_test.dart`
Play around and let me know what your thoughts are.

simolus3:
I'll take a look tomorrow, thanks! The description looks interesting though - is this to re-use common joins across queries?

dickermoshe:
## Annotate

### Simple Expression

Sorta, it's for adding columns using expressions without having to think about joins.
The idea is  stolen from django.

Here we have a table with users.

| id | first | last | age | weight | height |
|--------|--------|--------|--------|--------|--------|
| 0 | bob| jobs | 50 | 125 | 68 |
| 1 | steve| hope | 26 | 180 | 72 |
|  2 | mary | jane | 89 | 136 | 79 |

We would like to get the users BMI, we can add that as a column using this:

```dart
// Expressions for calculating users BMI
final annotation = db.managers.users.annotation((a)=> a.weight / (a.height * a.height) * Variable(703));
```
Now we can use this `annotation` object to read/filter/order from the database.
```dart
final users = await db.managers.users.withAnnotations([annotation])
  // Filter on the value
  .filter((f) => annotation.filter.moreThan(20))
  // Order with it
  .orderBy((o) => annotation.order.asc())
  .get();

for (final (user,refs) in users){
  // Get the value from the results
  final bmi = annotation.reads(annotation)
}
```

| id | first | last | weight | height | BMI |
|--------|--------|--------|--------|--------|--------|
| 0 | bob| jobs | 125 | 68 | ? |
| 1 | steve| hope  | 180 | 72 | ? |
|  2 | mary | jane | 136 | 79 | ? |

Now, we can read the result of this column, filter on it and order on it.
```dart
final patience = db.managers.withAnnotation([annotation]).filter((a)=> annotation.filter.moreThan(20));
```

---
### Annotate Referenced Tables

But the cool part is that joins are done automatically

Here's a Markdown table showing users and groups with 5 users and 2 groups, using a primary key to connect them:

| UserID (PK) | Username | GroupID (FK) |
|-------------|----------|--------------|
| 1           | Alice    | 1            |
| 2           | Bob      | 1            |
| 3           | Charlie  | 2            |
| 4           | David    | 2            |
| 5           | Eve      | 1            |

| GroupID (PK) | GroupName |
|--------------|-----------|
| 1            | Admin     |
| 2            | User      |

We can use annotation to add referenced data to the columns:
```dart
final isAdmin = db.managers.users.annotate((a) => a.group.groupName.equals("Admin"))
```
And now we have another column to filter on 

| UserID (PK) | Username | GroupID (FK) | isAdmin |
|-------------|----------|--------------|--------------|
| 1           | Alice    | 1            | true | 
| 2           | Bob      | 1            | true | 
| 3           | Charlie  | 2            | false| 
| 4           | David    | 2            | false| 
| 5           | Eve      | 1            | true | 

---

### Aggregations

And what's even cooler, aggregates:
```
final userCount = db.managers.groups.annotate((a)=> a.users((a) => a.pk.count()))
```

| GroupID (PK) | GroupName | userCount |
|--------------|-----------|-----------|
| 1            | Admin     | 3 |
| 2            | User      | 2 |

We could sort on this too!

```dart
db.managers.groups.withAnnotations([userCount]).orderBy((o)=>userCount.order.desc());
```

---

I'm coming from the django world. This pattern is extremely powerful. To the extent where after 5 years of using it I didn't know SQL (until I started contributing here)

dickermoshe:
There are some issues caused by [this](https://github.com/dart-lang/dartdoc/issues/3888)

I don't want to have duplicate doccoments for `Expressions` and `AggregateBuilder`

Let see what the dart team has to say

simolus3:
I think the preview in VS Code doesn't apply templates properly, it might work when the site is actually built with `dart doc`.

dickermoshe:
You're right. `dart doc` works

I've added the comments. While I don't feel comfortable having duplicate doccomments, when this is fixed we'll go back to `@macro`

simolus3:
I had a quick look at the test only, but I think I get the idea from your comment and the docs changes. This is pretty neat!! I think it's a good way to combine the manager APIs and the existing query builder, so that users can write more complex queries easily as long as they fit into the general CRUD pattern.
Also nice that we're able to resolve joins safely thanks to the foreign key analysis we're doing on the defined schema already.

Let me know once this (or parts of the implementation) are ready to be looked at. One thing I'm generally conscious about is the increase in generated code size. For instance, both the ordering and the annotation composer seem to define the same columns and have every similar methods for returning referenced composers - is there a way to extract parts of that into a shared definition for example?

dickermoshe:
Glad to hear you like it!

I'll see what I can do about reducing the generated code without introducing any breaking changes
But remember, the smaller the generated code, the more complex the library code is. 

Just something to think about.

dickermoshe:
I thought I had a really smart way reduce code duplication, but alas, it broke the joining functionality. 🥲

IMHO, I think an 18% (avg) increase in code is reasonable for this though. People can always disable the manager code.
It may be more code, but considering what drift offers, the code generation is still pretty lean.

This is the last feature needed make the Manager API complete. No more big code generation PRs after this.

dickermoshe:
It's ready for review

docs/docs/dart_api/manager.md
```diff
@@ -118,6 +118,23 @@ Any rows that meet the specified condition will be deleted.
 
 {{ load_snippet('manager_delete','lib/snippets/dart_api/manager.dart.excerpt.json') }}
 
+
+## Annotations
+
+It's quite common that you want to perform some complex calculations you data. Drift provides an easy way to write SQL expressions which can be read and/or used in filters and orderings.
```
simolus3: This sentence is broken, and I think this feature generally deserves a better introduction. Maybe the story should be that manager queries are great when selecting whole rows + references, but sometimes its beneficial to push more logic into SQL e.g. efficient aggregates? Then we can link + mention drift's existing support for these expressions and finally present this as the way to use them with manager APIs?

docs/docs/dart_api/manager.md
```diff
@@ -118,6 +118,23 @@ Any rows that meet the specified condition will be deleted.
...
+See the [expression](expressions.md) documentation for more information on how to write expressions.
+
+{{ load_snippet('manager_annotations','lib/snippets/dart_api/manager.dart.excerpt.json') }}
```
simolus3, line 143: One thing I've noticed in this snippet in particular, not sure if it's related to the snippet builder or this snippet. The comments have [a different indentation](https://deploy-preview-3234--moor.netlify.app/dart_api/manager/#annotations) on the site than the code. Is there an easy way to fix that?

docs/lib/snippets/dart_api/manager.dart
```diff
@@ -332,3 +332,65 @@ Future<void> customOrdering(AppDatabase db) async {
   db.managers.todoItems.orderBy((f) => f.contentThenCreatedAt());
 }
 // #enddocregion manager_custom_ordering
+
+void _managerAnnotations(AppDatabase db) async {
+// #docregion manager_annotations
+// First create an annotation with an expression you want to use
```
simolus3: Oh it is the snippet haha - check the website, the docregion comment doesn't have to be indented but IMO the comments should be aligned with the line they're annotating (easier to read and makes the snippet ready to copy).

docs/lib/snippets/dart_api/manager.dart
```diff
@@ -332,3 +332,65 @@ Future<void> customOrdering(AppDatabase db) async {
...
+  final titleLengthAnnotation =
+      db.managers.todoItems.annotation((o) => o.title.length);
+
+  /// Create a copy of the manager with the annotations you want to use
+  final manager =
+      db.managers.todoItems.withAnnotations([titleLengthAnnotation]);
+
+// Then use the annotation in a filter
+// This will filter all items whose title has exactly 10 characters
```
simolus3: I think for manager users who might not be too familiar with database operations or SQL, a point we should mention is that this check happens inside the database, speeding things up by not reading rows we don't care about into Dart. (This doesn't have to be mentioned in the comments alone, can also be expanded on in the docs)

drift/lib/src/runtime/manager/annotate.dart
```diff
@@ -0,0 +1,92 @@
...
+  final Set<JoinBuilder> _joinBuilders;
+  BaseAnnotation(this._joinBuilders);
+
+  /// Create an order by clause for this annotation
+  ColumnOrderings get order {
+    return ColumnOrderings(_expression);
+  }
+}
+
+/// A class that contains the information needed to create an annotation
```
simolus3: This comment explains nothing. It should mention what an annotation is (essentially a manager wrapper around arbitrary SQL expressions with reference information resolved), probably link to the `manager/#annotation` site for details and explain why this is distinct from `BaseAnnotation` / only for columns without converters.

I know it feels tedious to write dartdocs when the main documentation is available and provides more context, but documentation comments still serve as an entrypoint for people interested in more details or those studying the implementation. So they should provide some self-contained information.

drift/lib/src/runtime/manager/annotate.dart
```diff
@@ -0,0 +1,92 @@
...
+    extends BaseAnnotation<SqlType, $Table> {
+  @override
+  final Expression<SqlType> _expression;
+
+  /// Create a filter for this annotation
+  ColumnFilters<SqlType> get filter {
+    return ColumnFilters(_expression);
+  }
+
+  /// Read the result of the annotation from the [BaseReferences] object
```
simolus3: I think this comment would benefit from a tiny snippet adding an annotation to a query and then reading it, for example. I think you can add the snippet we already have for the `withAnnotations` method.

drift/lib/src/runtime/manager/annotate.dart
```diff
@@ -0,0 +1,92 @@
...
+  @override
+  final GeneratedColumnWithTypeConverter<DartType, SqlType> _expression;
+
+  /// Create a filter for this annotation
+  ColumnWithTypeConverterFilters<DartType, DartType, SqlType> get filter {
+    return ColumnWithTypeConverterFilters(_expression);
+  }
+
+  /// Converter function to convert from [SqlType] to [DartType]
+  final DartType Function(SqlType) $converter;
```
simolus3: If this field necessary? Don't we get it from the `GeneratedColumnWithTypeConverter`?

drift/lib/src/runtime/manager/composer.dart
```diff
@@ -1,129 +1,170 @@
...
-  }
+  /// This function only removes a single join builder. This is because the join builder may have been used by other composers.
+  /// ```
+  /// db.managers.todos.filter((f) => f.category.name('Math') & f.category.id(5))
+  /// ```
+  /// In this case we don't want `.id(5)` to remove all the join builders, because it is still needed by the `.name('Math')` filter.
+  /// Therefore, we remove a single join builder.
+  ///
+  /// Don't worry, when we eventualy build the the query, the duplicate join builders will be removed.
+  late final void Function(JoinBuilder) $removeJoinBuilderFromRootComposer;
```
simolus3, line 53: Thanks for the good explanation here :+1: . I do wonder though whether this (essentially an optimization, right?) is necessary? Databases should be smart enough to ignore the join internally if we're only filtering on a column we already have without the join. Is that the only purpose or does this also optimize things internally (e.g. because we might be adding result columns for joined tables automatically without this?)

drift/lib/src/runtime/manager/composer.dart
```diff
@@ -1,129 +1,170 @@
...
+  }) : $joinBuilder = joinBuilder {
+    this.$addJoinBuilderToRootComposer = $addJoinBuilderToRootComposer ??
+        ((JoinBuilder joinBuilder) => $joinBuilders.add(joinBuilder));
+    this.$removeJoinBuilderFromRootComposer =
+        $removeJoinBuilderFromRootComposer ??
+            ((JoinBuilder joinBuilder) => $joinBuilders.remove(joinBuilder));
+  }
+
+  /// A helper method for creating a new composer from an existing composer
+  Composer.fromComposer(Composer<Database, CurrentTable> composer)
```
simolus3: This might be better named as a `clone()` method.

drift/lib/src/runtime/manager/manager.dart
```diff
@@ -287,22 +306,23 @@ class TableManagerState<
...
+        withReferenceMapper: _withReferenceMapper,
+        filter: filter,
+        joinBuilders: newJoinBuilders,
+        orderingBuilders: orderingBuilders,
+        distinct: distinct,
+        limit: limit,
+        offset: offset,
+        prefetchHooksCallback: _prefetchHooksCallback,
+        prefetchedData: _prefetchedData,
+        addedColumns: addedColumns);
```
simolus3: Lets add the trailing comma back here

drift/lib/src/runtime/manager/ordering.dart
```diff
@@ -7,19 +7,16 @@ class ColumnOrderings<T extends Object> {
   /// It's used to expose ordering functions for a column
   ///
   /// {@macro manager_internal_use_only}
-  ColumnOrderings(this.column, {this.joinBuilders});
+  ColumnOrderings(this.column);
```
simolus3, line 10: Just to be sure I understood correctly - we can remove `joinBuilders` here because they're tracked in the top-level `ProcessedTableManager` instance now, right?

simolus3 (commented):
Some initial feedback, mostly on the public/API side of things.

Also perhaps a dumb comment, but why is this named annotations? Is there a reason for not just calling theses column (e.g. with `SmartColumn` or something to distinguish them from the schema columns) If Django has established precedent here that's probably fine, but generally I'd prefer to keep the naming close to SQL.

drift/lib/src/runtime/manager/ordering.dart
dickermoshe: Sorta, it's tracked by the `Composer` instead.

### Old Way
We used to pass the joins through the callback: 
```dart
ColumnOrderings order(f) {
   final ordering = f.id.asc();
   print(ordering.joinBuilders) // JoinBuilders are in this `ordering`
   return ordering;
}
```
### New Way
Now we store it in `f`
```dart
ColumnOrderings order(f) {
   final ordering = f.id.asc();
   print(f.joinBuilders); // JoinBuilders are in this `f` Composer
   return ordering;
}
```
So Internally the building looks like:
```dart
// Create a brand new empty composer
final composer = createNewComposer();
print(composer.joinBuilders) // [] - It's empty

// Build the ordering, this will also add some joins the the composer
final ordering = order(composer);
print(composer.joinBuilders) // [JoinBuilder(...)] - Now it's full
```

drift/lib/src/runtime/manager/annotate.dart
dickermoshe: Good point!

drift/lib/src/runtime/manager/composer.dart
dickermoshe: sqlite doesn't optimize for this. This is because it's possible that `todo.category` is pointing at a category which doesn't exist (e.g. `PRAGMA foreign_keys = OFF;`). So if sqlite rewrote the query and removed the join, you _technically_ would get different results.

docs/lib/snippets/dart_api/manager.dart
dickermoshe: I'll add it to the docs, that a great point!

docs/docs/dart_api/manager.md
dickermoshe: I lan on rewriting most of the docs and giving annotations its own page, so i didn't put much thought into these temporary docs.

But i completely agree, I'll update it

dickermoshe:
>Also perhaps a dumb comment, but why is this named annotations? Is there a reason for not just calling theses column (e.g. with SmartColumn or something to distinguish them from the schema columns) If Django has established precedent here that's probably fine, but generally I'd prefer to keep the naming close to SQL.

This is a solid point. Not super stoked about `SmartColumn`. Let's sleep on this

drift/lib/src/runtime/manager/annotate.dart
dickermoshe: Great point. Updated

drift/lib/src/runtime/manager/annotate.dart
dickermoshe: Done

dickermoshe:
Drizzle: custom fields
Prisma: computed fields
Django: annotation

I think `computed fields` is the most self explanatory even if it's not a "sql" term.

What do you think?

dickermoshe:
For another PR:

### `groupBy`

```dart
final categoryFirstLetter = db.managers.category.annotate((a) =>a.name.substr(1,1));
Map<String,Category> categoryByLetter = db.managers.category.groupBy(categoryFirstLetter)
```

### `aggregate`

```dart
final count = db.managers.category.annotate((a) =>a.id.count());
int categoryCount = db.managers.category.aggregate(count)
```

dickermoshe:
I renamed it computedFields, what do you think?

drift/lib/src/runtime/manager/composer.dart
simolus3: Oh right good catch! But the different semantics is something worth mentioning, I think. Personally, I read `todos.filter((f) => f.category.id(5))`  as "todos with an associated category having the id 5". If no such category exists, I'd expect no results to be returned (and hopefully users won't start relying on this...)
It can be surprising that `f.categoryId(5)` and `f.category.id(5)` are not equivalent, but I think around broken references, that's still an important distinction.

We can leave it as is, but let's publicly document this behavior around references somewhere.

simolus3 (approved):
I agree with computed fields as a name, thanks :+1: 

I think the "removing join optimizations" should be documented somewhere to avoid the surprise of users getting unexpected results with broken references (although that should be a rarity). Apart from that, this looks good to me.

dickermoshe:
The reason why I have it work this way is because there is no column named
“categoryId” just “category”. And I didn’t want to have “category” point to
the column itself and have ANOTHER field named “categoryObj” point to the
reference object.

This way there is just a single field that fulfills 95% of what is needed.

Issue #3219, opened 2024-09-14
Exported schema do not contain column `check()` constraints


coffeebeats:
Hey! Thanks for creating a fantastic library. I've found what I think is a bug, which I'm documenting below:

**Describe the bug**

It appears that schema exports do not include `check()` column constraints. This results in errors like the following, where schema migration tests pass but live database migrations fail `validateDatabaseSchema()`:

```
Schema does not match
  users:
   columns:
    creation_time:
     Not equal: `NOT NULL DEFAULT 1704096000` (expected) and `NOT NULL DEFAULT 1704096000 CHECK(NOT 1 AND "creation_time" IS NULL OR 1 AND "creation_time" IS NOT NULL AND "creation_time" > 0)` (actual)
```

In the error we see that the expected value, which is derived from the exported schema, does not include the `check()` constraint on the actual column (derived from the migrator's table information).

I've created a minimum reproducible example at [coffeebeats/example-drift-export-column-check-bug](https://github.com/coffeebeats/example-drift-export-column-check-bug). To see the error in action (and how migration tests hide this issue), clone the repository and follow the reproduction steps.

Please let me know if I can provide any other information. Thanks!

simolus3:
Thank you for this stellar bug report! The underlying problem here is that `drift_dev schema dump` only runs static analysis to infer the schema of your database. Some features (most notably views, but also some others, like `check()`) can't be statically checked however, they require running user code.

I've changed `drift_dev` in 9e3a3c26d279237df3b97bc6f3d039182a5083c0 to remember the Dart code passed to `check()` constraints. Unfortunately, that's not enough to fix this because the `creationTime` self-reference in the column definition does not work with the compact structure we generate for step-by-step migrations. I've confirmed that it works with this after the commit:

```patch
diff --git a/lib/tables.dart b/lib/tables.dart
index acb9acc..e42a824 100644
--- a/lib/tables.dart
+++ b/lib/tables.dart
@@ -10,6 +10,7 @@ class Users extends Table {
   // Column added in version 11
   // Column that can only be set to times after 1950.
   DateTimeColumn get creationTime => dateTime()
-      .check(creationTime.isBiggerThan(Constant(DateTime(2020))))
+      .check(CustomExpression<DateTime>('creation_time')
+          .isBiggerThan(Constant(DateTime(2020))))
       .withDefault(Constant(DateTime(2024, 1, 1)))();
 }
```

[...]

PR #3250, opened 2024-09-30
feat: all in one migration command

### LINKED ISSUE #3216: Generated Tests
### Testing Migrations

Tests like [these](https://drift.simonbinder.eu/docs/testing/#writing-tests) should be generated automatically.

dickermoshe:
Generates Dumped Schema, Tests and Steps with a single command.
Uses `build.yaml` config to define database and file paths.

Fixes: https://github.com/simolus3/drift/issues/3216

This basically wraps `generate`,`dump and `steps`.
A small refactor is needed to make the code in each of those commands reusable by this new command.


simolus3 (commented):
I was worried about supporting multiple databases, but splitting generated migration assets into different directories is a clever solution.
The implementation looks good to me, some things I'm wondering from a DX perspective:

1. Most users will have exactly one database - I think we should make that case a little easier by also allowing a simple `String` (as a file path) in the `databases` option (or perhaps with a separate `database` key).
2. Maybe we could use the entries of that map to not only configure the name of the database (which doesn't appear to be considered for file paths) but also the location under `test/` where the default test should be added to?
3. I kind of view this as a "make the first migration tooling setup easier" command. In particular, I imagine that users will want to use the output of `writeGeneratedTest()` as a starting point to add own migration tests too. So maybe we shouldn't overwrite the file every time and instead only write it if it doesn't exist?

dickermoshe:
1 will require fighting with freezed. Not sure how to do that? sealed unions?
2 & 3 are already in the works.

simolus3:
> 1 will require fighting with freezed

We're using json serializable. I think you can keep the field for this and then use `@JsonKey` with a `fromJson` annotation to keep maps unchanged and map and strings to a single-element map. Don't spend too much time on this, I can figure it out if needed.

dickermoshe:
We must resolve https://github.com/simolus3/drift/issues/3219#issuecomment-2384651922

dickermoshe:
3 I will generate a 2nd file for them to fill in if they want to do manual tests

dickermoshe:
@simolus3 How do you suppose I write tests for all this.

Most of this functionality is copied from `generate`, `steps` and `dump`.
The only additional feature is the data validation, which is just a scaffold users fill with before and after data.

dickermoshe:
@simolus3 
Read for review.
Tell me what you think of this and what you would like changed

dickermoshe:
@simolus3 
Read for review. Tests are all written, lmk if you think I've missed something.

I did not implement point 1 myself. Left for you to implement - it's likely quicker for you given your familiarity with json_serializable.

Regarding point 3, we could put a function in the generated `v_toV_.dart` files which are not overwritten.
What do you think this function would look like and when should it be ran?

Overall, I this is coming together much quicker than I would like.
This is a tool I would personally find very useful, however it's possible that I overlooked something.

---

Please try this out yourself in practice and Lmk what the user experience is like.
It's hard to put myself in the package users' shoes. As a maintainer, I might be missing things that are obvious to someone using the package.

drift_dev/lib/src/cli/commands/make_migrations.dart
```diff
@@ -0,0 +1,505 @@
...
+import 'package:path/path.dart' as p;
+
+class MakeMigrationCommand extends DriftCommand {
+  MakeMigrationCommand(super.cli);
+
+  @override
+  String get description => """
+Generates migrations utilities for drift databases
+
+### Usage
```
simolus3: We can't really do markdown on the CLI. You could try making it bold, but most of this command should be explained on the website as well.

drift_dev/lib/src/cli/commands/make_migrations.dart
```diff
@@ -0,0 +1,505 @@
...
+After defining your database for the first time, run this command to save the schema.
+When you are ready to make changes to the database, alter the schema in the database file, bump the schema version and run this command again.
+This will generate the following:
+
+1. A steps file which contains a helper function to write a migration from one version to another.
+
+  Example:
```
simolus3: Remove the trailing space here and in line 49:

```suggestion
  Example:
`` `

File: drift_dev/lib/src/cli/commands/make_migrations.dart [hunk 3]
```diff
@@ -0,0 +1,505 @@
...
+
+    /// The root directory where schema files for all databases are stored
+    /// e.g /drift_schemas/
+    final rootTestDir = Directory(
+        p.join(cli.project.directory.path, cli.project.options.testDir))
+      ..createSync(recursive: true);
+
+    if (cli.project.options.databases.isEmpty) {
+      cli.logger.info(
+          'No databases found in the build.yaml file. Check here to see how to add a database TODO: ADD LINK');
```
simolus3: Showing how to configure this tool should be part of the documentation, and then we can link to the relevant section here :)

drift_dev/lib/src/cli/commands/make_migrations.dart
```diff
@@ -0,0 +1,505 @@
...
+      exit(0);
```
simolus3: Don't exit with 0 here, it's not a successful command invocation. Use either 1 or 64 for usage errors.

drift_dev/lib/src/cli/commands/make_migrations.dart
```diff
@@ -0,0 +1,505 @@
...
+            dbName: entry.key,
+            relativeDbClassPath: entry.value);
+        return writer;
+      },
+    ));
+
+    for (var writer in databaseMigrationsWriters) {
+      // Dump the schema files for all databases
+      await writer.writeSchemaFile();
+      if (writer.schemaVersion == 1) {
```
simolus3: Checking for `writer.schemas.length > 1`would be more accurate I think.

drift_dev/lib/src/cli/commands/make_migrations.dart
```diff
@@ -0,0 +1,505 @@
...
+      return result;
+    }
+    final versions = schemas.keys.toList()..sort();
+    for (var i = 0; i < versions.length - 1; i++) {
+      final (from, fromSchema) = (versions[i], schemas[versions[i]]!);
+      final (to, toSchema) = (versions[i + 1], schemas[versions[i + 1]]!);
+      final fromTables = fromSchema.schema.whereType<DriftTable>();
+      final toTables = toSchema.schema.whereType<DriftTable>();
+      final commonTables = fromTables.where(
+          (table) => toTables.any((t) => t.schemaName == table.schemaName));
```
simolus3, line 445: Why are we restricting schema elements to tables here?

drift_dev/test/cli/make_migrations_test.dart
```diff
@@ -0,0 +1,193 @@
...
+          fail('Expected an error');
+        } catch (e) {
+          expect(e, isA<FatalToolError>());
+        }
+
+        // Change the db schema and bump the version
+        File(p.join(project.root.path, 'lib/db.dart'))
+            .writeAsStringSync(_dbWithNewColumnBump);
+        await project.runDriftCli(['make-migrations']);
+        expect(
```
simolus3, line 60: `package:test_descriptor` actually has a `validate` method that can validate entire directories (or at least multiple) files at once, which might make this test easier to understand. Can you check whether the validation part can be written [like in the example here](https://pub.dev/packages/test_descriptor#usage)?

drift_dev/lib/api/migrations.dart
```diff
@@ -214,3 +214,33 @@ class InitializedSchema {
...
+/// 1. Create a database at a specific version
+/// 2. Insert data into the database
+/// 3. Migrate the database to a target version
+/// 4. Validate that the data is valid after the migration
+Future<void> testStepByStepigrations<OldDatabase extends GeneratedDatabase,
+        NewDatabase extends GeneratedDatabase>(
+    {required SchemaVerifier verifier,
+    required OldDatabase Function(QueryExecutor) oldDbCallback,
+    required NewDatabase Function(QueryExecutor) newDbCallback,
+    required GeneratedDatabase Function(QueryExecutor) currentDbCallback,
```
simolus3: Let's make these names a little more descriptive. `oldDbCallback` -> `createOld`; `newDbCallback` -> `createNew`;  `currentDbCallback` -> `openTestedDatabase`?

simolus3 (commented):
This generally looks good to me. Some bigger notes are:

1. We should adopt this in the `example/migrations` project (if necessary, copy the existing schema dumps into the directory where this tool would expect them). Then let's add a section at the top of the readme showing how to use the new all-in-one command. We can keep the older commands as a reference for users who need manual control.
2. Since we expect that most users will want to run this command instead of the previous multiple invocations, let's highlight it more prominently. I think it deserves to be at the top of the migrations tooling page, mentioning that it does everything the other commands are doing in a single step, and that the other commands are then mostly intended for users needing more customization / integration into other tools.
3. Instead of pointing to `--help` on the documentation website, I think the website should give an introduction into the command or at least point to the migrations example. The colorful usage output looks great, but users will probably expect this information on the site as well.

drift_dev/lib/src/cli/commands/make_migrations.dart
dickermoshe: What else is insertable other than table rows?
I'm even realizing that custom row classes may be an issue if they don't implement `Insertable`, how do you suppose we deal with that?

dickermoshe:
Agree, moved around some around of the docs.

drift_dev/lib/src/cli/commands/make_migrations.dart
dickermoshe: never mind, seems it works fine with custom row classes, but what else should be inserted that we should validate?

dickermoshe:
I'm gonna do the migration docs rewrite in this PR

drift_dev/lib/src/cli/commands/make_migrations.dart
dickermoshe: _ping_ 😏

simolus3: Ah sorry, I thought this was responsible for also generating the schema code somehow.

The name and dartdocs of this class are a bit confusing since it essentially emits a test file for a migration? Maybe we should call it `_MigrationTestEmitter` or something, and mention that it's responsible for emitting a default test file verifying data integrity.

drift_dev/lib/api/migrations.dart
```diff
@@ -214,3 +214,33 @@ class InitializedSchema {
...
 }
+
+/// Utility function used by generated tests to:
+/// 1. Create a database at a specific version
+/// 2. Insert data into the database
+/// 3. Migrate the database to a target version
+/// 4. Validate that the data is valid after the migration
+Future<void> testStepByStepigrations<OldDatabase extends GeneratedDatabase,
+        NewDatabase extends GeneratedDatabase>(
+    {required SchemaVerifier verifier,
```
simolus3: I wonder if this should be an instance method on `SchemaVerifier`?

drift_dev/lib/api/migrations.dart
```diff
@@ -214,3 +214,33 @@ class InitializedSchema {
   /// ```
   DatabaseConnection newConnection() => _createConnection();
 }
+
+/// Utility function used by generated tests to:
+/// 1. Create a database at a specific version
+/// 2. Insert data into the database
+/// 3. Migrate the database to a target version
+/// 4. Validate that the data is valid after the migration
+Future<void> testStepByStepigrations<OldDatabase extends GeneratedDatabase,
```
simolus3: typo, missing `M`

examples/migrations_example/test/drift/default/migration_test.dart
```diff
@@ -0,0 +1,204 @@
...
+);
+
+test(
+  "default - migrate from v6 to v7",
+  () => testStepByStepigrations(
+    from: 6, to: 7, verifier: verifier, createOld: (e) => v6.DatabaseAtV6(e),
+    createNew: (e) => v7.DatabaseAtV7(e), openTestedDatabase: (e) => Database(e),
+    createItems: (b, oldDb) {
+      b.insertAll(oldDb.users, v6_to_v7.usersV6);
+b.insertAll(oldDb.groups, v6_to_v7.groupsV6);
```
simolus3: Just noticed this - please run `DartFormatter` in `drift_dev` before emitting code, all generated files must be formatted.

examples/migrations_example/test/drift/default/migration_test.dart
```diff
@@ -0,0 +1,204 @@
...
+    verifier = SchemaVerifier(GeneratedHelper());
+  });
+
+  test(
+  "default - migrate from v1 to v2",
+  () => testStepByStepigrations(
+    from: 1, to: 2, verifier: verifier, createOld: (e) => v1.DatabaseAtV1(e),
+    createNew: (e) => v2.DatabaseAtV2(e), openTestedDatabase: (e) => Database(e),
+    createItems: (b, oldDb) {
+      b.insertAll(oldDb.users, v1_to_v2.usersV1);
```
simolus3: Doesn't this have to be async?

examples/migrations_example/test/drift/default/validation/v10_to_v11.dart
```diff
@@ -0,0 +1,17 @@
+import '../schemas/schema_v10.dart' as v10;
+import '../schemas/schema_v11.dart' as v11;
+
+/// Run `dart run drift_dev make-migrations --help` for more information
+
+final usersV10 = <v10.UsersData>[];
```
simolus3: I'm not too sure about generating this into separate files. It adds a lot of new files, and users are expected to modify them either way right?

I don't see a strong advantage of these files over just inlining them in the generated example tests? E.g. what if we created

```dart

test("default - migrate from v10 to v11", () async {
  final oldUsers = [...]
  final migratedUsers = [...]

  // Now invoke test utility here
});
```

Is the problem that we'd then have to patch test files whereas we can emit these files once and ignore them if they already exists?
Since we can't generate meaningful examples to test consistency automatically, I don't think this is really worth it. I think I prefer the approach previously used in the migrations example which:

1. Tests all combinations of schema versions without data integrity checks.
2. For the migrations that are critical because they alter tables, write custom tests verifying data integrity.

I agree with the utility in `drift_dev` to make step 2 easier, but it can't be fully automated and we shouldn't try to. Maybe we can generate a single test file as an example if it doesn't already exist and then encourage users to write their migration tests in that file? It could contain these tests (with inlined versioned data lists for data integrity checks) as an example so that users know how to write them.

simolus3 (commented):
Thanks for updating the documentaiton, and sorry for the delay responding. Looking at these changes applied to the migrations example, I'm not too convinced about the value these semi-automated data integrity tests add since we can't add meaningful data automatically.

What if the tool generates a single test file if it doesn't exist already with a semi-useful example as a starting point (think e.g. the default test created by `flutter create`) that shows how migration tests could look like with the generated database, but leaves the decision which migrations need data integrity tests to the user?

examples/migrations_example/test/drift/default/validation/v10_to_v11.dart
dickermoshe: > I'm not too sure about generating this into separate files. It adds a lot of new files, and users are expected to modify them either way right?

They don't HAVE to do anything. It's completely optional.

> Is the problem that we'd then have to patch test files whereas we can emit these files once and ignore them if they already exists?

Exactly, I don't want to overwrite existing tests.

We could do this instead
This is generated once and never overwritten.

validate.dart
```dart
import "../schema_v1.dart" as v1;

final validation = Validate(
  v1toV2: (
    v1: V1Models(
      users: [v1.User(...)]
    ),
    v2: V2Models(
      users: [v2.User(...)]
    ),
  ),
  v2toV3: (
      v2: V2Models(
        users: [v2.User(...)]
      ),
      v3: V3Models(
        users: [v3.User(...)]
      ),
    )
);
```



The Validate class would be implemented in another file and would be overwritten on each run:
```dart
class Validate{
  final ({V1Models v1,V2Models v2}) v1toV2; 
  final ({V2Models v2,V3Models v3}) v2toV3; 
  //...
}
```


The downside is that users need to write the import themselves, which is ugly.
```dart
import "../schema_v1.dart" as v1;
```


> I agree with the utility in drift_dev to make step 2 easier, but it can't be fully automated and we shouldn't try to.

I hear your point, a random file with some lists feels weird. The above setup not as much, but still funny.

examples/migrations_example/test/drift/default/migration_test.dart
dickermoshe: No
image

drift_dev/lib/api/migrations.dart
dickermoshe: duh

drift_dev/lib/api/migrations.dart
dickermoshe: Just say the word...

drift_dev/lib/src/cli/commands/make_migrations.dart
dickermoshe: Great point, renamed

drift_dev/lib/api/migrations.dart
simolus3: Let's make it an instance method :)

simolus3: Also I think it should be called `testWithDataIntegrity` or something since it's not really related to step-by-step migrations.

examples/migrations_example/test/drift/default/validation/v10_to_v11.dart
simolus3: My point is that most migrations likely don't need a test verifying data integrity. Most migrations add new tables and columns, users know that and they know that inserting rows at the start and checking them in the end is a lot of effort that is just not needed for those tests.
In my experience, "dangerous" migrations that change column constraints or types in a way that could affect existing data are not so common. So my suggestion is that we don't generate these validation utilities for every migration. Still, it's important that we inform users that verification tests are possible and sometimes really helpful.

So, I think we should generate a single test file (if it doesn't exist already) that:

1. Uses the for-loop pattern previously shown in the migration example to test every possible migration sequence (this likely doesn't need customization for users, these tests are super simple but can still catch issues reliably. Basically the first 20% of effort that does 80% of the work).
2. Then, for any migration we know of (can just be the first one), sets up another test that looks like

```dart
test('migration from v1 to v2 does not corrupt data', () async {
  // TODO: Consider writing these kinds of tests when altering tables in a way that might affect existing rows.
  // The default migration tests run with an empty schema, so it's a recommended practice to also test with
  // data for relevant migrations.
  final oldData = [];  // TODO: Add expected data at version 1 using v1.User()
  final expectedNewData = []; // TODO: Add expected data at version 2 using v2.User()

  await verifier.testWithDataIntegrity(
    // parameters filled out
  )
});
```

(where `User` is replaced with a shared data class name for the schema versions)

We don't have to re-generate that test or the expected data ever. We've taught users about the importance of testing with data, now we have to hope that they will actually follow that advice when needed. That's the best we can do either way since users need to fill in the rows. I don't see how the `validate.dart` or the other additional files add enough value to be worth the clutter. Even with the manual import management I think it's still the better solution.

docs/docs/schema.md
```diff
@@ -0,0 +1,509 @@
+---
+
+title: Tables
+description: Define the schema of your database.
+
+---
+
+# Tables
```
simolus3: Did you mean to add this to the docs restructuring PR instead?

drift_dev/lib/api/migrations.dart
dickermoshe: Should I make this disable foreign keys and re-enable if they were enabled?
Edit: Wait, is there no way to run a custom statement and get a dynamic value?
e.g `PRAGMA foreign_keys;` > 0

examples/migrations_example/test/drift/default/validation/v10_to_v11.dart
dickermoshe: I completely agree!
This is a great idea.


> Basically the first 20% of effort that does 80% of the work

Solid Point

> Even with the manual import management I think it's still the better solution.

We could open the file and inject the imports to the top of the file.

dickermoshe: This is what I have now.
Modify this to your liking and that's what I'll do:



```dart
// ignore_for_file: unused_local_variable, unused_import
import 'package:drift_dev/api/migrations.dart';
import 'package:migrations_example/database.dart';
import 'generated/schema.dart';

import 'generated/schema_v1.dart' as v1;
import 'generated/schema_v2.dart' as v2;

void main() {
  driftRuntimeOptions.dontWarnAboutMultipleDatabases = true;
  late SchemaVerifier verifier;

  setUpAll(() {
    verifier = SchemaVerifier(GeneratedHelper());
  });

  group('default database', () {
    //////////////////////////////////////////////////////////////////////////////
    ////////////////////// GENERATED TESTS - DO NOT MODIFY ///////////////////////
    //////////////////////////////////////////////////////////////////////////////
    if (GeneratedHelper.versions.length < 2) return;
    for (var i
        in List.generate(GeneratedHelper.versions.length - 1, (i) => i)) {
      final oldVersion = GeneratedHelper.versions.elementAt(i);
      final newVersion = GeneratedHelper.versions.elementAt(i + 1);
      test("migrate from v$oldVersion to v$newVersion", () async {
        final schema = await verifier.schemaAt(oldVersion);
        final db = Database(schema.newConnection());
        await verifier.migrateAndValidate(db, newVersion);
        await db.close();
      });
    }
    //////////////////////////////////////////////////////////////////////////////
    /////////////////////// END OF GENERATED TESTS ///////////////////////////////
    //////////////////////////////////////////////////////////////////////////////

    /// Write data integrity tests for migrations that modify existing tables.
    /// These tests are important because the auto-generated tests only check empty schemas.
    /// Testing with actual data helps ensure migrations don't corrupt existing information.
    ///
    /// The following is an example of how to write such a test:
    test("migration from v1 to v2 does not corrupt data", () async {
      final oldUsersData = <v1
          .UsersData>[]; // TODO: Add expected data at version 1 using v1.UsersData
      final expectedNewUsersData = <v2
          .UsersData>[]; // TODO: Add expected data at version 2 using v2.UsersData

      await verifier.testWithDataIntegrity(
        oldVersion: 1,
        newVersion: 2,
        verifier: verifier,
        createOld: (e) => v1.DatabaseAtV1(e),
        createNew: (e) => v2.DatabaseAtV2(e),
        openTestedDatabase: (e) => Database(e),
        createItems: (batch, oldDb) {
          batch.insertAll(oldDb.users, oldUsersData);
        },
        validateItems: (newDb) async {
          expect(expectedNewUsersData, await newDb.select(newDb.users).get());
        },
      );
    });

    /// Add additional data integrity tests here
  });
}

```

drift_dev/lib/api/migrations.dart
simolus3: > Wait, is there no way to run a custom statement and get a dynamic value?

You could run these pragmas as a `customSelect`.

> Should I make this disable foreign keys and re-enable if they were enabled?

IIRC we're not doing this for regular migrations either right? If users disable foreign keys in their migrator ([possible with step-by-step migrations too](https://drift.simonbinder.eu/docs/migrations/step_by_step/#customizing-step-by-step-migrations)), that should work already.

Generally, the behavior in migration tests  should match that of real databases as closely as possible.

examples/migrations_example/test/drift/default/validation/v10_to_v11.dart
simolus3: Looks good, I've reworded some of the comments to make them fit into 80 char lines, but that template is exactly what I had in mind :+1:

simolus3 (approved):
Thanks for the great work! LGTM.

I'll just try to get the CI fixed :D

PR #3281, opened 2024-10-11
Linter for drift

dickermoshe:
_Dunno what happened to the old PR. I think the VScode Github Plugin must've closed it. Anyway, here's it as a new PR_
- [ ] reference latest schema in migration instead of step by step one
- [x] `createReturning` with ignore mode
- [x] offset without limit
- [x] await all migrations
~~- [ ] Manager report issues~~


dickermoshe:
@simolus3

See how I setup the tests. Lmk if you have a better way of setting them up.

Also, Some of the async code is copied from the dart sdk, I don't know how this affects licensing.

drift_dev/lib/src/lints/unawaited_futures_in_transaction.dart
```diff
@@ -0,0 +1,156 @@
...
+    });
+    context.registry.addInterpolationExpression((node) {
+      node.accept(_Visitor(this, reporter, _code));
+    });
+  }
+}
+
+// Copyright (c) 2016, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
```
simolus3: Let's also link to the unawaited futures link source where I guess this is coming from

drift_dev/test/linter_test/linter_test.dart
```diff
@@ -0,0 +1,20 @@
...
+
+void main() {
+  // Test the linter_test.dart file
+  test('linter', () async {
+    final workingDir = p.join(p.current, 'test/linter_test/pkg');
+    expect(
+        await Process.run('dart', ['pub', 'get'], workingDirectory: workingDir)
+            .then((v) => v.exitCode),
+        0);
+    expect(
```
simolus3: We'll also need examples that do have lints, right?

I would suggest:

1. Having testcases defined as strings here
2. Then creating a temporary directory to run `custom_lint` on test code. That directory can be created using `package:test_descriptor`.
3. Instead of running `pub get` which is quite slow, you can make use of the fact that the dependencies including `custom_lint` have already been resolved when tests are running an create a stub package config like this: https://github.com/simolus3/drift/blob/d7cc547836cdce038f76395908e137afd96cbc96/drift_dev/test/analysis/test_utils.dart#L115-L130 That config needs to be in `.dart_tool/package_config.json` of the temporary directory - that should allow running `custom_lint`.

Let me know if there's anything I can help with here

simolus3 (commented):
Some comments on the test infrastructure

drift_dev/test/linter_test/linter_test.dart
dickermoshe: `// expect_lint: drift_build_errors`
Will throw a lint error is an error is NOT found. It's tests the lint.

dickermoshe: It needs to hold the built code too. If we do it with strings, it will look like a mess.
We will have to store `db.dart` and `db.g.dart` in a single file as a muti-line string.

Riverpod lint has no tests, the test is another package next to it which just acts as a test:
https://github.com/rrousselGit/riverpod/tree/master/packages

drift_dev/lib/src/lints/unawaited_futures_in_transaction.dart
dickermoshe: Done

drift_dev/test/linter_test/linter_test.dart
dickermoshe: Let me know if you still want me to have the tests in strings.

dickermoshe:
Right now warnings in manager generation are not reported in the resolver, so the linter can't use them

The warning I'm focusing on are duplicate reverse names e.g.

```dart
class User extends Table {
  late final group = integer().references(Group, #id)
  late final supergroup = integer().references(Group, #id)
}
class Group extends Table {
//   `userRefs` has 2 relations!!!
}
```

How would I go about this?

dickermoshe:
I add a warning level to `DriftAnalysisError` to show orange squiggles instead of red for less severe errors

drift_dev/test/linter_test/linter_test.dart
simolus3: A separate package / directory to hold test examples is fine in that case :+1:

simolus3:
> Right now warnings in manager generation are not reported in the resolver, so the linter can't use them

Ideally, these warnings should be discovered in the analysis step instead of during code generation. I know that some of them are easier to report during codegen, is it a lot of work to port them over to something that could run in `FileAnalyzer`? The parts under `if (element is BaseDriftAccessor)` should have the same information available to them as the generator, especially in the end when all transitive tables have been added). I can also get started on this to set up the basic infrastructure for those warnings if that works for you.

dickermoshe:
Yeah, could you please do that?
I tried myself, but after an hour I gave up.

dickermoshe:
@simolus3 Any update on this.
Just need you to write some scaffolding where to do these analysis.

dickermoshe:
In FileAnalyzer I don't have access to the base element to report the error

dickermoshe:
Ping

simolus3:
I had this commit laying around: https://github.com/simolus3/drift/pull/3281/commits/b64d6023d4a4b8e19cd9503d04c00b8d71cad905#diff-d36a7b426417ff349d59d9768776e3b2d36ff88fcafe18d7f82c086db523c55d

I tried doing a rebase, but I appear to have messed up. I'm not sure why there are so many conflicts in the docs though...
I'll take another look at providing you with the correct definition spans though. You are essentially operating on a table model, right? Ideally we'd want to the same lints to work regardless of whether the table is defined in Dart or in a drift file, so I suppose you need a way to go from `DriftColumn`  to the defining source span? Let me know if I'm overlooking something here, otherwise I can work on passing that information through the analyzer.

dickermoshe:
drift_lint will only report IDE lints for `.dart`
Stated a discussion [here](https://github.com/invertase/dart_custom_lint/discussions/297) to confirm that.

However, the custom_lint cli will still report errors so I guess it could be helpful

dickermoshe:
At this point there is only a single lint which I need passed through all these layers.
Completely upending the entire analysis just for this would not be worth the effort

dickermoshe:
All done. I'll add docs once you give me the go ahead

dickermoshe:
@simolus3 Yup, we def got a flaky test

simolus3:
On the test, I blame the Dart update :D

drift_dev/lib/drift_dev.dart
```diff
@@ -0,0 +1,7 @@
+import 'package:custom_lint_builder/custom_lint_builder.dart';
+import 'package:drift_dev/src/lints/custom_lint_plugin.dart';
+
+/// This function is automaticly recognized by custom_lint to include this drift_dev package as a linter
+PluginBase createPlugin() {
```
simolus3, line 5: Not a huge fan of this, is there a way to tell `custom_lint` to use a different import for a package providing linters (my preferred library would be `lib/integrations/custom_lint.dart`).

But if that's not possible so be it, I can file an issue on `custom_lint` and we can leave this as-is until that gets resolved.

drift_dev/test/lint/test_pkg/pubspec.yaml
```diff
@@ -0,0 +1,24 @@
+name: test_pkg
+description: A starting point for Dart libraries or applications.
+version: 1.0.0
```
simolus3, line 3:
```suggestion
version: 1.0.0
publish_to: none
```

pubspec.yaml
```diff
@@ -5,4 +5,4 @@ environment:
   sdk: ">=3.3.0 <4.0.0"
 
 dev_dependencies:
-  melos: ^3.0.0
+  melos: ^6.1.0
```
simolus3, line 8: Oof, good catch 👍

simolus3 (commented):
Looks good to me, thanks! I hope we don't need too much documentation for this because this should just work for users already using `custom_lint`.

I imagine a simple page in the CLI directory noting that the feature exists and what kind of errors are generally reported would be helpful. We should also link to that page from the [analyze command docs](https://drift.simonbinder.eu/Tools/#analyze).

dickermoshe:
??

docs/docs/setup.md
```diff
@@ -34,12 +34,13 @@ adding a package to open database on the respective platform.
     dev_dependencies:
       drift_dev: ^{{ versions.drift_dev }}
       build_runner: ^{{ versions.build_runner }}
+      custom_lint: ^{{ versions.custom_lint }}
```
simolus3, line 37: I think we should have a small section somewhere mentioning this explicitly (it can also be linked at the bottom of this page).

I'm wary about mentioning the `custom_lint` dependency here. We're already suggesting users install a lot of dependencies.
IMO, we should:

1. Mention somewhere that if users already have `custom_lint`, drift will work seamlessly with that and provide its own lints.
2. Suggest (but less strongly than done here, essentially pretending `custom_lint` is a requirement) that users try `custom_lint` to get quicker feedback about potential issues related to drift.

I think the best way to do that would be to mention this on [/Tools](https://drift.simonbinder.eu/Tools/) and then add a bullet point at the end of this page next to the others.

simolus3 (commented):
Sorry for the delay (again...) I can take a final look later today and merge this if it's ready from your side.

I'll probably do a patch release before merging this (there are some bugfixes waiting to be released), that way we can ship this in a minor version and I'll have more time to test this.

dickermoshe:
Ill work on the docs.
Also seems CI is failing, I'll ping you when it's all ready

PR #3290, opened 2024-10-16
New schema docs

dickermoshe:
This PR:

- Restructures the layout of the site. This does not affect urls.
- Remove Type Converter and Custom row class docs. Move into table and dataclass docs respectively
- Add redirect for removed ages
- rewrite table docs
- Move a whole bunch of stuff into guides
- Add missing CSS style


docs/docs/_redirects
```diff
@@ -53,4 +53,6 @@
 /docs/platforms/postgres/ /Platforms/postgres/
 /docs/advanced-features/daos/ /dart_api/daos/
 /api/* https://pub.dev/documentation/drift/latest/index.html
+https://drift.simonbinder.eu/custom_row_classes/ /dart_api/dataclass/
+https://drift.simonbinder.eu/type_converters/ /dart_api/tables/#custom-types
```
simolus3: Shouldn't these be without the domain like the others above?

docs/docs/dart_api/dataclass.md
```diff
@@ -1,23 +1,112 @@
...
+    - Example: `UserInfo` -> `UserInfoData`
+
+
+To use a custom name use the `@DataClassName` annotation.
+
+**Example:**
+
+{{ load_snippet('data-class-name','lib/snippets/dart_api/dataclass.dart.excerpt.json') }}
+
+## Json serialization
```
simolus3: I think this should not be this far up  - and there should be a little warning saying that using custom row classes is the preferred approach for JSON serialization.

docs/docs/dart_api/dataclass.md
```diff
@@ -1,23 +1,112 @@
 ---
 
-title: Custom row classes
-description: Use your own classes as data classes for drift tables
+title: Dataclass
```
simolus3: I'm still not too happy with the generic `Dataclass` name, does "Generated classes" with a description like "Overview of classes generated by drift for reading and writing to the database" work?

docs/docs/dart_api/dataclass.md
```diff
@@ -1,23 +1,112 @@
...
+{
+  "id": 1,
+  "title": "Todo 1",
+  "created_at": "2024-02-29T12:00:00Z"
+}
+```
+
+### Custom json keys
+
+To use a custom name for JSON serialization, use the `@JsonKey` annotation.
```
simolus3:
```suggestion
To use a custom name for JSON serialization, use the `@JsonKey` annotation.
Note that the `@JsonKey` class from `package:drift` is not same as the `@JsonKey` annotation from `package:json_annotation`, and the two are not compatible with each other.
```

docs/docs/dart_api/dataclass.md
```diff
@@ -1,23 +1,112 @@
...
+```
+For more details on customizing column names in SQL, refer to the [column name](tables.md#column-names) documentation.
+
+## Companions
+
+In addition to the generated dataclass representing a complete row, Drift also generates a companion object for each table, which represents a partial row and can be used to update existing rows.
+
+<div class="annotate" markdown>
+
+{{ load_snippet('generated-companion','lib/snippets/dart_api/dataclass.dart.excerpt.json') }}
```
simolus3: I think we should use a tabbed view here with Manager vs. the Core query builder instead of having it in a single snippet.

docs/docs/dart_api/dataclass.md
```diff
@@ -1,23 +1,112 @@
...
+    builders:
+      drift_dev:
+        options:
+          use_sql_column_name_as_json_key : true
+```
+For more details on customizing column names in SQL, refer to the [column name](tables.md#column-names) documentation.
+
+## Companions
+
+In addition to the generated dataclass representing a complete row, Drift also generates a companion object for each table, which represents a partial row and can be used to update existing rows.
```
simolus3: I think this is a pretty short explanation for an important subject. I think we should take something [like this](https://github.com/simolus3/drift/blob/6d31b37ae52ffa23f21e3e075b25ae7210790dc9/docs/docs/dart_api/writes.md?plain=1#L42-L56) as a better reference on how to introduce companions.

We could also mention that companions are not as relevant with the manager API as their structure is reflected by the callbacks for creating and updating rows. (similar to the annotation below)

docs/docs/dart_api/dataclass.md
```diff
@@ -1,23 +1,112 @@
...
+
+{{ load_snippet('generated-companion','lib/snippets/dart_api/dataclass.dart.excerpt.json') }}
+
+</div>
+1. `o()` is just a helper function that creates a `UsersCompanion`.
+
+### Value object
+When using the companion object to update a row, optional fields must be wrapped in a `Value` object. This is used by Drift to distinguish between `null` and not present values.
+
+{{ load_snippet('generated-value','lib/snippets/dart_api/dataclass.dart.excerpt.json') }}
```
simolus3: This example really isn't too helpful. Values were also used in the example above for updates, perhaps we could reference that here.

> This is used by Drift to distinguish between `null` and not present values.

A better explanation would be to mention what drift actually does with that: For updates, there's a distinction between setting the value of a column to `null` vs. not touching it at all. We can't represent without introducing some kind of optional structure, which is what `Value` is.

docs/docs/dart_api/expressions.md
```diff
@@ -76,6 +76,20 @@ bitwise operations:
...
+### BigInt
+
+You may want to cast an expression to a `BigInt` if:
+
+- The result of an arithmetic operation will be extremely large[^1].
+- You are compiling to JavaScript.
+
+[^1]: Like bigger than 4,503,599,627,370,496!
+
+Using `dartCast<BigInt>()` will ensure that the result is interpreted as a `BigInt` by drift.
```
simolus3, line 86: We should also note that this doesn't change the behavior in SQL at all.

docs/docs/dart_api/expressions.md
```diff
@@ -76,6 +76,20 @@ bitwise operations:
 
 {{ load_snippet('bitwise','lib/snippets/dart_api/expressions.dart.excerpt.json') }}
 
+### BigInt
+
+You may want to cast an expression to a `BigInt` if:
```
simolus3:
```suggestion
You may want to cast an expression to a `BigInt` if both of these apply to your app:
```

docs/docs/dart_api/expressions.md
```diff
@@ -76,6 +76,20 @@ bitwise operations:
...
+
+- The result of an arithmetic operation will be extremely large[^1].
```
simolus3: What's the point of being imprecise here? We can just write 
```suggestion
- The result of an arithmetic operation is not representable by a `double` (i.e., it is larger than `2^52`).
```

docs/docs/dart_api/manager.md
```diff
@@ -1,17 +1,11 @@
 ---
 
-title: Manager
+title: Queries
```
simolus3: IMO we should leave this as-is before the restructuring is more complete. At the moment we'd then effectively have multiple pages on queries which is more confusing than before.

We can name it "Queries with Manager" or something and also point out in the introduction/getting started guide that there are two ways to write these queries (like we already do on the new table page).

simolus3 (commented):
LGTM apart from the remaining comments on other pages. I can take care of them too.

docs/docs/_redirects
dickermoshe: Yes. Remove the leading url

docs/docs/dart_api/dataclass.md
dickermoshe: Why?

simolus3: Drift is a relational persistence library, and ideally shouldn't have to do anything to do with serialization. It's a historical artifact that I regret, but it's too late to fix this now.

Our serialization capabilities are incredibly restricted compared to packages that actually intend to do serialziation, which is why I recommend using custom row classes to combine drift with other libraries where necessary.

simolus3 (approved):
I've kept some of the existing terminology (like row classes) for consistency with the rest of the pages.
This is now in a state where I think the two pages are a very clear improvement on the current site - if you have more ideas on how these pages should be improved, we can obviously iterate on this more.

PR #3297, opened 2024-10-21
Add custom flutter run with headers to the web documentation

Unbreathable:
Hi,

I noticed that in the documentation for Drift it says that there is no way to add the custom CORS headers to the `flutter run` command. However, recently I stumbled upon a way to do it that I found in another package's documentation.

I propose to just put the custom `flutter run` command where the old notice was, but it could also go somewhere else of course.

I hope that this helps,
Unbreathable.


simolus3 (approved):
Thanks!

PR #3308, opened 2024-10-25
Fix bug where generated tests would not use the actual db class name but the hardcoded name "Database"

saibotma:


kuhnroyal:
Nice, noticed this as well.

simolus3:
Applied onto `development` in 84ce98ade6f995b4b560f2aa5c2bbec9c5d51d75, I've also released this in `drift_dev` version `2.21.2`. Thank you for the fix!

PR #3311, opened 2024-10-27
non-null-refs getter for non-null references

### LINKED ISSUE #3287: `withReferences()` should return non-nullable manager for required relation
_title says it all_ (at least for me)

dickermoshe:
Closes: https://github.com/simolus3/drift/issues/3287

until now the `withReferences` getter would always return a nullable manager even if a reference was required, this takes it into account

Tests weren't required because the drift test code itself now uses this without `?.` getters


drift/example/main.g.dart
```diff
@@ -939,7 +939,7 @@ final class $$TodoItemsTableReferences
       db.todoCategories.createAlias(
           $_aliasNameGenerator(db.todoItems.categoryId, db.todoCategories.id));
 
-  $$TodoCategoriesTableProcessedTableManager? get categoryId {
+  $$TodoCategoriesTableProcessedTableManager get categoryId {
     if ($_item.categoryId == null) return null;
```
simolus3: That `return null` doesn't look correct to me here, shouldn't we do a null assertion (`!`) instead? Or does this just need to be re-generated (the `drift_dev` changes seem to suggest that, and the `examples/app` code doesn't have this either).

simolus3 (commented):
Let's also add a changelog entry in `drift_dev` for this, apart from that this LGTM :+1:

dickermoshe:
Single test is failing, don't think it's my fault though

drift_dev/test/linter_test/pkg/custom_lint.log
```diff
@@ -0,0 +1,4597 @@
+The Dart VM service is listening on http://127.0.0.1:46360/h3mHpPGsdXw=/
```
simolus3: Probably due to other gitignores, but this file should be deleted.

dickermoshe:
@simolus3 Merge?

simolus3 (approved):
Yes. Let's do a squashed merge though, no need for the log to end up in the history :D

PR #3312, opened 2024-10-27
Fix `Referencing other tables` docs

AhmedLSayed9:


simolus3:
Hm that naming scheme makes sense to me, why close this? cc @dickermoshe

AhmedLSayed9:
> Hm that naming scheme makes sense to me, why close this? cc @dickermoshe

Ops. I thought I've re-opened another PR.
Re-opened now.

PR #3314, opened 2024-10-28
`managers` extends public classes

### LINKED ISSUE #3015: Database can no longer be mocked with new manager API

dickermoshe:
Instead of managers being fields on private classes:
```dart
class _Database extends GeneratedDatabase {
  Managers get managers => Managers(this);
}
Managers {
  // Private: Bad for Mockito
  final _Database db;
}

```
We now have them on extension on the public class:
```dart
class _Database extends GeneratedDatabase {
  ...
}
Managers {
  // Public
  final Database db;
}

extension DatabaseManagersExt on Database {
  Managers get managers => Managers(this);
}
```

This will allow mockito to do whatever it wants as Manager doesn't rely on a private class
closes https://github.com/simolus3/drift/issues/3015

### DIFF
--- a/docs/lib/snippets/drift_files/custom_queries.drift.dart
+++ b/docs/lib/snippets/drift_files/custom_queries.drift.dart
@@ -5,7 +5,6 @@ import 'package:drift_docs/snippets/_shared/todo_tables.drift.dart' as i1;
 
 abstract class $MyDatabase extends i0.GeneratedDatabase {
   $MyDatabase(i0.QueryExecutor e) : super(e);
-  $MyDatabaseManager get managers => $MyDatabaseManager(this);
   late final i1.$CategoriesTable categories = i1.$CategoriesTable(this);
   late final i1.$TodoItemsTable todoItems = i1.$TodoItemsTable(this);
   i0.Selectable<CategoriesWithCountResult> categoriesWithCount() {
--- a/docs/lib/snippets/modular/many_to_many/json.drift.dart
+++ b/docs/lib/snippets/modular/many_to_many/json.drift.dart
@@ -138,7 +138,6 @@ typedef $$ShoppingCartsTableProcessedTableManager = i0.ProcessedTableManager<
 
 abstract class $JsonBasedDatabase extends i0.GeneratedDatabase {
   $JsonBasedDatabase(i0.QueryExecutor e) : super(e);
-  $JsonBasedDatabaseManager get managers => $JsonBasedDatabaseManager(this);
   late final i1.$BuyableItemsTable buyableItems = i1.$BuyableItemsTable(this);
   late final i2.$ShoppingCartsTable shoppingCarts =
       i2.$ShoppingCartsTable(this);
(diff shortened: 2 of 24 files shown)


simolus3:
Why do they have to be extensions? So that mockito doesn't try to mock them?

dickermoshe:
What does the Managers class use?
```dart
Managers {
  // Private: Bad for Mockito
  final _Database db;
}
```
A private `_Database` class, so `db.managers` can't be mocked.

Hmm, lets make it work with the public one:
```dart
Managers {
  // Public!
  final Database db;
}
```
Ok, but wait:
```dart
class _Database extends GeneratedDatabase {
  Managers get managers => Managers(this); // BAD! `this` is not a `Database` its a `_Database `
}
```
So the only way `this` is a `Database` is if we use an `extension`.
```dart
extension DatabaseManagersExt on Database {
  Managers get managers => Managers(this);
}
```

What do you think?

dickermoshe:
There is a bug in mockito which is not allowing us to mock individual managers, I'm going to make a repro and open an issue on their repo, it's a mockito issue, not a drift one

simolus3:
Is there a way to test that we're not breaking mockito as part of the CI? We're already running that builder on `drift` so I think it should be possible? It would be good if we could detect regressions automatically once this is fixed.

dickermoshe:
@simolus3 Did you ever have this?
The analyzer returning the underlying type instead of the typedef itself?

https://github.com/dart-lang/mockito/issues/775#issuecomment-2442901089

This is blocking the mocking of actual managers.

dickermoshe:
I've added the database manager (db.managers) to the mockito so we know that now works, before it didn't.
However table managers still cant be mocked (db.managers.todos) due to a mockito bug.

This PR includes a mock that tests itself

https://github.com/simolus3/drift/pull/3314/files#diff-1cc77f2eccd6a270ee306cd04f5e9fd10d81e1431736ba0bbf8cfc7555a0d990R6-R9

dickermoshe:
Opened a [PR](https://github.com/dart-lang/mockito/pull/776) on mockito, if you know someone at google that can expedite this review I would appreciate it.


 But this PR is good to go. Let mockito figure themselves out.

dickermoshe:
My PR for mockito is complete. Hopefully they cherry-pick it into a release.

dickermoshe:
Argh!
You can't mock extensions!
https://github.com/dart-lang/build/issues/4768

dickermoshe:
Modular generation fixes this entire mess! Maybe we should add this to the docs. It makes a public base class

PR #3325, opened 2024-11-04
Use external package for snippets

simolus3:
Use the existing `code_snippets` package to highlight external snippets on the documentation site. This will require updates to fix the CSS to match the updated values extracted from the vs code theme.

In my mind, advantages of the `code_snippets` package are:

1. The ability to link to dartdoc / source code for referenced imports and methods.
2. More accurate syntax highlighting for drift files, which is not exactly SQL.
3. More accurate syntax highlighting for dart files by parsing the entire file using analyzer APIs instead of using the tree-sitting grammar which can't recognize everything correctly.
4. Less custom code in `docs/`.

At the moment, going back to `code_snippets` makes things worse due to the old css and outdated `code_snippets` builder. I'll try to fix those obviously, but I'm wondering if I'm overlooking additional things I should take into consideration.


dickermoshe:
image

dickermoshe:
some colors are off

PR #3347, opened 2024-11-21
[fix]fixed document's dead link

luxun1910:
the link of started guide was dead, so changed it to the correct one.


simolus3 (approved):
I've hopefully fixed the link in 2c29a3529c58256f2d17a06b561661b046b264e4, but it's always good to use the canonical URL.

PR #3363, opened 2024-11-30
fix: Add conditional export to support --wasm build

koji-1009:
Added `if (dart.library.html)` to `src/web/channel_legacy.dart` export.

With this change, `src/web/channel_legacy.dart` will only be referenced if `dart:html` is available, and builds with the `--wasm` option will succeed.

---

If we write the following migration code from IndexedDB to WasmDB, we need to do two imports, `drift/wasm.dart` and `drift/web.dart`.

```dart
import 'package:drift/wasm.dart';
import 'package:drift/web.dart';

const _databaseName = 'nice_db';

LazyDatabase getDatabase() => LazyDatabase(() async {
      final result = await WasmDatabase.open(
        databaseName: _databaseName,
        sqlite3Uri: Uri.parse('sqlite3.wasm'),
        driftWorkerUri: Uri.parse('drift_worker.dart.js'),
        initializeDatabase: () async {
          final storage = await DriftWebStorage.indexedDbIfSupported(
            _databaseName,
          );
          await storage.open();

          final blob = await storage.restore();
          await storage.close();

          return blob;
        },
      );

      return result.resolvedExecutor;
    });
```

In a build targeting the web, the current code work. (`flutter run -d chrome`)
However, in builds with the `--wasm` option, code using `dart:html` is not available. (`flutter run -d chrome --wasm`)

```
$ flutter run -d chrome --wasm
Launching lib/main.dart on Chrome in debug mode...
Target dart2wasm failed: ProcessException: Process exited abnormally with exit code 64:
../../../../.pub-cache/hosted/pub.dev/drift-2.22.0/lib/src/web/channel_legacy.dart:1:8: Error: Dart library 'dart:html' is not available on this platform.
import 'dart:html';
       ^
Context: The unavailable library 'dart:html' is imported through these packages:

    main.dart => package:example => package:encrypt_file_loader => package:drift => dart:html
```

### DIFF
--- a/drift/lib/src/web/channel_legacy_dummy.dart
+++ b/drift/lib/src/web/channel_legacy_dummy.dart
@@ -1,0 +1,27 @@
+import 'package:stream_channel/stream_channel.dart';
+
+/// Extension to transform a raw `MessagePort` from web workers into a Dart
+/// [StreamChannel].
+extension PortToChannel on dynamic {
+  /// Converts this port to a two-way communication channel, exposed as a
+  /// [StreamChannel].
+  ///
+  /// This can be used to implement a remote database connection over service
+  /// workers.
+  ///
+  /// The [explicitClose] parameter can be used to control whether a close
+  /// message should be sent through the channel when it is closed. This will
+  /// cause it to be closed on the other end as well. Note that this is not a
+  /// reliable way of determining channel closures though, as there is no event
+  /// for channels being closed due to a tab or worker being closed.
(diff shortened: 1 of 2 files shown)


simolus3 (approved):
Thanks! I think this is a good way to ease the migration, as we can continue to keep this setup until `dart:html` is removed for good.

koji-1009:
The code migration support via `extentsion on dynamic` looks great. I was not aware of that!
Thanks for the review!

PR #3371, opened 2024-12-11
Make the example app use the new migration tooling

dickermoshe:
also update everything to work with flutter 3.27 and update CI


dickermoshe:
we're running into dart version issues

examples/app/test/drift/default/migration_test.dart
```diff
@@ -0,0 +1,81 @@
...
+  // The following template shows how to write tests ensuring your migrations
+  // preserve existing data.
+  // Testing this can be useful for migrations that change existing columns
+  // (e.g. by alterating their type or constraints). Migrations that only add
+  // tables or columns typically don't need these advanced tests. For more
+  // information, see https://drift.simonbinder.eu/migrations/tests/#verifying-data-integrity
+  // TODO: This generated template shows how these tests could be written. Adopt
+  // it to your own needs when testing migrations with data integrity.
+  test("migration from v1 to v2 does not corrupt data", () async {
+    // Add data to insert into the old database, and the expected rows after the
```
simolus3: I think we should move the existing schema integrity test in the top-level `test/` directory of the app into this file, and then delete the top-level `migration_test.dart` file.

simolus3 (approved):

simolus3:
Could you update the commands mentioned in the `Migration` section under its readme as well? LGMT apart from that.

PR #3374, opened 2024-12-12
support for predefined path_provider types

### LINKED ISSUE #3372: Default Flutter Drift Database in My Documents on Windows
This is really bad. There should be an enum that defaults to the current folder for backwards compat.
The default should change for drift 3.0

dickermoshe:
Closes https://github.com/simolus3/drift/issues/3372

@simolus3 Does this seem ok to you. Do have another idea on how to resolve this?

### DIFF
--- a/drift_flutter/lib/src/connect.dart
+++ b/drift_flutter/lib/src/connect.dart
@@ -5,6 +5,7 @@ export 'unsupported.dart'
 export 'package:drift/src/web/wasm_setup/types.dart';
 // ignore: implementation_imports
 import 'package:drift/src/web/wasm_setup/types.dart';
+import 'package:path_provider/path_provider.dart';
 
 /// Web-specific options used to open drift databases.
 ///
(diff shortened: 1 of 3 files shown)


dickermoshe:
I've got no idea why CI is failing here.
We may have a flaky test here

drift_flutter/lib/src/connect.dart
```diff
@@ -67,10 +68,41 @@ final class DriftNativeOptions {
   /// a custom database path in another directory.
   final Future<String> Function()? databasePath;
 
+  /// The directory where the Drift database files are stored.
+  ///
+  /// If a custom [databasePath] is provided, this option must not be used.
+  ///
+  /// If no custom [databasePath] is provided and this option is not used, the
+  /// default directory is [DriftDatabaseDirectory.applicationDocumentsDirectory].
+  final DriftDatabaseDirectory? directory;
```
simolus3, line 77: Can we make this a `Future<Directory> Function()` and then let users use `directory: getApplicationSupportDirectory` instead?

I think this has similar ergonomics while also allowing different directory sources.

drift_flutter/lib/src/connect.dart
```diff
@@ -67,10 +68,41 @@ final class DriftNativeOptions {
...
+
   /// Create drift options effective when opening drift databases on native
   /// platforms.
   const DriftNativeOptions({
     this.shareAcrossIsolates = false,
     this.databasePath,
-  });
+    this.directory,
+  }) : assert(databasePath == null || directory == null,
```
simolus3, line 85: We should find a way to nudge users into a applying a reasonable default without breaking their apps silently (a compiler error is annoying, but at some point we want to change the default and, since not everyone is reading changelogs, working code shouldn't suddenly use a different directory).

There are different options that come to mind here, e.g.

1. Adding something like a fallback path that would be considered when looking for existing databases and defaults to `applicationDocumentsDirectory`. When we open the database and don't find it in the intended directory, we check if a file exists in the fallback directory and move it if necessary. That makes changing the default less breaking.
2. Making `directory` a required parameter and also requiring the native options on `driftDatabase`. So users have to specify something, and we recommend specifying `applicationSupportDirectory` _unless_ a published version of the app is already using something else.

I don't like option 1 too much because it makes things messy internally and is still breaking (users might now expect the path in other scenarios, e.g. when sharing the database with some native code also opening it).
For the required parameter, we'd probably have to remove the `databasePath` argument from the default constructor and add a named constructor (e.g. `withCustomPath`) that has that argument but no `directory` argument.
What are your thoughts on this?

simolus3 (commented):
I agree with the changes, but we should keep a path to eventually change the defaults in mind.

The test failure could be a flake, a bug in drift (that most likely has been there before), or also a regression in the Dart SDK. I'll keep an eye on it if it occurs more often, I think for now we can restart the CI.

drift_flutter/lib/src/connect.dart
dickermoshe: Smart 
Let's make it a `FutureOr<Directory> Function()` and we can just keep a single argument.

drift_flutter/lib/src/connect.dart
simolus3: Do you see any terrible downsides of the second approach @dickermoshe? Otherwise I'd start to prepare that in the next days so that we can do a release soon.

dickermoshe: i would love for there to be a sensible default, but i guess that ship has sailed. we can't change the default and expect people to notice. we can't require people to enter the native options, because they may be making a web only app

dickermoshe: seems like both approaches have issues

simolus3:
I've chosen a similar approach to yours without changing any defaults in the linked commit.

Hopefully, changing the suggestion on the "getting started" documentation page is enough.

Issue #3381, opened 2024-12-18
[drift_flutter] driftDatabase is broken in Flutter 3.27.0 on WASM Web


alestiago:
**Describe the bug**

Flutter 3.27.1 relies on [Dart version 3.6.0](https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=&ved=2ahUKEwiuqcLDtrGKAxWDRTABHZDAG9oQFnoECBUQAQ&url=https%3A%2F%2Fgithub.com%2Fdart-lang%2Fsdk%2Fblob%2Fmain%2FCHANGELOG.md&usg=AOvVaw2HHKDefECw3gD35BLlhGPg&opi=89978449) introduced a breaking change:

> The condition dart.library.js is now false on conditional imports in dart2wasm. Note that it was already a static error to import dart:js directly (see https://github.com/dart-lang/sdk/issues/55266).

This means that the conditional imports in `drift_flutter/lib/src/connect.dart` now resolve to `unssuported.dart` instead of `web.dart` when using `--wasm`:

https://github.com/simolus3/drift/blob/1146a69782282d87df1b73098392ddedf3283415/drift_flutter/lib/src/connect.dart#L1

Hence, **`drift_flutter` is broken for Web with Flutter 3.27.0 or above**.

**Reproductive steps**

1. Ensure you're using Flutter 3.27.1 or above:

```sh
fluter --version
```

2. Create a Flutter web application:

```sh
flutter create my_app --platform web
```

3. Add `drift_flutter` dependency (from my_app):

```sh
flutter pub add drift_flutter:0.2.2
```

4. Update the `main.dart` to:

```dart
import 'package:drift_flutter/drift_flutter.dart';
import 'package:flutter/material.dart';

void main() {
  runApp(const MyApp());
}

class MyApp extends StatelessWidget {
  const MyApp({super.key});

  // This widget is the root of your application.
  @override
  Widget build(BuildContext context) {
    const isRunningWithWasm = bool.fromEnvironment('dart.tool.dart2wasm');
    print('isRunningWithWasm: $isRunningWithWasm');

    const hasJsLibrary = bool.fromEnvironment('dart.library.js');
    print('hasJsLibrary: $hasJsLibrary');

    const hasFfiLibrary = bool.fromEnvironment('dart.library.ffi');
    print('hasFfiLibrary: $hasFfiLibrary');

    driftDatabase(name: 'Hello');

    return ColoredBox(color: Colors.red);
  }
[...]

PR #3382, opened 2024-12-18
fix: use dart.tool.dart2wasm in connect.dart conditional import

alestiago:
Potential fix for https://github.com/simolus3/drift/issues/3381.

Is still unclear to me from the research made what's the "migration" steps from the Dart team given the deprecation discussed in https://github.com/dart-lang/sdk/issues/55266 for these conditional imports.


simolus3:
Thanks for the report and PR! I think instead of adding another condition, it's easier to just match for `dart.library.js_interop` being available (we don't allow older SDKs without that library, so this also doesn't break existing users).

I've done that in 0f48c51a68f473c1c0e64ca9c0b8c0a9d3d8c7da, I'll publish the version as soon as the CI has passed.

PR #3383, opened 2024-12-18
Update dependencies to support latest `dart_style` version

Carapacik:


simolus3:
FYI @dickermoshe: It appears to be impossible to support both Flutter 3.24 and the latest `analyzer` / `dart_style` packages in `drift_dev`. So I had to (again, ...) disable the tests with the older version. I still think we should generally support two stable versions, but drift also shouldn't prevent anyone from using the latest analyzer versions.

dickermoshe:
Flutter 3.21, 3.24, 3.27 have each been released with either major bugs or
massively breaking changes which the flutter team seems to pretend aren’t
that major.

Because of this many people don’t migrate to the latest versions of flutter
when they become available. Most of drift users are flutter developers.
Many of the bugs that we find in drift are mission critical. If we have to
ignore depreciations in order to land these fixes in people‘s codebases I
think it’s worth it.

We should at least be supporting two versions of flutter. Or at least wait
until a couple of hot fixes come in.

I consider 3.27 to basically be a beta release that’s how unstable things
have gotten

dickermoshe:
Also, the latest version of analyzer is many many major versions ahead of
where flutter is at.

dickermoshe:
@simolus3 There will be instances where even the latest versions of flutter won't be supported, so users will have to wait 6 months for fixes to critical bugs?!
We have to find a happy medium

dickermoshe:
@Carapacik What is the motivation for this? I'm trying to understand the value this would bring you.
(btw, sick work on swagger_parser)

Carapacik:
I also think it's worth waiting for the next version of Flutter before merging this PR.
Take a look at [this](https://github.com/dart-lang/dart_style/issues/1253) and [this](https://pub.dev/packages/dart_style/changelog#300).

Carapacik:
@dickermoshe We are actively working on a complete rewrite of the parser in swagger_parser.

Carapacik:
@dickermoshe  @simolus3  If not for this, then everything would work on flutter 3.24.
image

Carapacik:
We can do this in drift_dev:
```yaml
dependency_overrides:
  collection: ^1.18.0
```

Carapacik:
Alternatively, I can turn it into a **draft** and update it with the release of the next stable version of Flutter.
But it works on the current stable version (3.27.0).

simolus3:
Let's see if we can find a way to `// ignore: deprecation` a way out of this.

In my mind, it's important that we generally _support_ the latest versions of all dependencies so that we don't hold everyone back. I underestimated the importance of supporting older Flutter versions though, I agree that we have to find a way to do both.

simolus3:
Looks like the scope and `enclosingElement` things really make this impossible. I'll apply the `lints` updates to `develop` right away, but then let's mark this as a draft until the next Flutter version.

simolus3:
The good news is that we can support the latest `dart_style` without the latest `analyzer`: 79e696719aa5d44b5edd30eb886e1fe5443a8b8f.

Carapacik:
@simolus3 I think this PR will wait for the release of the next version of Dart.

Carapacik:
I think it can be merged and squashed if you want.
Also[ new version ](https://github.com/Carapacik/drift/tree/align-flutter-3.27)with new analyzer and Flutter >=3.27

Carapacik:
@simolus3 All tests passing(except one)

drift/pubspec.yaml
```diff
@@ -16,33 +16,33 @@ environment:
...
 
 dependencies:
-  async: ^2.5.0
-  convert: ^3.0.0
-  collection: ^1.15.0
-  meta: ^1.3.0
+  async: ^2.11.0
+  convert: ^3.1.0
+  collection: ^1.18.0
+  meta: ^1.12.0
```
simolus3: I don't really see the point of these updates? Are we inadvertently using features added in these later versions? This just restricts dependencies for no reason (or at least no reason apparent to me)?

I suppose technically we're incompatible with the oldest version declared here due to other dependencies which might require newer versions, but that doesn't mean that we'd have to update them.

simolus3 (commented):
I've fixed the failing test on `develop`, forgot to re-run build-runner after updating dependencies.

drift/pubspec.yaml
Carapacik: They all correspond to pinned dependencies in Flutter version 3.22(Dart 3.4)

Carapacik: If it is critical, I will lower these versions back.

dickermoshe:
Exactly,  `dart pub downgrade` is a really awesome tool for seeing how low you can go.

simolus3 (approved):
Pubspec and analyzer changes look good to me. I've had to downgrade `mockito` on `develop` because the latest version does not configure the formatter correctly, breaking builds. Rebasing should fix this branch.

Carapacik:
@simolus3  Please squash commits

PR #3387, opened 2024-12-20
Add null ordering to manager API

simolus3:
Requested in https://github.com/simolus3/drift/discussions/3386.


dickermoshe (approved):
Can't believe I missed this.

PR #3399, opened 2025-01-04
Modular manager references

### LINKED ISSUE #3398: cannot prefetch references if we use modules
**Cannot prefetch references if we use modules** 

Hi, I cannot access to prefetched references if I use drift_dev:modular manually.
If you build  the following code, it works fine without build.yaml and without drift  modular : 

```
part 'db.g.dart';

class Tsid {
  final BigInt value;

  const Tsid(this.value);
}

class TsidTypeConverter extends TypeConverter<Tsid, BigInt> {
  const TsidTypeConverter();

  @override
  Tsid fromSql(BigInt fromDb) {
    return Tsid(fromDb);
  }

  @override
  BigInt toSql(Tsid value) {
    return value.value;
  }
}

@DriftAccessor(tables: [
  DynamicRelations,
  DynamicTableProperties,
])
class DynamicDatabaseDao extends DatabaseAccessor<Database>
    with $DynamicDatabaseDaoMixin {
  DynamicDatabaseDao(super.attachedDatabase);

  $$DynamicRelationsTableTableManager get relationsManager =>
      attachedDatabase.managers.dynamicRelations;

  Future<DynamicTableProperty?> getRelation(Tsid relationId) => relationsManager
      .withReferences()
      .filter((f) => f.id.equals(relationId))
      .getSingle()
      .then((value) => value.$2.sourcePropertyId.getSingle());
}

class DynamicRelations extends Table {
  Int64Column get id => int64().map(const TsidTypeConverter())();

  Int64Column get sourcePropertyId => int64()
      .map(const TsidTypeConverter())
      .references(DynamicTableProperties, #id)();

  Int64Column get targetPropertyId => int64()
      .map(const TsidTypeConverter())
      .references(DynamicTableProperties, #id)();

  DateTimeColumn get creationDate =>
      dateTime().withDefault(currentDateAndTime)();

  @override
  String get tableName => 'dynamic_database__dynamic_relations';

  @override
  Set<Column<Object>>? get primaryKey => {id};
}

class DynamicTableProperties extends Table {
  Int64Column get id => int64().map(const TsidTypeConverter())();

  Int64Column get tableId => int64().map(const TsidTypeConverter())();

  @override
  Set<Column<Object>>? get primaryKey => {id};

  @override
  List<Set<Column<Object>>>? get uniqueKeys => [
        {name, tableId}
      ];

  @override
  String get tableName => 'dynamic_database__properties_names';
}

@DriftDatabase(tables: [
  DynamicRelations,
  DynamicTableProperties,
], daos: [
  DynamicDatabaseDao
], views: [])
abstract class Database extends $Database {
  final String databaseName;

  Database({
    required this.databaseName,
    required QueryExecutor executor,
  }) : super(executor);

  @override
  int get schemaVersion => 1;
}
```
 pubspec.yaml is:
```
dependencies:
  flutter:
    sdk: flutter

  drift: ^2.23.0
  drift_flutter: ^0.2.4

dev_dependencies:
  flutter_test:
    sdk: flutter


  flutter_lints: ^5.0.0
  drift_dev: ^2.23.0
  build_runner: ^2.4.14
```

Now, if in the project we want to use drift dev modular configuration, like this: 

```
targets:
  $default:
    builders:
      drift_dev:
        # disable drift's default builder, we're using the modular setup
        # instead.
        enabled: false

      # Instead, enable drift_dev:analyzer and drift_dev:modular manually:
      drift_dev:analyzer:
        enabled: true
        options: &options
          # Drift build options, as per https://drift.simonbinder.eu/docs/advanced-features/builder_options/
          store_date_time_values_as_text: true
          named_parameters: true
          sql:
            dialect: sqlite
            options:
              version: "3.39"
              modules: [fts5]
      drift_dev:modular:
        enabled: true
        # We use yaml anchors to give the two builders the same options
        options: *options

```

And in the db.dart file we will change the
`part 'db.g.dart'` 

with : 

`import db.drift.dart
`

We will have the following error code : 

**The getter 'sourcePropertyId' isn't defined for the type 'BaseReferences<GeneratedDatabase, $DynamicRelationsTable, DynamicRelation>'.
Try importing the library that defines 'sourcePropertyId', correcting the name to the name of an existing getter, or defining a getter or field named 'sourcePropertyId'.**

Errore detected in the following line (sourcePropertyId is the problem): 
```
  Future<DynamicTableProperty?> getRelation(Tsid relationId) => relationsManager
      .withReferences()
      .filter((f) => f.id.equals(relationId))
      .getSingle()
      .then((value) => value.$2.sourcePropertyId.getSingle());

```

Any idea/suggestion?
Thank you

simolus3:
The manager templates used to disable generating a typed reference class when modular code-generation was enabled, instead referring to the general `BaseReferences` class in the runtime package.

This adds support for that feature by resolving the modular-specific issues (like referring to typedefs and classes by their imported name if needed).

This is tested by the modular package which can now use that feature. But @dickermoshe it would be helpful if you could take a quick look to make sure I didn't miss anything.

Closes https://github.com/simolus3/drift/issues/3398.

### DIFF
--- a/drift_dev/lib/src/writer/manager/table_manager_writer.dart
+++ b/drift_dev/lib/src/writer/manager/table_manager_writer.dart
@@ -87,12 +87,15 @@ class _TableManagerWriter {
       final fieldNameCount = allFieldNames
           .where((fieldName) => fieldName == relation.fieldName)
           .length;
-      if (fieldNameCount != 1 && table.id.isDefinedInDart) {
-        print(
-            "Duplicate orderings/filters detected for field \"${relation.fieldName}\" on table \"${table.entityInfoName}\"."
-            " Filter and orderings for this field wont be generated."
-            " Use the @ReferenceName() annotation to resolve this issue."
-            " See https://drift.simonbinder.eu/docs/manager/#name-clashes for more information");
+      if (fieldNameCount != 1) {
+        if (table.id.isDefinedInDart) {
+          print(
+              "Duplicate orderings/filters detected for field \"${relation.fieldName}\" on table \"${table.entityInfoName}\"."
+              " Filter and orderings for this field wont be generated."
+              " Use the @ReferenceName() annotation to resolve this issue."
+              " See https://drift.simonbinder.eu/docs/manager/#name-clashes for more information");
+        }
+
         return false;
       }
       return true;
--- a/drift_dev/lib/src/writer/manager/table_manager_writer.dart
+++ b/drift_dev/lib/src/writer/manager/table_manager_writer.dart
@@ -87,7 +87,7 @@ class _TableManagerWriter {
       final fieldNameCount = allFieldNames
           .where((fieldName) => fieldName == relation.fieldName)
           .length;
-      if (fieldNameCount != 1) {
+      if (fieldNameCount != 1 && table.id.isDefinedInDart) {
         print(
             "Duplicate orderings/filters detected for field \"${relation.fieldName}\" on table \"${table.entityInfoName}\"."
             " Filter and orderings for this field wont be generated."
@@ -170,7 +170,7 @@ class _TableManagerWriter {
           .add(_templates.relatedAnnotations(leaf: leaf, relation: relation));
     }
 
-    if (!scope.generationOptions.isModular && relations.isNotEmpty) {
+    if (relations.isNotEmpty) {
       leaf.write(_templates.rowReferencesClass(
           table: table,
           relations: relations,
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -273,6 +273,7 @@ jobs:
           dart pub upgrade
           dart run build_runner build --delete-conflicting-outputs
           dart run bin/example.dart
+          dart test
       - name: Integration test for migrations example
         working-directory: examples/migrations_example
         run: |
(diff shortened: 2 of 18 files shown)


drift_dev/lib/src/writer/manager/table_manager_writer.dart
```diff
@@ -88,11 +88,14 @@ class _TableManagerWriter {
           .where((fieldName) => fieldName == relation.fieldName)
           .length;
       if (fieldNameCount != 1) {
-        print(
-            "Duplicate orderings/filters detected for field \"${relation.fieldName}\" on table \"${table.entityInfoName}\"."
-            " Filter and orderings for this field wont be generated."
-            " Use the @ReferenceName() annotation to resolve this issue."
-            " See https://drift.simonbinder.eu/docs/manager/#name-clashes for more information");
+        if (table.id.isDefinedInDart) {
```
simolus3, line 91: I've disabled the warnings for tables defined in `.drift` files since there currently is no way to fix it.

dickermoshe: That makes sense

dickermoshe (changes requested):
Looks great.

I think we should add tests so that all manager tests are also ran on modular code.

btw, when using the manager it's really helpful to do modular generation because otherwise the generic is private (_$AppDatabase) so I'm pumped about this

simolus3:
> I think we should add tests so that all manager tests are also ran on modular code.

This is tricky to pull off without copy-pasting all the tests and keeping them in sync :thinking: After all, we'd have to run both the default and the modular builder on the same database definitions, but they would generate separate and independent manager classes.

As a first step, I've added a simple test to the `modular` package.

dickermoshe (approved):
If you feel the testing is sufficient.
btw, with the status on this PR https://github.com/simolus3/drift/pull/3281

PR #3404, opened 2025-01-11
Fine-grained query intercepting

shroff:
I want to be able to intercept queries executed only in certain transaction contexts, and I couldn't find another/easier way.

I thought of filing an issue for this, but it just seemed easier to do.

### DIFF
--- a/drift/lib/src/runtime/api/connection_user.dart
+++ b/drift/lib/src/runtime/api/connection_user.dart
@@ -471,7 +471,7 @@ abstract class DatabaseConnectionUser {
   /// See also:
   ///  - the docs on [transactions](https://drift.simonbinder.eu/docs/transactions/)
   Future<T> transaction<T>(Future<T> Function() action,
-      {bool requireNew = false}) async {
+      {bool requireNew = false, QueryInterceptor? interceptor}) async {
     final resolved = resolvedEngine;
 
     // Are we about to start a nested transaction?
@@ -489,6 +489,9 @@ abstract class DatabaseConnectionUser {
     }
 
     return await resolved.doWhenOpened((executor) {
+      if (interceptor != null) {
+        executor = executor.interceptWith(interceptor);
+      }
       final transactionExecutor = executor.beginTransaction();
       final transaction = Transaction(this, transactionExecutor);
 
--- a/drift/lib/src/runtime/api/connection_user.dart
+++ b/drift/lib/src/runtime/api/connection_user.dart
@@ -468,16 +468,10 @@ abstract class DatabaseConnectionUser {
   ///   transaction can continue to run if it catched the exception thrown by
   ///   the inner transaction when it aborted.
   ///
-  /// The optional [interceptor] can be used to apply a [QueryInterceptor] on
-  /// the opened transaction. This is useful to modify the behavior of the
-  /// executor for this specific transaction only. Note that lifecycle methods
-  /// will be called starting with [QueryInterceptor.beginTransaction] for the
-  /// transaction that is about to begin.
-  ///
   /// See also:
   ///  - the docs on [transactions](https://drift.simonbinder.eu/docs/transactions/)
   Future<T> transaction<T>(Future<T> Function() action,
-      {bool requireNew = false, QueryInterceptor? interceptor}) async {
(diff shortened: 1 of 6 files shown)


drift/lib/src/runtime/api/connection_user.dart
```diff
@@ -489,6 +489,9 @@ abstract class DatabaseConnectionUser {
     }
 
     return await resolved.doWhenOpened((executor) {
+      if (interceptor != null) {
+        executor = executor.interceptWith(interceptor);
```
simolus3: I think we should only intercept the `transactionExecutor` here (so we need another extension on `TransactionExecutor` to apply interceptors that returns a `TransactionExecutor` too).

The parameter suggests that the interceptor will only be applied to the transaction, so that way feels more consistent.

drift/lib/src/runtime/api/connection_user.dart
```diff
@@ -471,7 +471,7 @@ abstract class DatabaseConnectionUser {
   /// See also:
   ///  - the docs on [transactions](https://drift.simonbinder.eu/docs/transactions/)
   Future<T> transaction<T>(Future<T> Function() action,
-      {bool requireNew = false}) async {
+      {bool requireNew = false, QueryInterceptor? interceptor}) async {
```
simolus3: We should also mention this in the documentation comment, e.g. something like `The optional [interceptor] can be used to apply a [QueryInterceptor] on the opened transaction. This is useful to modify the behavior of the executor for a specific transaction only.`

simolus3 (commented):
Thanks for the contribution! The approach looks ok as an easy way to only apply the interceptor in some scopes.

We'll also need an test for this (see `interceptor_test.dart` as an example), something that applies an interceptor for a transaction and asserts that it was applied for the methods within the transaction block.

drift/lib/src/runtime/api/connection_user.dart
shroff: I thought about only intercepting the `transactionExecutor`, but that felt incomplete since it wouldn't invoke the `beginTransaction` callback, which is where the lifecycle of the transaction actually begins. This way the `begin/commit/rollbackTransaction` methods would be matched up, instead of leaving them unpaired (only the top-level one for nested).

In my case, I do need to initialize some things when the transaction begins, so that would require extra bookkeeping in the `runXxx` methods.

If you still think you would prefer to only intercept the `transactionExecutor`, then I would be happy to respect that choice and do it that way. Let me know.

simolus3:
How about an approach that allows applying a `QueryInterceptor` on some arbitrary scope? E.g., what if we had an API like this:

```dart
await database.withInterceptor(interceptor: yourInterceptor, () async {
  // In this block, the interceptor is active
  await database.transaction(() async {
    // intercepted beginTransaction(), and so on...
  });
});
```

This has the advantage that it doesn't just apply to transactions, but may also be used for batches or any other drift APIs. Whenever some specific operation should run with an interceptor, you'd just use that method. It also gives you control over whether `beginTransaction` should be intercepted (if you call `transaction` inside `withInterceptor`, yes; if you call `withInterceptor` inside `transaction`, no).

I think it can be implemented by adding this to `DatabaseConnectionUser`:

```dart
Future<T> withQueryInterceptor<T>(
  Future<T> Function() action, {
  required QueryInterceptor interceptor,
}) async {
  return await resolvedEngine.doWhenOpened((executor) {
    final inner = _ExclusiveExecutor(this,
        executor: executor.interceptWith(interceptor));
    return _runConnectionZoned(inner, action);
  });
}
```

(Think because I didn't test this at all).

shroff:
That's a great idea! And the API is much cleaner and doesn't complicate existing functionality.

In terms of naming, `withInterceptor` or `withQueryInterceptor` might lead to confusion with `QueryExecutor.interceptWith`. How about `.intercept` (like `.transaction`, or `.exclusively`), or `.runWithInterceptor`?

simolus3:
`runWithInterceptor` sounds good 👍

simolus3 (approved):
Thanks for the contribution, this looks good to me :+1: I've added a brief mention of this feature in the documentation.

PR #3420, opened 2025-01-19
Feature: Add dispose method to DatabaseTracker for resource management

### LINKED ISSUE #3419: Unclosed SQLite Tracker Database in DatabaseTracker
[https://github.com/simolus3/drift/blob/develop/drift/lib/src/sqlite3/database_tracker.dart](https://github.com/simolus3/drift/blob/develop/drift/lib/src/sqlite3/database_tracker.dart)

I noticed that the DatabaseTracker class currently lacks a dispose() method to explicitly close the _db resource. This could potentially lead to resource leaks, especially in long-running applications or during Flutter hot restarts.

Would it be okay if I worked on adding a `dispose()` method to ensure proper cleanup of the in-memory SQLite database and any related resources?

suojae:
### Summary
This pull request introduces a `dispose()` method to the `DatabaseTracker` class to address potential resource leaks. The dispose method ensures proper cleanup of in-memory SQLite databases and tracked resources.

### Changes
- Added a `dispose()` method in `DatabaseTracker` to explicitly release tracked database connections.
- Introduced an `_isDisposed` flag to prevent method calls after disposal, ensuring safe resource management.
- Updated the `closeExisting()` method to work in conjunction with `dispose()` for complete cleanup.
- Added new tests in `database_tracker_test.dart` to validate:
  - Successful cleanup of database resources using `dispose()`.
  - Throwing of `StateError` when methods are called after disposal.
  - Safe handling of multiple calls to `dispose()`.

### Motivation
The lack of a `dispose()` method in `DatabaseTracker` can lead to unclosed SQLite connections, especially in long-running applications or during hot restarts in Flutter. This implementation ensures proper resource management and avoids potential deadlocks or memory leaks.

### Related Issue
- Fixes #3419

### DIFF
--- a/drift/CHANGELOG.md
+++ b/drift/CHANGELOG.md
@@ -7,6 +7,7 @@
   JSON.
 - Add `runWithInterceptor` method to databases to only apply interceptors in
   a restricted block.
+- Add `dispose()` method to `DatabaseTracker` to explicitly release in-memory SQLite resources.
 
 ## 2.23.1
 
--- a/drift/lib/src/sqlite3/database_tracker.dart
+++ b/drift/lib/src/sqlite3/database_tracker.dart
@@ -24,6 +24,9 @@ final DatabaseTracker tracker = DatabaseTracker();
 class DatabaseTracker {
   final Database _db;
 
+  /// Whether this [DatabaseTracker] has been disposed.
+  bool _isDisposed = false;
+
   /// Creates a new tracker with necessary tables.
   DatabaseTracker()
       : _db = sqlite3.open(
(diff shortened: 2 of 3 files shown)


simolus3 (commented):
Thanks!

One thing that worries me is that it's not impossible to open regular `NativeDatabase`s after closing the tracker.
Could we make a getter for `_isDisposed` public? Then `lib/native.dart` can be adopted to only mark the database as opened or closed when the tracker is active. That would allow using native databases without tracking them when the tracker is closed.

suojae:
Thank you for the feedback. I’ve added a public getter isDisposed to allow external code to check the tracker’s disposal state. This will help ensure that lib/native.dart can properly handle database operations only when the tracker is active.

Please let me know if there are any additional adjustments you'd like!

PR #3422, opened 2025-01-19
Fix manager references around custom row classes

### LINKED ISSUE #3335: A table having foreign key reference which is not reflected in a custom data class throws an error
As per [docs](https://drift.simonbinder.eu/custom_row_classes/?h=custom#using-custom-classes): A table can have additional columns not reflected in a custom data class. Drift will simply not load those columns when mapping a row.

Despite that, the following code can't run:
```dart
class AnotherTable extends Table {
}

@UseRowClass(Item)
class Items extends Table {
  IntColumn get anotherTable => integer().references(AnotherTable, #id)();
}

class Item {
  Item({required this.id});

  final int id;
}
```
It throws the following error: `Error: The getter 'anotherTable' isn't defined for the class 'Item'.`

simolus3:
In the manager classes responsible for resolving references, our approach for generating subsequent queries was to:

1. Map the row for the original query into the row class for the source table.
2. Extract the foreign key by fetching the field from the row class.
3. Using that to build the subsequent query.

Especially around custom row classes, step 2 can be broken because these classes are not guaranteed to contain fields for all columns. Luckily, we have access to the raw result set and can use that to resolve a specific column regardless of whether it is exposed through a field.
@dickermoshe, let me know if I missed something in the changed templates. But this also seems to be covered quite well by existing tests. So I'm confident this didn't break anything.

Closes https://github.com/simolus3/drift/issues/3335

### DIFF
--- a/docs/lib/snippets/_shared/todo_tables.drift.dart
+++ b/docs/lib/snippets/_shared/todo_tables.drift.dart
@@ -38,13 +38,14 @@ final class $$TodoItemsTableReferences extends i0
                   .id));
 
   i1.$$CategoriesTableProcessedTableManager? get category {
-    if ($_item.category == null) return null;
+    final $_column = $_itemColumn<int>('category');
+    if ($_column == null) return null;
     final manager = i1
         .$$CategoriesTableTableManager(
             $_db,
             i3.ReadDatabaseContainer($_db)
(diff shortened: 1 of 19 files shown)


dickermoshe (approved):
Very smart!

PR #3423, opened 2025-01-19
Address CI issues in DatabaseTracker

### LINKED ISSUE #3419: Unclosed SQLite Tracker Database in DatabaseTracker

suojae:
This PR addresses the CI issues reported in the [previous PR](https://github.com/simolus3/drift/pull/3420).
The CI failures appear to be environment-specific. 

#### Related Issue
- Fixes #3419

### DIFF
--- a/drift/lib/src/sqlite3/database_tracker.dart
+++ b/drift/lib/src/sqlite3/database_tracker.dart
@@ -24,6 +24,12 @@ final DatabaseTracker tracker = DatabaseTracker();
 class DatabaseTracker {
   final Database _db;
 
+  /// Whether this [DatabaseTracker] has been disposed.
+  bool _isDisposed = false;
+
+  /// Public getter to check if this tracker has been disposed.
+  bool get isDisposed => _isDisposed;
+
   /// Creates a new tracker with necessary tables.
   DatabaseTracker()
       : _db = sqlite3.open(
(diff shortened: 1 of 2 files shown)


simolus3:
Looks like you also need to run `dart format`.

suojae:
Thanks! I’ve addressed the formatting issue as suggested. Please let me know if there’s anything else I can help with or any additional changes you'd like me to make. ☺️

simolus3:
Thanks, and sorry for being slow to respond. We still need to investigate the crash in the test (see the failing github action). It looks like we might close the database twice somehow, or fail to unregister a native callback. That shouldn't be possible (and might be a bug in `package:sqlite3`). But it's weird that these changes surface that issue.

suojae:
Thanks! I’ll take a look at sqlite3 and see if I can spot anything related to the crash. I'll let you know if I find anything useful :)

PR #3427, opened 2025-01-20
FTS5 highlight/snippet nullable

FaFre:
In case the indexed value is null (e.g. multiple columns in the FTS5 table) highlight and snippet also return null.


simolus3:
Thanks! It looks like the tests need to be updated as well. Also, we currently assume all fts5 columns are non-nullable (`nullable: true` should be set when creating table columns in `_Fts5Module.parseTable` but isn't). So we should probably fix that too.

I'm thinking about a way to make this non-breaking... Perhaps we should only consider the columns to be nullable when the user has opted into a certain sqlite versions in the build options. But I can take care of that part.

PR #3468, opened 2025-02-11
Fix column type in `migrations` docs

AhmedLSayed9:


docs/docs/Migrations/api.md
```diff
@@ -61,7 +61,7 @@ For instance, if we wanted to cast a column before copying it, we could use:
 
 ```dart
 columnTransformer: {
-  todos.category: todos.category.cast<int>(),
+  todos.category: todos.category.cast<String>(),
```
simolus3: This snippet doesn't appear to be related to the other one?

docs/docs/Migrations/api.md
```diff
@@ -101,7 +101,7 @@ After re-running your build and incrementing the schema version, you can write a
 
 The important part here is the `columnTransformer` - a map from columns to expressions that will
 be used to copy the old data. The values in that map refer to the old table, so we can use
-`todos.category.cast<int>()` to copy old rows and transform their `category`.
+`todos.category.cast<String>()` to copy old rows and transform their `category`.
```
simolus3: The patch snippet above uses the wrong types (thanks for the fix!), but isn't a cast to `int` correct in this case? The text mentions that we've switching from `text()` to `integer()`, so the correct migration involves a `cast<int>()` or am I overlooking something?

simolus3 (commented):
Thanks for fixing the patch snippet for the table. The other changes aren't clear to me though, could you take another look (or let me know if I'm wrong)? Thanks!

docs/docs/Migrations/api.md
AhmedLSayed9: It's not directly related but I think it's mentioned as an intro and matched with the sub-section below it.

docs/docs/Migrations/api.md
AhmedLSayed9: My bad. I got confused by the previous line `The values in that map refer to the old table` I thought it was referring to the value at the generic type. Oh well.

That said, using the wrong type doesn’t trigger any errors at compile time or runtime, correct?

simolus3: A cast can't trigger errors at compile time. Using the wrong type would be okay for most tables, but would throw a runtime error in `STRICT` tables.

simolus3 (approved):
Thanks!

docs/docs/Migrations/api.md
AhmedLSayed9: What If `category` column was non-nullable text() and we want to change it to nullable text(), do we still need to write:
```
return MigrationStrategy(
  onUpgrade: (m, old, to) async {
    if (old <= yourOldVersion) {
      await m.alterTable(
        TableMigration(todos, columnTransformer: {
          todos.category: todos.category.cast<String>(),
        }),
      );
    }
  },
);
```
or just increase `schemaVersion`, or none?

simolus3: > What If `category` column was non-nullable text() and we want to change it to nullable text()

You need to increase your `schemaVersion` and write an empty `TableMigration`:

```dart
await m.alterTable(TableMigration(todos));
```

If you're unsure about which migrations to write, [guided migrations](https://drift.simonbinder.eu/Migrations/#guided-migrations) and their associated tests may be helpful.

PR #3469, opened 2025-02-12
[Web] Fix Shared Worker Probe

brian-superlist:
The sharedWorker probe fails due to a incorrect type castings. You cannot cast a JSBoolean to a `bool` value directly, you must go through the `toDart` conversion.

This failure leads drift to exclude the shared IndexedDb implementation from being used, falling back to the unsafe implementation.

This PR fixes the casts, allowing the probe to properly read the data and include `sharedIndexedDb` as the correct chosen implementation.

### DIFF
--- a/drift/lib/src/web/channel_legacy.dart
+++ b/drift/lib/src/web/channel_legacy.dart
@@ -1,4 +1,4 @@
-import 'dart:html';
+import 'dart:html'; // ignore: deprecated_member_use
 
 import 'package:stream_channel/stream_channel.dart';
 
--- a/drift/lib/src/web/wasm_setup/protocol.dart
+++ b/drift/lib/src/web/wasm_setup/protocol.dart
@@ -3,8 +3,8 @@
 import 'dart:js_interop';
 import 'dart:js_interop_unsafe';
 
-import 'package:web/web.dart' hide WorkerOptions;
 import 'package:sqlite3/wasm.dart';
+import 'package:web/web.dart' hide WorkerOptions;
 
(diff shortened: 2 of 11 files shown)


drift/lib/src/web/wasm_setup/protocol.dart
```diff
@@ -3,8 +3,8 @@
 import 'dart:js_interop';
 import 'dart:js_interop_unsafe';
 
-import 'package:web/web.dart' hide WorkerOptions;
 import 'package:sqlite3/wasm.dart';
+import 'package:web/web.dart' hide WorkerOptions;
```
brian-superlist, line 7: "Organize imports" noise, I can remove it if it's annoying.

drift/lib/src/web/channel_legacy.dart
```diff
@@ -1,4 +1,4 @@
-import 'dart:html';
+import 'dart:html'; // ignore: deprecated_member_use
```
brian-superlist, line 1: My first set of CI Tests failed because the dart analyzer complained about this import. I've specifically ignored it now.

simolus3: Ignoring is actually the best thing we can do here. This library is guarded behind conditional exports and only kept here for backwards compatibility.

simolus3 (approved):
The changes look good to me, thanks! I've also ported our web integration tests to `dart2wasm` on this branch (which catches the issue, we check that we can use the expected implementations for each browser).

PR #3470, opened 2025-02-12
Update the generated TODO to follow the Flutter style

AhmedLSayed9:
This fixes the lint complaint about [flutter_style_todos](https://dart.dev/tools/linter-rules/flutter_style_todos)

### DIFF
--- a/drift_dev/lib/src/cli/commands/make_migrations.dart
+++ b/drift_dev/lib/src/cli/commands/make_migrations.dart
@@ -424,7 +424,7 @@ void main() {
   // (e.g. by alterating their type or constraints). Migrations that only add
   // tables or columns typically don't need these advanced tests. For more
   // information, see https://drift.simonbinder.eu/migrations/tests/#verifying-data-integrity
-  // TODO: This generated template shows how these tests could be written. Adopt
+  // TODO(User): This generated template shows how these tests could be written. Adopt
   // it to your own needs when testing migrations with data integrity.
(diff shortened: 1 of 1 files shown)


simolus3 (commented):
This one I'm not so sure about. It's a lint, but you're also supposed to fix the todos...

It's also not a lint enabled by `flutter_lints`, did you enable that manually or where is that coming from?

AhmedLSayed9:
> This one I'm not so sure about. It's a lint, but you're also supposed to fix the todos...
> 
> It's also not a lint enabled by `flutter_lints`, did you enable that manually or where is that coming from?

It's enabled by default for users using [very_good_analysis](https://pub.dev/packages/very_good_analysis).

Using `// TODO(username): message.` will still ask the user to fix the todo but it won't make the analyzer fail (if `flutter_style_todos` is enabled).

simolus3 (changes requested):
I'm still not convinced about this.

First, adding a dummy name goes against the intention of the lint:

> TODOs should include the string TODO in all caps, followed by the GitHub username of the person with the best context about the problem referenced by the TODO in parenthesis.

So `// TODO: This gen...` and `// TODO(User): This gen...` both violate the rule, it's just that the linter can't understand the second violation.

And then, if `TODO`s without names are a lint, isn't that just a good sign for the user running `make-migrations` to either resolve them right away or to add their name?

Since we don't know the user running the command, I think the current template is appropriate (but feel free to respond or open another PR if there's anything I'm obviously overlooking with my reasoning here).

PR #3474, opened 2025-02-13
SQLite Window Function Expression

kedarkarkee:
The PR adds support for Window Function Expression as per specification on [SQLite Window Functions](https://www.sqlite.org/windowfunctions.html)

Currently using it in personal project. Thought it would be a good addition to the plugin.

# Basic Usage
 ```dart
 // Simple running total example
 final runningTotal = WindowFunctionExpression<double>(
   items.price.sum(), // aggregate function
   orderBy: [OrderingTerm.asc(items.date)],
 );

 // Use in a query
 final query = select(items).addColumns([runningTotal]);
 ```

 # Partitioning

 Partition data into groups before applying the window function:

 ```dart
 // Running total per category
 final runningTotalByCategory = WindowFunctionExpression<int>(
   items.price.sum(),
   orderBy: [OrderingTerm.asc(items.date)],
   partitionBy: [items.categoryId],
 );
 ```

 # Frame Boundaries

 ## Rows Frame
 Operates on physical rows:

 ```dart
 // Moving average of previous 3 rows and current row
 final movingAverage = WindowFunctionExpression<double>(
   items.price.avg(),
   orderBy: [OrderingTerm.asc(items.date)],
   boundary: RowsFrameBoundary(
     start: -3, // 3 rows preceding
     end: 0,    // current row
   ),
 );
 ```

 ## Groups Frame
 Operates on groups of rows with same ORDER BY values:

 ```dart
 // Include current group and one group before
 final groupFrame = WindowFunctionExpression<int>(
   items.quantity.sum(),
   orderBy: [OrderingTerm.asc(items.category)],
   boundary: GroupsFrameBoundary(
     start: null, // Unbounded preceding groups
     end: 0,    // current group
   ),
 );
 ```

 ## Range Frame
 Operates on value ranges (default):

 ```dart
 // Sum of items within price range of ±10 from current row
 final rangeFrame = WindowFunctionExpression<int>(
   items.quantity.sum(),
   orderBy: [OrderingTerm.asc(items.price)],
   boundary: RangeFrameBoundary(
     start: -10, // 10 units less than current value
     end: 10,    // 10 units more than current value
   ),
 );
 ```

### DIFF
--- a/drift/lib/src/runtime/query_builder/expressions/window.dart
+++ b/drift/lib/src/runtime/query_builder/expressions/window.dart
@@ -5,6 +5,8 @@ part of '../query_builder.dart';
 /// Window functions perform calculations across a set of table rows that are somehow
 /// related to the current row. This expression class allows you to construct window
 /// function queries with partitioning, ordering, and frame specifications.
+///
+/// More information at [Window Functions](https://www.sqlite.org/windowfunctions.html) documentation.
 class WindowFunctionExpression<T extends Object> extends Expression<T> {
   /// The aggregate or window function to apply (e.g., SUM, AVG).
   final Expression<T> function;
@@ -54,6 +56,8 @@ class WindowFunctionExpression<T extends Object> extends Expression<T> {
   /// EXCLUDE clause, GROUPS frame types, window chaining, and
   /// support for `<expr> PRECEDING` and `<expr> FOLLOWING` boundaries in RANGE frames
   /// are only available from sqlite 3.28.0, released on 2019-04-16.
+  ///
+  /// More information at [Window Functions](https://www.sqlite.org/windowfunctions.html) documentation.
   ///
   /// # Basic Usage
   ///
@@ -248,11 +252,25 @@ enum _FrameType {
 /// Zero indicates the current row
 ///
 /// {@endtemplate}
+///
+/// More information at [FrameBoundary](https://www.sqlite.org/windowfunctions.html#frame_boundaries) documentation.
 sealed class FrameBoundary extends Component {
-  /// The start of the frame boundary.
+  /// The start of the frame boundary, relative to the current row.
+  ///
+  /// A value of [null] indicates that frame includes all prior rows, groups or range bounds.
+  /// A value of 0 indicates that frame starts at the current row.
+  ///
+  /// A negative or positive value indicates that frame starts at the specified offset
+  /// before or after the current row.
   final num? start;
 
-  /// The end of the frame boundary.
+  /// The end of the frame boundary, relative to the current row.
+  ///
+  /// A value of [null] indicates that frame includes all following rows, groups or range bounds.
+  /// A value of 0 indicates that frame ends at the current row.
+  ///
+  /// A negative or positive value indicates that frame ends at the specified offset
+  /// before or after the current row.
   final num? end;
 
   /// The type of frame for the boundary.
@@ -338,7 +356,7 @@ class GroupsFrameBoundary extends FrameBoundary {
   /// Note that GROUPS Frame Type is only available from sqlite 3.28.0, released on 2019-04-16.
   /// Most devices will use an older sqlite version.
   ///
-  /// {@macro boundary}
+  /// {@macro drift_window_boundary}
   const GroupsFrameBoundary({
     int? start,
     int? end = 0,
@@ -357,7 +375,7 @@ class RangeFrameBoundary extends FrameBoundary {
   /// If multiple rows have the same value in the `ORDER BY` column, they
   /// are all included in the frame.
   ///
-  /// {@macro boundary}
+  /// {@macro drift_window_boundary}
   ///
   /// Note that passing value except 0 (CURRENT ROW) or null (UNBOUNDED) to [start] or [end] in RANGE Frame is only available from sqlite 3.28.0, released on 2019-04-16.
   /// Most devices will use an older sqlite version.
--- a/drift/lib/src/runtime/query_builder/expressions/window.dart
+++ b/drift/lib/src/runtime/query_builder/expressions/window.dart
@@ -1,0 +1,371 @@
+part of '../query_builder.dart';
+
+/// Represents a window function expression in SQL.
+///
+/// Window functions perform calculations across a set of table rows that are somehow
+/// related to the current row. This expression class allows you to construct window
+/// function queries with partitioning, ordering, and frame specifications.
+class WindowFunctionExpression<T extends Object> extends Expression<T> {
+  /// The aggregate or window function to apply (e.g., SUM, AVG).
+  final Expression<T> function;
+
+  /// The ordering terms that define how rows are sorted within each partition.
+  /// Must not be empty as window functions require an ORDER BY clause.
+  final List<OrderingTerm> orderBy;
+
+  /// Optional list of expressions to partition the rows by.
+  ///
+  /// When specified, the window function calculations are performed separately
+  /// for each distinct combination of the partition values.
+  final List<Expression>? partitionBy;
+
+  /// Optional frame boundary that defines which rows to include in the
+  /// window calculations relative to the current row.
+  ///
+  /// If not specified, defaults to a RANGE BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW.
+  ///
+  /// Usage:
+  ///
+  /// ```dart
+  /// // RANGE Boundary
+  ///  boundary: RangeFrameBoundary(),
+  ///
+  /// // ROWS Boundary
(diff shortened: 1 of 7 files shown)


drift/lib/src/runtime/query_builder/expressions/window.dart
```diff
@@ -0,0 +1,371 @@
...
+/// If null, the boundary is unbounded.
+///
+/// Negative value indicates a preceding boundary
+///
+/// Positive value indicates a following boundary
+///
+/// Zero indicates the current row
+///
+/// {@endtemplate}
+abstract class FrameBoundary<T extends num> extends Expression<num> {
```
simolus3: Can we make this `sealed`? Also I'm not sure if this should extend `Expression` since it's not a general-purpose expression that could be used in other contexts. You can inherit from `Component` if you just need `writeInto`.

drift/lib/src/runtime/query_builder/expressions/window.dart
```diff
@@ -0,0 +1,371 @@
+part of '../query_builder.dart';
+
+/// Represents a window function expression in SQL.
+///
+/// Window functions perform calculations across a set of table rows that are somehow
+/// related to the current row. This expression class allows you to construct window
+/// function queries with partitioning, ordering, and frame specifications.
```
simolus3, line 7: Let's perhaps also link to the SQLite docs on that syntax here.

drift/lib/src/runtime/query_builder/expressions/window.dart
```diff
@@ -0,0 +1,371 @@
...
+///
+/// Negative value indicates a preceding boundary
+///
+/// Positive value indicates a following boundary
+///
+/// Zero indicates the current row
+///
+/// {@endtemplate}
+abstract class FrameBoundary<T extends num> extends Expression<num> {
+  /// The start of the frame boundary.
```
simolus3: Let's also mention what a `null` value means here and in `end`. So perhaps something like "The start of the frame boundary, relative to the current row, group, or order-by value. A value of `null` includes all prior rows, a value of 0 only includes the current element".

drift/lib/src/runtime/query_builder/expressions/window.dart
simolus3: I also don't think it needs to be generic, we only support ints here right?

simolus3 (commented):
This looks amazing, thank you for the contribution! I think we should also mention this feature in the changelog and on the documentation site, but I can take care of those parts.

drift/lib/src/runtime/query_builder/expressions/window.dart
kedarkarkee: Done

drift/lib/src/runtime/query_builder/expressions/window.dart
kedarkarkee: Done on [5022712](https://github.com/simolus3/drift/pull/3474/commits/5022712a2e6d0629389de17dc0fafddd34505068)

drift/lib/src/runtime/query_builder/expressions/window.dart
kedarkarkee: Done on [1e411f9](https://github.com/simolus3/drift/pull/3474/commits/1e411f9db241ee257429b7b05ec30e946ca8a4a7)

drift/lib/src/runtime/query_builder/expressions/window.dart
kedarkarkee: Yeah, it doesn't need not be generic. Done on [5022712](https://github.com/simolus3/drift/pull/3474/commits/5022712a2e6d0629389de17dc0fafddd34505068) .

According to docs, 

> For Range Boundary, the `<expr>` does not have to be an integer. It can evaluate to a real number as long as it is constant and non-negative.

 So, `RangeFrameBoundary` takes a `num` whereas `Row` and `Group` take an `int`.

drift/lib/src/runtime/query_builder/expressions/window.dart
```diff
@@ -0,0 +1,389 @@
...
+
+/// Rows boundary for the window frame.
+class RowsFrameBoundary extends FrameBoundary {
+  /// Constructs a ROWS frame with the given [start] and [end] boundaries.
+  ///
+  /// A ROWS frame operates on physical rows. The [start] and [end] parameter specifies how
+  /// many rows before and/or after the current row should be included in the frame.
+  ///
+  /// {@macro drift_window_boundary}
+  const RowsFrameBoundary({
```
simolus3: These classes aren't doing anything but instantiate `FrameBoundary` with different frame types. To reduce the API surface, I think we can replace them with named constructors in the `FrameBoundary` class:

```dart
  /// ... existing docs from RowsFrameBoundary constructor
  const FrameBoundary.rows({int? start, int? end = 0, this.exclude}): this._(start, end, _FrameType.rows);
```

(and similar constructors for the other two boundary types)

drift/lib/src/runtime/query_builder/expressions/window.dart
```diff
@@ -0,0 +1,389 @@
...
+    context.writeWhitespace();
+    boundary.writeInto(context);
+    context.buffer.write(')');
+  }
+}
+
+/// A partition-by clause as part of a window function statement. The clause can consist
+/// of multiple [Expression]s, with the first terms being primary partition and
+/// the later terms will work as a nested partition for the previous partition and so on.
+class PartitionBy extends Component {
```
simolus3: Since this class is only used to implement `writeInto`, let's make it private.

simolus3 (commented):
Two more nits about reducing addded API elements. Apart from that, this looks good to me 👍 I'll also write some notes about this on the documentation website and merge this later today.

drift/lib/src/runtime/query_builder/expressions/window.dart
kedarkarkee: The `sealed/abstract` class doesn't allow instantiating. Instead I can rename `FrameBoundary` as private `_BaseBoundary` and create another `FrameBoundary` class with the named constructors as below.

```dart
class WindowFunctionExpression<T extends Object> extends Expression<T> {
// ...properties
  WindowFunctionExpression(
    this.function, {
    required this.orderBy,
    this.partitionBy,
    this.boundary = const FrameBoundary.range(),
  }) {
    if (orderBy.isEmpty) {
      throw ArgumentError.value(
        orderBy,
        'orderBy',
        'Must not be empty',
      );
    }
  }
}

sealed class _BaseBoundary extends Component {
  /// docs
  final num? start;

  /// docs
  final num? end;

  /// The type of frame for the boundary.
  final _FrameType _frameType;

  /// docs
  final FrameExclude? exclude;

  const _BaseBoundary._(
    this.start,
    this.end,
    this._frameType, {
    this.exclude,
  }) : assert(
          start == null || start <= 0 || (end == null || end > 0),
          'Invalid frame specification. A FOLLOWING start boundary must have an end boundary as FOLLOWING or UNBOUNDED.',
        );

  @override
  void writeInto(GenerationContext context) {
    // write into logic
  }

  void _writeBoundary(GenerationContext context, num exp) {
    //  write into logic
  }
}

/// docs
class FrameBoundary extends _BaseBoundary {
  /// ...docs
  const FrameBoundary.rows({
    int? start,
    int? end = 0,
    super.exclude,
  }) : super._(start, end, _FrameType.rows);

  /// ...docs
  const FrameBoundary.groups({
    int? start,
    int? end = 0,
    super.exclude,
  }) : super._(start, end, _FrameType.groups);

  /// ...docs
  const FrameBoundary.range({
    num? start,
    num? end = 0,
    super.exclude,
  }) : super._(start, end, _FrameType.range);
}
```

simolus3: If we move the constructors to `FrameBoundary`, it would no longer have to be `sealed` or `abstract` (in fact, better to make it `final`).

drift/lib/src/runtime/query_builder/expressions/window.dart
kedarkarkee: Done [e91953a](https://github.com/simolus3/drift/pull/3474/commits/e91953a15483c508dff97f99c03c83d7b7cac2ea)

drift/lib/src/runtime/query_builder/expressions/window.dart
kedarkarkee: Done [e91953a](https://github.com/simolus3/drift/pull/3474/commits/e91953a15483c508dff97f99c03c83d7b7cac2ea)

PR #3479, opened 2025-02-20
Fix typo that refers to an non-existent internal table

PIG208:
This was slightly misleading as the implementation suggests that 'sqlite_' is the prefix for these tables.

This was brought up here:
  https://github.com/zulip/zulip-flutter/pull/1248#discussion_r1961002211


simolus3 (approved):
Thanks!

PR #3510, opened 2025-03-25
Finish migrating drift_dev to element2.

stereotype441:
This change fixes several uses of the old analyzer element model that
are internal to drift_dev so that they now use the new element model.
    
There should be no functional change.

### DIFF
--- a/drift_dev/analysis_options.yaml
+++ b/drift_dev/analysis_options.yaml
@@ -5,3 +5,7 @@ analyzer:
     strict-casts: true
   exclude:
     - "**/*.g.dart"
+
+linter:
+  rules:
+    analyzer_use_new_elements: true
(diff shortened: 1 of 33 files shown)


stereotype441:
@simolus3 would you mind reviewing this?

I'm not sure why there is a check failure; I don't think it's related to this PR.

simolus3:
Thanks, I'll take a look 👍

I don't think we'll get the CI green here, it has to pull packages from pub and most of the build / `source_gen` dependencies haven't been released yet. I still see lots of failing tests in `drift_dev` with these changes, I'll see what I can do about them and then I'll merge it into `element2` so that you could use that as an upstream target.

simolus3:
I have applied your changes in e9c1db1233b376b2c9ed30a4e303c9eb33805635. Didn't get all tests to pass today, but I think `drift_dev` is in a state where it should mostly work now.

stereotype441:
> I have applied your changes in [e9c1db1](https://github.com/simolus3/drift/commit/e9c1db1233b376b2c9ed30a4e303c9eb33805635). Didn't get all tests to pass today, but I think `drift_dev` is in a state where it should mostly work now.

Great, thank you! And thanks for the fixes.

simolus3:
These changes have since been cherry-picked onto `element2` and later `develop`.

PR #3514, opened 2025-03-29
Expose orderBy in groupConcat

alexmercerind:


alexmercerind:
I didn't quite understand "Most devices will use an older sqlite version." in existing "drift_aggregate_filter" comment. Doesn't Drift ship its own version of SQLite?

simolus3 (approved):
Thanks!

simolus3:
> Doesn't Drift ship its own version of SQLite?

Not necessarily, only when using `drift_flutter` which depends on `sqlite3_flutter_libs`. So the warning is still warranted since some users might be using sqflite or the SQLite from the system.

PR #3518, opened 2025-04-04
Add Non nulls filter

### LINKED ISSUE #3505: isNotNull() is missing in managers filter (only isNull() is available)
### Discussed in https://github.com/simolus3/drift/discussions/3503

<div type='discussions-op-text'>

<sup>Originally posted by **iamiota** March 24, 2025</sup>
How can I filter a `DateTime` field that is `not null` with managers?
Screenshot 2025-03-24 at 17 25 50

`await ref.read(AppDatabase.provider).managers.test.filter((row) => row.deletedAt.isNotNull()).get();`
</div>

There should be a filter for this built in

dickermoshe:
Closes https://github.com/simolus3/drift/issues/3505


simolus3 (approved):
Thanks!

PR #3523, opened 2025-04-08
Adapt to renamed methods in source_gen.

stereotype441:
When a version of the `source_gen` package is published that works with the new analyzer element model, it will not use `2` suffixes on the methods `spanForElement`, `isAssignableFrom`, and `isExactly`. (See
https://github.com/dart-lang/source_gen/pull/750). Accordingly, `drift_dev` need to refer to these methods without the `2` suffix.


stereotype441:
@simolus3 Here's one more small change that will be needed for the new analyzer element model.

simolus3 (approved):
Thanks!

PR #3534, opened 2025-04-16
Prepare for upcoming element2 release of package:build.

davidmorgan:
The "build" packages will be released, like source_gen, with element2 APIs under the same names as the current element1 APIs.

This should now be approximately the final state that will work after source_gen+build are released for element2.

I added some dependency overrides so it's actually possible to analyze against the work-in-progress releases.


simolus3:
Thanks!

PR #3537, opened 2025-04-16
make query builder const

dickermoshe:
This PR migrates the query builder API to be deeply const. 

It uses `built_collection` for collections. These are only used internally. 

Future work:
- Make the dependency on built_collection more flexible
- If a public API has a list argument, BuiltList or List.unmodifiable should be used to ensure that future mutations don't break anything.
- Users only supply `Iterables`, no lists
- Fields which contain Built* classes should be marked private

Coderabbit: https://github.com/dickermoshe/drift/pull/1#pullrequestreview-2774504356


drift/lib/src/connections/result_set.dart
```diff
@@ -9,19 +10,21 @@ import 'package:collection/collection.dart';
 /// &lt;expr&gt; AS c1`) and also remembers its position ([index]).
 typedef ColumnPosition = ({String name, int index});
 
+@immutable
```
simolus3, line 13: It's a `final` class with a `const` constructor now, I don't think the `@immutable` annotation contributes anything here.

drift/lib/src/query_builder/schema/column.dart
```diff
@@ -18,22 +21,37 @@ base class SchemaColumn<T extends Object> extends Expression<T> {
...
 
-  SchemaColumn({
+  const SchemaColumn({
     required this.name,
     required this.type,
     this.isNullable = true,
+    this.owningResultSet,
   });
 
+  SchemaColumn<T> copyWith({
```
simolus3, line 33: `copyWith` on a type with subclasses doesn't really work (or we'd have to override this for the table column implementation too).

Given that `SchemaColumn` is mostly constructed internally and de-facto immutable outside of `owningResultSet` which is only `late` due to the circular reference, I'm not sure if we need `copyWith` here. My preferred approach would be to make `owningResultSet` `late final` - users are not going to create their own `SchemaColumn` instances (maybe we can make the constructor `@internal` to spell it out).

drift/lib/src/query_builder/schema/column_constraints.dart
```diff
@@ -13,6 +15,7 @@ abstract base class ColumnConstraint implements SqlComponent {
   }
 }
 
+@immutable
```
simolus3, line 18: Basically my opinion is that `@immutable` is helpful for abstract classes (to get a lint if subclasses aren't), but for `final` classes that are already immutable, it's just noise.

drift/lib/src/query_builder/schema/entities.dart
```diff
@@ -1,15 +1,22 @@
...
 import '../statements/statement.dart';
 
 /// Some abstract schema entity that can be stored in a database. This includes
 /// tables, triggers, views, indexes, etc.
+@immutable
 abstract interface class DatabaseSchemaEntity {
   /// The (unalised) name of this entity in the database.
   String get entityName;
+
+  const DatabaseSchemaEntity();
```
simolus3, line 14: Remove, this is an `interface class`

drift/lib/src/query_builder/schema/result_set.dart
```diff
@@ -8,20 +11,21 @@ import '../types.dart';
...
-  List<SchemaColumn> get columns;
+  Iterable<SchemaColumn> get columns;
 
   String get aliasOrName => alias ?? entityName;
 
-  late final Map<String, SchemaColumn> columnsByName = {
+  late final BuiltMap<String, SchemaColumn> columnsByName = {
     for (final column in columns) column.name: column,
-  };
+  }.build();
```
simolus3, line 25: For things like these, I don't see what value `BuiltMap` provides over say an `UnmodifiableMapView`. Also, we should be very careful when making the public type a `BuiltMap` because changing it back to a regular map is now a breaking change.

So I think this should ideally be:

```dart
  late final Map<String, SchemaColumn> columnsByName = UnmodifiableMapView({
    for (final column in columns) column.name: column,
  });
```

drift/lib/src/query_builder/results.dart
```diff
@@ -10,58 +13,78 @@ import 'schema/column.dart';
...
 final class ResultSetStructure {
   /// For [Expression] instances added to a query, the position of the column
   /// added for that expression.
-  final Map<Expression, ColumnPosition> expressions;
+  final BuiltMap<Expression, ColumnPosition> expressions;
 
   /// For each [ResultSet] that has been added to a query in its entirety, the
   /// a list of [ColumnPosition]s for each column in the result set.
-  final Map<ResultSet, List<ColumnPosition>> tables;
+  final BuiltMap<ResultSet, BuiltList<ColumnPosition>> tables;
```
simolus3, line 24: I know you've kept this as a point for later, but I think we should be careful from the start not to leak `built_collection` types into our API.

For instance, here we could make `expressions` and `tables` a private field with their current type and then add a `Map<Expression, ColumnPosition> get expressions => _expressions;` getter. This means that `ResultSetStructure` can only be copied efficiently through the methods we provide here, but that may be fine.

drift/lib/src/query_builder/schema/result_set.dart
```diff
@@ -8,20 +11,21 @@ import '../types.dart';
 import 'column.dart';
 import 'entities.dart';
 
+@immutable
 mixin ResultSet<Row extends Object, Self extends ResultSet<Row, Self>>
     implements ResultSetDsl, DatabaseSchemaEntity {
   String? get alias;
 
-  List<SchemaColumn> get columns;
+  Iterable<SchemaColumn> get columns;
```
simolus3, line 19: I might be missing something, but why is this now an `Iterable`?

drift/lib/src/query_builder/statements/delete.dart
```diff
@@ -17,25 +22,53 @@ final class DeleteStatement<Row extends Object,
...
+  @override
+  DeleteStatement<Row, RS> withWhereClause(WhereClause whereClause) =>
+      _copyWith(whereClause: whereClause);
+
+  DeleteStatement<Row, RS> _withReturning() =>
+      _copyWith(returning: ReturningClause(resultSet));
+
+  DeleteStatement<Row, RS> _copyWith({
+    GeneratedTable<Row, RS>? resultSet,
+    WhereClause? whereClause,
```
simolus3, line 54: The problem with these kinds of `copyWith` methods is that you can't set fields back to `null`.

This is why I would suggest an approach where:

1. There is no public API to mutate statements (so no mutable `WhereClause? whereClause;`).
2. But, `WhereClause?` is a mutable field internally, encapsulated through a getter.
3. Then, when we want to add a where clause, we do something like `DeleteStatement<Row, RS> where(Expression<bool> expression) => copy().._whereClause = ...`.

simolus3 (commented):
I think this is a good approach, but I see some flaws worth discussing. I'm happy to address them myself / take parts of this branch into `new-query-builder` myself. So don't consider this as something you'd have to change, this is is just to continue the discussion.

First, I think we should make our classes "publicly immutable", meaning that users can't do things like `select(users)..whereClause = ...`. For safety, we also need to copy internally in methods like `where()`, but I think the `copyWith` pattern is generally quite dangerous when nullable fields are involved and should be involved. So the correct internal implementation to add a `where` clause is:

```dart
UpdateStatement<Row, RS> where(Expression<bool> condition) {
  final copy = clone();
  // Internal mutability to implement these specialized copyWith methods
  copy._where = switch(copy.where) {
    null => Where(condition),
    var where => where.withAddedCondition(condition),
  };
}
```

I also realized halfway through the review that the `built_collection`s classes don't implement the Dart SDK interfaces. That invalidates half of my review comments, but IMO makes them very unsuitable for this. I think the way to go is:

1. Regular `Map`, `List`, ... types in public methods / getters.
2. Use a mutable map/list internally with care and expose with `UnmodifiableMapView`. Make sure our methods copy instead of modifying the internal map.

Also another thing that occurred to me: Making the classes immutable breaks backwards compatibility. Because right now, the style to write queries is `(select(users)..where(someCondition)..orderBy([someClauses])).get()`. This exact snippet works with the new query builder too, but it would stop working as soon as we make the structures immutable (because the `where` and `orderBy` are now silently dropped).
I think breaking backwards compatibility in a way that unchanged code continues to compile while doing something completely different is really bad. So maybe the solution is to keep mutable APIs and embrace the `..` style of operators, even though it can be cumbersome in a lot of cases.

drift/lib/src/query_builder/statements/delete.dart
dickermoshe: There are very few fields which are mutated. The solution for this is to create a `copyWith*` fields for each field. This would allow you to set a value back to null.
```dart
class Foo {
  final int bar;
  Foo copyWithBar(int? newBar){
    return Foo(newBar);
  }
}
```
There are not _that_ many fields, it should not be too big of an issue.

drift/lib/src/query_builder/schema/result_set.dart
dickermoshe: Because `BuiltList` implements `Iterable`, not `List`.
`List` is inherently a mutable class. It has methods such as `add` and `addAll`. If external libraries should not be mutating this list directly, they should only get an `Iterable`.

drift/lib/src/query_builder/results.dart
dickermoshe: You are still returning a reference to the original Map, there is no copy there.

drift/lib/src/query_builder/schema/result_set.dart
dickermoshe: See the comment below

drift/lib/src/query_builder/schema/column_constraints.dart
dickermoshe: Do you mean `const` classes? `final` classes don't enforce immutability.
I'll remove `@immutable` from `const` classes if you want.
However, many times `const` needs to be removed fto make the constructor work and we may forget to add a `@immutable` later.

It's your call.

drift/lib/src/connections/result_set.dart
dickermoshe: https://github.com/simolus3/drift/pull/3537#discussion_r2049665120

dickermoshe:
<details><summary>Details</summary>
<p>


> For safety, we also need to copy internally in methods like where(), but I think the copyWith pattern is generally quite dangerous when nullable fields are involved and should be involved.

https://github.com/simolus3/drift/pull/3537#discussion_r2049653989

> Regular Map, List, ... types in public methods / getters.
    Use a mutable map/list internally with care and expose with UnmodifiableMapView. Make sure our methods copy instead of modifying the internal map.

I tried this. There are 3 main issues.
1. `UnmodifiableMapView` extends `Map`, this means that all the mutating function on `Map` still exist on `UnmodifiableMapView`. All it does is make everything crash at runtime. `BuiltMap` don't contain these methods so everything is compile time safe.
2. Copying everything ourself has 2 sub-issues:
    a. You would have tons of unneeded copies. built_collection is smart enough to only copy a list once a mutation is performed. Because we/users rarely mutate the original list, this will lead to many, many wasted copied, which are very expensive.
    b. It is really, really hard to locate every instance of a collection being passed around and remembering to copy it. I tried doing this without built_collection/fast_immutable_collections at first. After switching to a 3rd party package I was shocked how many instances of missing copies there were.

> Also another thing that occurred to me: Making the classes immutable breaks backwards compatibility. Because right now, the style to write queries is (select(users)..where(someCondition)..orderBy([someClauses])).get(). This exact snippet works with the new query builder too, but it would stop working as soon as we make the structures immutable (because the where and orderBy are now silently dropped).

You cannot have your cake and eat it too. 
Either have `..` and we'll have an API which users recognize to be mutable, or have a single `.` which users expect it to be immutable.
What you're proposing would be a nightmare to use.


</p>
</details> 

> I think breaking backwards compatibility in a way that unchanged code continues to compile while doing something completely different is really bad.

This is a really solid point. Mutability it is.

IMHO, we should use the pattern that code_builder uses. See discord.

drift/lib/src/query_builder/statements/delete.dart
dickermoshe: N/A

drift/lib/src/query_builder/results.dart
dickermoshe: N/A

drift/lib/src/query_builder/schema/result_set.dart
dickermoshe: N/A

drift/lib/src/query_builder/schema/entities.dart
dickermoshe: N/A

drift/lib/src/query_builder/schema/result_set.dart
dickermoshe: N/A

drift/lib/src/query_builder/schema/column_constraints.dart
dickermoshe: N/A

drift/lib/src/query_builder/schema/column.dart
dickermoshe: N/A

drift/lib/src/connections/result_set.dart
dickermoshe: N/A

PR #3538, opened 2025-04-18
Make query API only mutable

dickermoshe:
In the future split select and customSelect.


drift/lib/src/query_builder/statements/query.dart
```diff
@@ -15,14 +15,12 @@ abstract mixin class SingleTableStatementMixin<
     RS extends ResultSet<Row, RS>,
     Self extends SingleTableStatementMixin<Row, RS, Self>> {
```
simolus3: Is there another place where `Self` is still used now? If not, I think we can get rid of this.

simolus3 (approved):
LGTM, thanks for following up on this

PR #3543, opened 2025-04-28
Add file path to error message in make-migrations script

Mr-Pepe:
During development, migrations may need to be built repeatedly for the same schema version. `dart run drift_dev make-migrations` fails if a schema file with different content already exists for a schema version and suggests deleting it. However, it is not clear which exact file needs to be deleted.

This change adds the schema file path to the error message.


simolus3 (approved):
Thanks!

PR #3613, opened 2025-07-17
Update example code for daos

JeanJPNM:
Marks category as nullable, removes the use of the deprecated isNull function and simplifies the name of the part file.


simolus3 (approved):
Thanks!

PR #3634, opened 2025-08-24
added  initializeDatabase to web options

rohitlymrik:
added  initializeDatabase to web options to pass it to open function.

### DIFF
--- a/drift_flutter/lib/src/connect.dart
+++ b/drift_flutter/lib/src/connect.dart
@@ -6,6 +6,9 @@ export 'unsupported.dart'
     if (dart.library.ffi) 'native.dart';
 
 export 'package:drift/src/web/wasm_setup/types.dart';
+import 'dart:async';
+import 'dart:typed_data';
+
 // ignore: implementation_imports
 import 'package:drift/src/web/wasm_setup/types.dart';
 import 'package:sqlite3/common.dart';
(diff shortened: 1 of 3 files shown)


drift_flutter/lib/src/connect.dart
```diff
@@ -33,12 +36,14 @@ final class DriftWebOptions {
...
-  DriftWebOptions({
-    required this.sqlite3Wasm,
-    required this.driftWorker,
-    this.onResult,
-  });
+  DriftWebOptions(
+      {required this.sqlite3Wasm,
+      required this.driftWorker,
+      this.onResult,
+      this.initializeDatabase});
```
simolus3: Let's add a trailing comma after `this.initializeDatabase` to keep the format.

drift_flutter/pubspec.yaml
```diff
@@ -4,7 +4,6 @@ version: 0.2.5
 repository: https://github.com/simolus3/drift
 homepage: https://drift.simonbinder.eu/
 issue_tracker: https://github.com/simolus3/drift/issues
-resolution: workspace
```
simolus3: Did you run into issues without this? I think the package should still be part of the workspace.

simolus3 (commented):
Thanks for the contribution! Could you also add a changelog entry to `drift_flutter`?

drift_flutter/pubspec.yaml
rohitlymrik: yes it should. my miss and this is a local change and not for repository.

drift_flutter/lib/src/connect.dart
rohitlymrik: done.

simolus3 (approved):

PR #3641, opened 2025-09-02
feat(mariadb):enableMigrations

easychen:
Add enableMigrations to  lib/services/mariadb_database.dart make it same as other db provider

```dart
static Future<MariaDBDatabaseImpl> connect({
    required String host,
    required int port,
    required String database,
    required String user,
    required String password,
  }) async {
    final pool = MySQLConnectionPool(
      host: host,
      port: port,
      userName: user,
      password: password,
      databaseName: database,
      maxConnections: 10
    );

    final mariadbDatabase = MariaDBDatabase(pool: pool, enableMigrations: false);
    return MariaDBDatabaseImpl(mariadbDatabase);
  ```

### DIFF
--- a/extras/drift_mariadb/lib/src/mariadb_database.dart
+++ b/extras/drift_mariadb/lib/src/mariadb_database.dart
@@ -9,8 +9,9 @@ class MariaDBDatabase extends DelegatedDatabase {
     required MySQLConnectionPool pool,
     bool isSequential = true,
     bool logStatements = false,
+    bool enableMigrations = true,
   }) : super(
-          _MariaDelegate(() => pool, true),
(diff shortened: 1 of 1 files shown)


simolus3 (commented):
This looks good to me, thanks for the contribution! Do you want to add an entry to the changelog? I can do it otherwise.

easychen:
> This looks good to me, thanks for the contribution! Do you want to add an entry to the changelog? I can do it otherwise.

I'm not very familiar with the format, so it would be great if you could add it to the changelog. Thanks!

simolus3:
... I just realized we're not publishing these packages and they don't have a changelog, so we don't need one for this.

simolus3 (approved):

PR #3657, opened 2025-09-18
Avoid incorrect cast in sql_js in wasm builds

ebraminio:
In order to avoid:

    Type 'double' is not a subtype of type 'int' in type cast

Raised when drift is built with `flutter run -d chrome --wasm` it's needed to first cast the value to double then turn it into int and it does makes sense as int isn't native type in JS while double is the default numeric type in JS.

Fixing this make these errors in an external project (stream-chat) go:

image

### DIFF
--- a/drift/lib/src/web/sql_js.dart
+++ b/drift/lib/src/web/sql_js.dart
@@ -114,7 +114,7 @@ class SqlJsDatabase {
 
   /// Returns the `user_version` pragma from sqlite.
   int get userVersion {
-    return (_selectSingleRowAndColumn('PRAGMA user_version;') as double)
+    return (_selectSingleRowAndColumn('PRAGMA user_version;') as num)
         .toInt();
   }
 
@@ -153,7 +153,7 @@ class SqlJsDatabase {
   /// [export].
   int lastInsertId() {
     // load insert id. Will return [{columns: [...], values: [[id]]}]
-    return (_selectSingleRowAndColumn('SELECT last_insert_rowid();') as double)
+    return (_selectSingleRowAndColumn('SELECT last_insert_rowid();') as num)
         .toInt();
   }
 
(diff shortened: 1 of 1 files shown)


drift/lib/src/web/sql_js.dart
```diff
@@ -114,7 +114,8 @@ class SqlJsDatabase {
 
   /// Returns the `user_version` pragma from sqlite.
   int get userVersion {
-    return _selectSingleRowAndColumn('PRAGMA user_version;') as int;
+    return (_selectSingleRowAndColumn('PRAGMA user_version;') as double)
```
simolus3:
```suggestion
    return (_selectSingleRowAndColumn('PRAGMA user_version;') as num)
```

drift/lib/src/web/sql_js.dart
```diff
@@ -152,7 +153,8 @@ class SqlJsDatabase {
   /// [export].
   int lastInsertId() {
     // load insert id. Will return [{columns: [...], values: [[id]]}]
-    return _selectSingleRowAndColumn('SELECT last_insert_rowid();') as int;
+    return (_selectSingleRowAndColumn('SELECT last_insert_rowid();') as double)
```
simolus3:
```suggestion
    return (_selectSingleRowAndColumn('SELECT last_insert_rowid();') as num)
```

simolus3 (commented):
Let's use `num` here, apart from that this looks good to me.

ebraminio:
> Let's use num here, apart from that this looks good to me.

Makes sense. Confirmed it works in my case so applied it in a separate commit.

ebraminio:
Resolved dartfmt complain which was existed before this change AFAICS

simolus3:
I've merged this manually in cdd61503eb2ea2cbbf43110702bf973b7e06be18 - thanks for the contribution!

Btw, I still recommend migrating to the newer [`package:drift/wasm.dart` library](https://drift.simonbinder.eu/platforms/web/#migrating-from-packagedriftwebdart). `package:drift/web.dart` is not tested with dart2wasm.

PR #3672, opened 2025-09-27
Expose similar interface to computeWithDatabase on web

Cunibon:
I created this PR after I learned about the computeWithDatabase function from Simon and was dissapointed that it does not take over handling web for me.
Since all the neccessarry parts are already there I just split up the isolate.dart file and export the correct implementation from isolate.dart depending on the platform.

All of this is just for convenience and parity with Isolate.run.

- Split up isolate.dart file
- Added web extension


simolus3 (commented):
Thank you for your contribution! I see the value of having something like `computeWithDatabase` on the web, even if it doesn't really do anything.

I'm less sure however about letting people import `package:drift/isolate.dart` on the web, I think that should be a native-only library without platform-specific imports.

What if we make `computeWithDatabase` a method on `GeneratedDatabase`, so that it's available on all platforms? And then in that method we can use the isolate implementation on native platforms and a direct call on the web?

Cunibon:
Just to clarify, would we move `serializableConnection` and `computeWithDatabase` into the `GeneratedDatabase` and then use a platform specific import here to call the correct function?

Or would we change the extensions to point at `GeneratedDatabase` instead of `DatabaseConnectionUser` and basically keep the same logik with the conditional export?

If its the latter would we move the extension to the api folder of db_base.dart

simolus3:
I should have been a bit clearer here: For compatibility with tools like `build_web_compilers`, we shouldn't even import `dart:isolate` on cross-platform libraries like `drift/drift.dart`. This is the same reason I don't want `drift/isolate.dart` to offer cross-platform extensions, just having APIs that depend on e.g. `SendPort` in their public interface is enough to break compilation with `build_web_compilers`.

So `serializableConnection` can't be moved to `GeneratedDatabase`, because `Isolate` (and by extension `DriftIsolate`) is not a type available in shared code. But my idea would be to:

1. Add a `computeWithDatabase()` method to `GeneratedDatabase`, it should delegate to an internal top-level method (say `computeWithDatabaseImplementation`) we don't export.
2. We then use the existing conditional-imports trick to make `computeWithDatabaseImplementation` use the isolates API on native platforms and to just call the function directly on the web.

Cunibon:
Sorry for the late update.

I hope I applied your comment correctly, let me know if you have any more input!

I was a bit unsure how to handle `serializableConnection`. We could also pull it out of isolate.dart and move it into the native implementation, calling it with a reference to the `GeneratedDatabase`, similar to how `computeWithDatabaseImplementation` is setup now.

drift/lib/src/runtime/api/compute_with_database_implementation/compute_with_database_native.dart
```diff
@@ -0,0 +1,75 @@
...
+/// drift will already use an isolate to run your SQL statements. Using
+/// [computeWithDatabase] is beneficial when an an expensive work unit needs
+/// to use the database, or when creating the SQL statements itself is
+/// expensive.
+/// In particular, note that [computeWithDatabase] does not create a second
+/// database connection to sqlite3 - the current one is re-used. So if you're
+/// using a synchronous database connection, using this method is unlikely to
+/// take significant loads off the main isolate. For that reason, the use of
+/// `NativeDatabase.createInBackground` is encouraged.
+Future<Ret>
```
simolus3, line 9: Since this is an internal method, we can remove the documentation comment.

drift/lib/src/runtime/api/compute_with_database_implementation/compute_with_database_unsupported.dart
```diff
@@ -0,0 +1,13 @@
+import 'dart:async';
```
simolus3, line 1: Maybe call this file `_unsupported` instead of `_web`.

drift/lib/src/runtime/api/runtime_api.dart
```diff
@@ -9,6 +9,9 @@ import 'package:meta/meta.dart';
 import '../devtools/devtools.dart' as devtools;
 import '../query_builder/query_builder.dart';
 
+import 'compute_with_database_implementation/compute_with_database_native.dart'
+    if (dart.library.js) 'compute_with_database_implementation/compute_with_database_web.dart';
```
simolus3: Can we switch these imports around? Let's import `_web` by default (but give it another name) and then `if (dart.library.isolate)` we import the `_native` implementation?

drift/lib/src/runtime/api/db_base.dart
```diff
@@ -151,4 +151,64 @@ abstract class GeneratedDatabase extends DatabaseConnectionUser
       return true;
     }());
   }
+
+  /// On native platforms this spawns a short-lived isolate to run the [computation] with a drift
+  /// database.
+  /// On web platforms this will run the [computation] on the current eventloop.
```
simolus3:
```suggestion
  /// On web platforms, this will run the [computation] on the current JavaScript context.
```

simolus3 (commented):
I like the structure of the implementation now 👍 

I don't think we have to migrate `serializableConnection()` as well, that doesn't really make sense outside of native platforms.

Cunibon:
Had to do a rebase since the generated file was also modified on "develop" after I created the branch, sorry for the inconvenience.

drift/lib/src/runtime/api/runtime_api.dart
simolus3: Sorry, I think this suggestion is what's causing the integration tests with `build_web_compilers` to fail. My bad! Can you try reverting this to what you had originally (except that `dart.library.js` should be replaced with `dart.library.js_interop`)?

Cunibon:
From what I can tell this is a mime type error? Seems a bit random given the changes made.

I tried running `dart compile js main.dart -o build/main.dart2js.js` locally to see if that throws for some reason but that seems to work fine

simolus3:
> From what I can tell this is a mime type error? Seems a bit random given the changes made.

Earlier in the logs there's a compiler error because `dart2js` can't find the native implementation file. This is likely a bug in `build_web_compilers`, which doesn't even make the native file available because it thinks `dart:isolate` is not available on the web and thus doesn't have to be read. `dart2js` makes `dart:isolate` available with a stub implementation though, and thus seems to pick the wrong implementation file? At least that's my guess without taking a deeper look at it.

Cunibon:
Ah yes, very straigtforward xd

So switching back to `dart.library.js_interop` should do the trick then, right?
I will quickly push the change.

simolus3:
> So switching back to `dart.library.js_interop` should do the trick then, right?

Well let's hope so :D

simolus3:
Ok I think that's a flake, let me try to restart the action

simolus3 (approved):
Thank you for your contribution!

PR #3679, opened 2025-09-30
Add devtool for downloading db

ghost:
- Adds a devtool action `download` that returns the base64-encoded data and the db name.
- Adds a "Download database" button that prompts the user for a location, then saves the sqlite db to that file. This only works in vscode. In the browser, it won't prompt and will just write to the user's downloads directory.

Had a few questions:
- Should I ensure that this is only shows for sqlite databases? I'm not super sure if this would work correctly for other supported databases
- Is it safe to `import 'package:web/web.dart'` in the extension? Do I need to add this to the pubspec as a dependency or is it already implied because it's a devtool?
   - I'd like to use ExportController in devtools_app, but it's not in devtools_app_shared  (https://github.com/flutter/devtools/issues/9451)

I ran build.sh and committed the files on a separate branch, if you'd like to scrappily test it:
```yaml
dependency_overrides:
  drift:
    git:
      url: https://github.com/kevinhikaruevans/drift-tester.git
      ref: test
      path: drift/
```

### DIFF
--- a/drift/lib/src/runtime/devtools/service_extension.dart
+++ b/drift/lib/src/runtime/devtools/service_extension.dart
@@ -1,13 +1,13 @@
 import 'dart:async';
 import 'dart:convert';
 import 'dart:developer';
-import 'dart:io';
 
 import 'package:drift/drift.dart';
 import 'package:drift/src/remote/protocol.dart';
 import 'package:drift/src/runtime/executor/transactions.dart';
 import 'package:meta/meta.dart';
-import 'package:path/path.dart' as p;
+
+import 'platform_unsupported.dart' if (dart.library.io) 'platform_native.dart';
 
 import '../api/runtime_api.dart';
 import 'devtools.dart';
@@ -28,22 +28,11 @@ class DriftServiceExtension {
 
     switch (action) {
       case 'download':
-        final destination = p.join(Directory.systemTemp.path,
-            "${DateTime.now().toUtc().millisecondsSinceEpoch}.tmp");
-        final database = tracked.database;
-
-        await database.exclusively(() async {
-          await database.customStatement('VACUUM INTO ?;', [destination]);
-        });
-
-        final file = File(destination);
-        final encoded = base64Encode(await file.readAsBytes());
-
-        await file.delete();
+        final exported = await exportDatabase(tracked.database);
 
         return {
-          'database': database.runtimeType.toString(),
-          'data': encoded,
+          'database': tracked.database.runtimeType.toString(),
+          'data': base64.encode(exported),
         };
       case 'subscribe-to-tables':
         final stream = tracked.database.tableUpdates();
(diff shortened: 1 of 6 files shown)


drift/lib/src/runtime/devtools/service_extension.dart
```diff
@@ -1,11 +1,14 @@
 import 'dart:async';
 import 'dart:convert';
 import 'dart:developer';
+import 'dart:io';
```
simolus3: This would probably break web support - but you can let me worry about finding a good solution on the web.

drift/lib/src/runtime/devtools/service_extension.dart
```diff
@@ -1,11 +1,14 @@
...
 
 import 'package:drift/drift.dart';
 import 'package:drift/src/remote/protocol.dart';
 import 'package:drift/src/runtime/executor/transactions.dart';
 import 'package:meta/meta.dart';
+import 'package:path/path.dart' as p;
+import 'package:path_provider/path_provider.dart';
```
simolus3: Drift must not depend on Flutter-only packages like `path_provider`.

Can we use `Directory.systemTemp` from `dart:io` instead? (that still leaves web support behind but we can figure that one out later).

simolus3 (commented):
Just some early feedback. Wiring this up to support web databases will probably require some tricks because we don't necessarily have access to the file system hosting the database in the service extension (it could be in a worker).

This generally looks quite nice though 👍

simolus3 (commented):
Since this wouldn't work on the web (I've added it to the big pile of improvements I'll hopefully be able to focus soon-ish), I've used conditional imports to throw an `UnsupportedError` right away.

I noticed that if the call throws, the current error handling is a bit poor. You can try that by running say `example/app` with `flutter run -d chrome`, then starting the extension with `flutter run -d chrome --dart-define=use_simulated_environment=true` and pasting the devtools URL from the first extension. Attempting to download the database prints a stacktrace to the console, I think it would be nice if we could show say a snackbar or a toast with the error message.

ghost-458:
Would it be possible to hide the download button if downloads aren't supported?

simolus3:
That's also a good idea - I'm not sure if we can get the target platform through the VM service protocol (haven't looked though). If not we could introduce a drift-specific method returning whether downloads should be available (essentially checking for `kIsWeb` with conditional imports).

ghost-458:
How about:
- creating a variable (or function) called `isExportSupported` in platform_native/platform_unsupported
- creating a message `get-supported-features` that returns `isExportSupported`
- in the extension, in details.dart, use a FutureBuilder that calls `get-supported-features`
   - if exports are supported, display the button as usual
   - if not supported, disable the download button w/ a tooltip (`DevToolsTooltip`) that says the feature is not supported for this database

simolus3:
That approach sounds good to me :+1:

simolus3:
Thanks for your contribution!

PR #3705, opened 2025-11-06
docs: Update TodosDao constructor to use attachedDatabase

developerjamiu:
Drift 3 now uses `attachedDatabase` instead of the previous `db`


simolus3:
Thanks for the contribution! I've used this opportunity to move the inline snippet into a proper Dart file that gets included (e5149c6d0860baff4be40f9a2757fdf8006df3de). That also contains your changes.

PR #3707, opened 2025-11-08
Fix range error in page caching

ViveNoctem:
When I search for "conflict" on the drift documentation page, I get an SqliteException.

It looks like the `cachedPages` array is accessed in the if condition before the bounds are checked.


simolus3 (approved):
Thanks!

PR #3708, opened 2025-11-09
Fix scanner if source ends with an unterminated string

ViveNoctem:
Fixed range error if the scanner input ended with a single apostrophe.
Returned value when parsing unterminated string literal was one character short.

### DIFF
--- a/sqlparser/test/scanner/single_token_test.dart
+++ b/sqlparser/test/scanner/single_token_test.dart
@@ -74,14 +74,32 @@ void main() {
     expect(scanner.tokens[1].type, TokenType.eof);
   });
 
-  test('issues error for unterminated string literals', () {
-    final scanner = Scanner("'unterminated")..scanTokens();
+  group("parse unterminated string literals", () {
+    test('issues error', () {
+      final scanner = Scanner("'unterminated")..scanTokens();
 
-    expect(
-      scanner.errors,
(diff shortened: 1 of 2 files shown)


sqlparser/test/scanner/single_token_test.dart
```diff
@@ -74,14 +74,33 @@ void main() {
...
-      scanner.errors,
-      contains(const TypeMatcher<TokenizerError>()
-          .having((e) => e.message, 'message', 'Unterminated string')),
-    );
+      expect(scanner.tokens, isNotEmpty);
+      expect(scanner.tokens[0], isA<StringLiteralToken>());
+      expect(
+        (scanner.tokens[0] as StringLiteralToken).value,
+        equals("unterminated"),
+      );
```
simolus3: Nit: Use a single `expect` call:

```suggestion
      expect(scanner.tokens, [
        isA<StringLiteralToken>()
            .having((e) => e.value, 'value', 'unterminated'),
        isA<Token>().having((e) => e.type, 'type', TokenType.eof),
      ]);
```

simolus3 (approved):
Thanks again, nice catch!

sqlparser/test/scanner/single_token_test.dart
ViveNoctem: I added it to the commit

PR #3720, opened 2025-12-05
Migrate IndexedDB databases to OPFS

simolus3:
This builds ontop of the work started by @Cunibon in https://github.com/simolus3/drift/pull/3695.

The main changes are that this variant:

- Does not expose a general API to move between web storage API implementations, only for IndexedDB to OPFS.
- Copies both the main database and a journal file if it exists.
- Copies on the client side before opening the database instead of doing this via an initialize callback.
- Deletes the old IndexedDB database after moving it to OPFS.

@Cunibon, does this also work for you? Let me know if this API is too restrictive for you. Otherwise I'd release this variant.

### DIFF
--- a/drift/CHANGELOG.md
+++ b/drift/CHANGELOG.md
@@ -1,6 +1,9 @@
 ## 2.30.0-wip
 
 - Add `rightOuterJoin` and `fullOuterJoin`.
+- Wasm: Add the `WasmProbeResult.moveFromIndexedDBToOpfs()` method and the
+  `moveExistingIndexedDbToOpfs` parameter on `WasmDatabase.open`. They can be
+  used to move an existing database stored in IndexedDB to OPFS.
 
 ## 2.29.0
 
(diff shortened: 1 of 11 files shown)


Cunibon:
That should also work for my use case and is a lot cleaner than dealing with the initialised state like I tried.

Thank you very much! 
Also sorry I didn't get to finish the original MR, I got tangled up with other stuff.

Cunibon:
Just a thought, since the moving process is now independent of the database being initialized, could this just be a separate function you can call before you start the database?

This way the open function would not need the extra Boolean parameter, it's up to the user to do the migration call explicitly.

simolus3:
> Just a thought, since the moving process is now independent of the database being initialized, could this just be a separate function you can call before you start the database?

Interesting idea 👍 I suppose it should be on `WasmProbeResult`, which would let you check for existing databases to migrate them if they exist.

I still think having this parameter on `WasmDatabase.open` (which is essentially a shortcut for `probe()` + `WasmProbeResult.open`) makes sense, but that way you'd have full control over when to call this.

Issue #3727, opened 2025-12-19
Add a comment to ignore generated files for coverage


homeserve-lsaudon:
Add `// coverage:ignore-file` in generated files like `*.drift.dart` or `*.step.dart`

freezed do this [here](https://github.com/rrousselGit/freezed/blob/e51b801fa694e771cf14f5bf57d4902516730b84/packages/freezed/lib/builder.dart#L22)

simolus3:
There are a few prior issues for this (https://github.com/simolus3/drift/issues/2225, https://github.com/simolus3/drift/issues/2541, https://github.com/simolus3/drift/issues/2824).

My stance on this has been that we should not add this comment by default. After all, you're still kind of responsible for generated code in your project and so it might be reasonable to expect generated code to be covered by tests as well (or at least some people might prefer that).

As the other issues show, it's possible to to make drift emit this comment as an opt-in. If we do this by default, we should at least make an opt-out possible. But I'm also wondering if this is just a matter of documenting this option much better.

PR #3728, opened 2025-12-20
refactor: mutualize header for generated files

lsaudon:
#3727


drift_dev/lib/src/utils/dartfmt.dart
```diff
@@ -9,7 +9,10 @@ String formatDartCode(
 }) {
   var input = includeWidthComment
       ? '''
+// GENERATED BY drift_dev, DO NOT MODIFY.
 // dart format width=80
+// coverage:ignore-file
```
simolus3: As you can see in the diff for generated files in `drift/`, adding this here doesn't really work. The reason is that `includeWidthComment` is `false` in the default build setup (because we rely on a `source_gen` post-process builder to add the comment).

I like the approach of adding this in a central place, but maybe a better way is to extract the prefix comment into a constant somewhere and reference that in the existing places.

lsaudon:
I can mutalize and delete the ignored coverage comment to comply with your response in the issue.

simolus3 (approved):
Thanks!

PR #3739, opened 2026-01-09
Make sure to honor moveExistingIndexedDbToOpfs bool

Cunibon:
I noticed we define the moveExistingIndexedDbToOpfs bool but dont acutally use it in the code, so currently all drift users would automatically move to OPFS if their browser allows it.


simolus3 (commented):
Oops 😅 Thanks for fixing this!

simolus3 (approved):

PR #3740, opened 2026-01-14
feat: add option to serialize DateTime values as UTC

MichaelDark:


simolus3 (commented):
I'm happy with these changes, thank you for your contribution 👍 

Could you also add a test to `drift/test/database/data_class_test.dart`?

PR #3742, opened 2026-01-14
fix: match version with name

homeserve-lsaudon:
I think this is a mistake, but I'm not sure.

All versions have the same value as the constant name except for 50 and 51.


simolus3:
Thanks!

simolus3 (approved):

PR #3751, opened 2026-02-01
Expose 'watchWithAdditionalTables' to the developer

### LINKED ISSUE #3750: Expose 'watchWithAdditionalTables' to the developer
I have this use case:

```
  Stream<Map<String, RecipeData>> streamFiltered(Set<String> tagDataFilters) {
    final query = baseQuery;

    query.where(table.archived.equals(false));

    if (tagDataFilters.isNotEmpty) {
      final requiredTagIds = tagDataFilters;

      final tagSubquery = db.selectOnly(db.recipeTagTable)
        ..addColumns([db.recipeTagTable.recipeId])
        ..where(db.recipeTagTable.tagId.isIn(requiredTagIds))
        ..groupBy(
          [db.recipeTagTable.recipeId],
          having: db.recipeTagTable.tagId
              .count(distinct: true)
              .equals(requiredTagIds.length),
        );
      query.where(table.id.isInQuery(tagSubquery));
    }

    return query.watch().map(mapResult);
  }
```

This query is reactive for for the table involved in my baseQuery, but not the one in my subquery.

If we expose the currently internal 'watchWithAdditionalTables' function (in select_with_join.dart) I could specify my other table and have my result reactive to all table changes.

It would simply empower the user.

Cunibon:
- Expose watchWithAdditionalTables
- Delete extension

Closes https://github.com/simolus3/drift/issues/3750

### DIFF
--- a/drift/lib/drift.dart
+++ b/drift/lib/drift.dart
@@ -19,7 +19,6 @@ export 'src/runtime/executor/executor.dart';
 export 'src/runtime/executor/interceptor.dart';
 export 'src/runtime/query_builder/query_builder.dart'
     hide
-        JoinedSelectStatementAdditionalTables,
         CaseWhenExpressionWithBase,
         BaseCaseWhenExpression,
         SelectWithoutTables,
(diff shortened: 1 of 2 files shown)


simolus3:
As discussed in the linked issue, drift should already watch all referenced tables. There is probably a bug in here somewhere, but should be fixed instead of exposing this functionality as a workaround API.

PR #3769, opened 2026-03-11
fix(sqlparser): colon in line comment no longer breaks drift file parsing

### LINKED ISSUE #3763: Imported table definition not found in `.drift`-file
I migrated today from SQLite 2 to 3, and now the builder is showing me an unknown table warning.

```bash
search_queries.drift:                   
  `definitions_json_view` could not be found in any import.                                                        
      ╷                                                                                                                         
  642 │             FROM definitions_json_view   
```

However, in my other .drift file, that view is defined, and it is included as an `import ...` as well as in the DB's `include: {...}` section (that warning was also not shown in v2)

```sql
CREATE VIEW IF NOT EXISTS definitions_json_view AS
    SELECT
        TB3XDT.term_bank_id,
...
```
The build finishes, but the undefined view causes a huge number of error messages.
Did something change, and I overlooked it in the docs? Any help would be appreciated.

ersanKolay:
## Summary

Fixes #3763.

A colon inside a line comment (e.g. `-- VIEW : For Definitions`) was incorrectly matched by the `_statementMeta` regex as a named statement header. This caused the scanner to truncate the comment and parse the remaining text as SQL tokens, which broke all subsequent table/view definitions in the file.

The `_statementMeta` regex now requires the colon to be the last non-whitespace character on the line (`\s*$`), so regular comments containing colons are no longer mistaken for named statements.

## What changed

**`sqlparser/lib/src/reader/tokenizer/scanner.dart`** - one line: added `\s*$` to the regex

Before: `^\s*\w+\s*(?:\(.*\)\s*)?:`
After: `^\s*\w+\s*(?:\(.*\)\s*)?:\s*$`

## Test plan

- Added regression test with a comment containing a colon before a `CREATE VIEW`
- All 726 existing sqlparser tests pass
- `dart format` and `dart analyze` clean

### DIFF
--- a/sqlparser/lib/src/reader/tokenizer/scanner.dart
+++ b/sqlparser/lib/src/reader/tokenizer/scanner.dart
@@ -536,6 +536,8 @@ class Scanner {
 
   static final _importComment = RegExp(r'^\s*import.*;', caseSensitive: false);
   // match `foo:` or `myQuery (:variable AS TEXT):`
+  // The colon must be at the end of the line to avoid matching regular
+  // comments that happen to contain colons (e.g. `-- VIEW : description`).
   static final _statementMeta =
-      RegExp(r'^\s*\w+\s*(?:\(.*\)\s*)?:', caseSensitive: false);
+      RegExp(r'^\s*\w+\s*(?:\(.*\)\s*)?:\s*$', caseSensitive: false);
 }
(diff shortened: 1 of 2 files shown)


simolus3 (approved):
Thank you!

PR #3783, opened 2026-04-11
fix(streams): emit DatabaseDisconnectedException before closing watch() streams

### LINKED ISSUE #3782: watch() stream closes silently when database is closed — StreamBuilder freezes on stale data
### Describe the bug

When the drift database is closed (e.g. app backgrounded, connection reset), `QueryStream.close()` in `stream_queries.dart` calls `listener.controller.close()` on every active `watch()` stream. This sends a `done` event downstream.

Flutter's `StreamBuilder` handles `done` by stopping rebuilds and showing the last snapshot — silently. The widget appears functional but is frozen on stale data. There is no error, no indication the stream ended, and no way to distinguish "database closed" from "no more updates coming".

### Steps to reproduce

1. Open a database and call `watch()` on any query, observed via `StreamBuilder`
2. Close the database while the widget is still alive (e.g. background the app, or call `db.close()` explicitly)
3. Re-open the database and write new rows
4. Observe: `StreamBuilder` never rebuilds — it is frozen on the last snapshot with no error state

### Expected behavior

`watch()` should emit an error (e.g. a typed `DatabaseClosedException`) before the stream closes, so callers can detect the shutdown and re-subscribe after reconnection. `StreamBuilder.snapshot.hasError` would then be true and the UI can react.

### Proposed fix

In `QueryStream.close()` in `drift/lib/src/runtime/executor/stream_queries.dart` (around line 363), emit a typed error before closing each controller:

```dart
Future<void> close() async {
  _isClosed = true;

  final listenersDone = Future.wait([
    for (final listener in _listeners)
      listener.controller
        ..addError(const DatabaseDisconnectedException())
        ..close(),
  ]);
  _listeners.clear();
  await listenersDone;
}
```

Where `DatabaseDisconnectedException` is a new typed exception in the drift API, allowing callers to specifically handle database shutdown vs. query errors.

### Why this matters

For long-lived navigation sessions (e.g. a turn-by-turn nav app watching a route table), a silent stream close means the UI silently shows a stale route after the database reconnects. An observable error lets the app re-subscribe and recover.

### System info

- drift version: latest develop
- Platform: Flutter (Android/iOS/Linux)

*AI-assisted — authored with Claude, reviewed by Komada.*

aki1770-del:
Fixes #3782

## Problem

When a database is closed while `watch()` streams have active listeners, `QueryStream.close()` called `controller.close()` directly. This sent a `done` event with no error, making it impossible for `StreamBuilder` (or any other consumer) to distinguish a database shutdown from a normal stream completion. The widget silently froze on stale data.

## Fix

Two changes:

1. **`drift/lib/src/runtime/exceptions.dart`** — add `DatabaseDisconnectedException`, a new public exception class that callers can check against.
2. **`drift/lib/src/runtime/executor/stream_queries.dart`** — in `QueryStream.close()`, call `controller.addError(const DatabaseDisconnectedException())` on each `MultiStreamController` before calling `controller.close()`.

Callers can now detect shutdown via:

```dart
if (snapshot.hasError && snapshot.error is DatabaseDisconnectedException) {
  // database was closed; re-subscribe after reopening
}
```

The existing `done`-event behavior is preserved — the stream still closes. Only the distinguishable error is added before it.

*AI-assisted — authored with Claude, reviewed by Komada.*

### DIFF
--- a/drift/lib/src/runtime/exceptions.dart
+++ b/drift/lib/src/runtime/exceptions.dart
@@ -44,6 +44,24 @@ class DriftWrappedException implements Exception {
   }
 }
 
+/// Exception thrown on active [watch()] streams when the database is closed.
+///
+/// When a database is closed while a [watch()] stream has active listeners,
+/// drift emits this exception before sending the done event. This lets callers
+/// distinguish a database shutdown from a normal stream completion and
+/// re-subscribe after the database is reopened.
+///
+/// In Flutter, check for this in [StreamBuilder] via
+/// `snapshot.hasError && snapshot.error is DatabaseDisconnectedException`.
+class DatabaseDisconnectedException implements Exception {
+  /// Constant constructor.
+  const DatabaseDisconnectedException();
+
+  @override
+  String toString() => 'DatabaseDisconnectedException: '
+      'The database was closed while this stream was still active.';
+}
+
 /// Exception thrown by drift when rolling back a transaction fails.
 ///
 /// When using a `transaction` block, transactions are automatically rolled back
--- a/drift/lib/src/runtime/executor/stream_queries.dart
+++ b/drift/lib/src/runtime/executor/stream_queries.dart
@@ -363,8 +363,17 @@ class QueryStream<Rows extends Object> {
   Future<void> close() async {
     _isClosed = true;
 
-    final listenersDone = Future.wait(
-        [for (final listener in _listeners) listener.controller.close()]);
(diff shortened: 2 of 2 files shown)


simolus3:
Thanks for the contribution! But I don't think we should add this to drift, see https://github.com/simolus3/drift/issues/3782#issuecomment-4229310028. If throwing an exception is more convenient for you, you could always use a package like rxdart to inject one just before the stream emits a done event.

PR #3787, opened 2026-04-26
Add experimental DuckDB dialect to drift

Hu-Wentao:
## Problem
Currently, `drift_duckdb` still masquerades as `SqlDialect.sqlite`, causing the Drift DSL to reuse the type semantics of SQLite under DuckDB. The most straightforward issue is that both `DriftSqlType.int` and `DriftSqlType.bigInt` use `INTEGER`, while for DuckDB, 64 - bit integers should be `BIGINT`. This can lead to errors or sub - optimal results in the schema and cast SQL.

## Scope of this PR
- Introduce `SqlDialect.duckdb` in drift core.
- Map built-in types to DuckDB base names: `BOOLEAN`, `TEXT`, `BIGINT`, `DOUBLE`, `BLOB`.
- Keep DuckDB parameter placeholders on `?`, and prevent sqlite-style indexed placeholders such as `?1`.
- Reuse generic `CAST` and `ON CONFLICT` SQL generation where verified to work for DuckDB.
- Restrict DuckDB insert modes to an explicit minimal supported set instead of silently reusing sqlite behavior.
- Add focused SQL generation tests for DuckDB schema types, variables, casts, and a minimal insert/select/returning flow.

## What is intentionally not covered yet
- Full executor-layer support in `drift_duckdb`.
- Auto-increment / identity / sequence semantics.
- DuckDB-specific advanced types such as `UUID`, `STRUCT`, `LIST`, `HUGEINT`, etc.
- Deeper migration behavior differences.
- Full validation of all Drift upsert and conflict edge cases on DuckDB.

## Test coverage
- Schema generation emits `BIGINT` for `int` and `bigInt`.
- Schema generation emits `BOOLEAN` for `bool`.
- Schema generation emits `DOUBLE` for `double`.
- DuckDB variables use `?` and do not emit sqlite-style indexed placeholders.
- `cast<int>()` and `cast<BigInt>()` generate `CAST(... AS BIGINT)`.
- Minimal insert, select, and `RETURNING *` SQL generation paths are covered.

## ref
- [INSERT / ON CONFLICT / RETURNING](https://duckdb.org/docs/stable/sql/statements/insert)
- [Prepared Statements](https://duckdb.org/docs/stable/sql/query_syntax/prepared_statements)

### DIFF
--- a/drift/lib/src/runtime/query_builder/statements/insert.dart
+++ b/drift/lib/src/runtime/query_builder/statements/insert.dart
@@ -511,11 +511,21 @@ enum InsertMode implements Component {
       throw ArgumentError('$this not supported on postgres');
     }
 
-    ctx.buffer.write(
-      _insertKeywords[ctx.dialect == SqlDialect.postgres
-          ? InsertMode.insert
-          : this],
-    );
+    if (ctx.dialect == SqlDialect.duckdb &&
+        this != InsertMode.insert &&
+        this != InsertMode.replace &&
+        this != InsertMode.insertOrReplace &&
+        this != InsertMode.insertOrIgnore) {
+      throw ArgumentError('$this not supported on duckdb');
+    }
+
+    final effectiveMode = ctx.dialect == SqlDialect.postgres
+        ? InsertMode.insert
+        : ctx.dialect == SqlDialect.duckdb && this == InsertMode.replace
+        ? InsertMode.insertOrReplace
+        : this;
+
+    ctx.buffer.write(_insertKeywords[effectiveMode]);
   }
 }
 
--- a/drift/lib/src/runtime/query_builder/statements/insert.dart
+++ b/drift/lib/src/runtime/query_builder/statements/insert.dart
@@ -519,11 +519,12 @@ enum InsertMode implements Component {
       throw ArgumentError('$this not supported on duckdb');
     }
 
-    final effectiveMode = ctx.dialect == SqlDialect.postgres
-        ? InsertMode.insert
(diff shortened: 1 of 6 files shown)


drift/lib/src/runtime/query_builder/statements/insert.dart
```diff
@@ -511,11 +511,21 @@ enum InsertMode implements Component {
...
+        this != InsertMode.insertOrReplace &&
+        this != InsertMode.insertOrIgnore) {
+      throw ArgumentError('$this not supported on duckdb');
+    }
+
+    final effectiveMode = ctx.dialect == SqlDialect.postgres
+        ? InsertMode.insert
+        : ctx.dialect == SqlDialect.duckdb && this == InsertMode.replace
+        ? InsertMode.insertOrReplace
+        : this;
```
simolus3: A switch expression might make this a bit more readable, e.g.

```dart
final effectiveMode = switch (ctx.dialect) {
  SqlDialect.postgres => InsertMode.insert,
  SqlDialect.duckdb when this == InsertMode.replace => InsertMode.insertOrReplace,
  _ => this,
};
```

simolus3 (commented):
Thanks for the contribution!

simolus3 (approved):
Thanks!

Hu-Wentao:
Let me add one more thing. I recommend using the skill I developed to help you achieve automatic release. @simolus3 

https://github.com/Hu-Wentao/wyatt_skills/tree/main/skills/publish-flutter-package

PR #3803, opened 2026-05-21
sqlparser: added alter table add/drop constraint 

### LINKED ISSUE #3801: sqlparser: ADD/DROP CONSTRAINT
SQLite 3.53 enhanced ALTER TABLE with ADD/DROP CONSTRAINT.
The sqlparser package doesn't support this new syntax at the moment.

You recently added the SET/DROP NOT NULL syntax. If you're not already working on this, I'd like to try adding ADD/DROP CONSTRAINT to the sqlparser package.

Currently this seems to only be able to add CHECK constraints and DROP CHECK/NOT NULL constraints
ADD CONSTRAINT also has an optional conflict-clause, that doesn't exist in the standard column-constraint CHECK syntax
https://sqlite.org/lang_altertable.html
https://www.sqlite.org/syntax/column-constraint.html

ViveNoctem:
Fixes #3801

Added support for the new SQLITE 3.53 `ADD CONSTRAINT` and `DROP CONSTRAINT` syntax.

`ADD CONSTRAINT` can add named/unnamed `CHECK` constraints to the table constraints.
`DROP CONSTRAINT` can drop named `CHECK` or `NOT NULL` constraints.

---

In the sqlite documentation `ADD CHECK` has an optional conflict clause.
The sqlite repl doesn't throw an error adding the conflict clause but it seems to be ignored and would always ABORT on conflict.
I ignored the conflict clause for the time being, but can add it if you want.

I don't really like `parseOnlyCheckConstraints` in `Parser`. Wanted to reuse the existing code to parse table constraints, but 
add constraint only supports check constraint.

In `SchemaBuffer` the code to search for a constraint in the table and column constraints is kinda duplicated.
Didn't had a good idea to merge it because `_TableConstraint` and `_ColumnConstraint` are completely different Objects.

Should the `ReferenceResolver` resolve the columns in the check constraint expression?
It doesn't at the moment. Seems the table is not added to the `StatementScope`.

### DIFF
--- a/sqlparser/lib/src/reader/parser.dart
+++ b/sqlparser/lib/src/reader/parser.dart
@@ -2381,14 +2381,30 @@ extension Parser on ParserState {
 
     if (_matchOne(TokenType.add)) {
       final first = _previous;
+
+      final constraint = tableConstraintOrNull(parseOnlyCheckConstraints: true);
+      if (constraint != null) {
+        return AddConstraint(checkTable: constraint as CheckTable)
+          ..setSpan(first, _previous);
+      }
+
       _matchOne(TokenType.column);
       return AddColumn(_columnDefinition())..setSpan(first, _previous);
     }
 
     if (_matchOne(TokenType.drop)) {
       final first = _previous;
+
+      final constraintName = _constraintNameOrNull();
+      if (constraintName != null) {
+        return DropConstraint(name: constraintName.identifier)
+          ..setSpan(first, _previous);
+      }
+
       _matchOne(TokenType.column);
-      final name = _consumeIdentifier('Expected column to drop');
+      final name = _consumeIdentifier(
+        'Expected column to drop or CONSTRAINT keyword',
+      );
       return DropColumn(Reference.fromTokens(columnName: name))
         ..setSpan(first, _previous);
     }
@@ -2630,13 +2646,21 @@ extension Parser on ParserState {
     _error('Expected a constraint (primary key, nullability, etc.)');
   }
 
-  TableConstraint? tableConstraintOrNull({bool requireConstraint = false}) {
+  TableConstraint? tableConstraintOrNull({
+    bool requireConstraint = false,
+    bool parseOnlyCheckConstraints = false,
+  }) {
     final first = _peek;
     final nameToken = _constraintNameOrNull();
     final name = nameToken?.identifier;
 
     TableConstraint? result;
-    if (_match([TokenType.unique, TokenType.primary])) {
+    if (_matchOne(TokenType.check)) {
+      final expr = _expressionInParentheses();
+      result = CheckTable(name, expr);
+    } else if (parseOnlyCheckConstraints) {
+      return null;
+    } else if (_match([TokenType.unique, TokenType.primary])) {
       final isPrimaryKey = _previous.type == TokenType.primary;
 
       if (isPrimaryKey) {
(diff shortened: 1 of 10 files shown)


sqlparser/lib/src/reader/parser.dart
```diff
@@ -2381,12 +2381,25 @@ extension Parser on ParserState {
 
     if (_matchOne(TokenType.add)) {
       final first = _previous;
+
+      final constraint = tableConstraintOrNull(parseOnlyCheckConstraints: true);
+      if (constraint != null && constraint is CheckTable) {
```
simolus3: Can we remove the `is` check and just do an `as CheckTable` when it's non-null? If `parseOnlyCheckConstraints` is enabled we shouldn't return another type of constraint (and crashing on programming errors is what we should do).

sqlparser/lib/src/reader/parser.dart
```diff
@@ -2381,12 +2381,25 @@ extension Parser on ParserState {
...
       final first = _previous;
+
+      final constraintName = _constraintNameOrNull();
+      if (constraintName != null) {
+        return DropConstraint(name: constraintName.identifier)
+          ..setSpan(first, _previous);
+      }
+
       _matchOne(TokenType.column);
       final name = _consumeIdentifier('Expected column to drop');
```
simolus3:
```suggestion
      final name = _consumeIdentifier('Expected column to drop or CONSTRAINT keyword');
```

sqlparser/test/utils/schema_buffer_test.dart
```diff
@@ -226,5 +226,168 @@ CREATE TABLE tbl (
...
+              ),
+            ],
+          );
+        });
+      });
+    });
+
+    group("drop constraint", () {
+      test("check", () {
+        process(
```
simolus3, line 319: Could you also add a test to drop a table-level constraint (e.g. the same with `CREATE TABLE original (a INTEGER, CONSTRAINT check_constraint CHECK (a > 0));`)?

simolus3 (commented):
This looks great, thanks for your contribution! Could you also add a changelog entry please (I'd usually add a `## 0.44.5 (unreleased)` heading).

simolus3:
>  I ignored the conflict clause for the time being, but can add it if you want.

IMO this is fine for now.

> I don't really like `parseOnlyCheckConstraints` in `Parser`. Wanted to reuse the existing code to parse table constraints, but

It's not the first workaround of this kind in the parser :D

> Didn't had a good idea to merge it because `_TableConstraint` and `_ColumnConstraint` are completely different Objects.

We could introduce a shared superclass / interface, but IMO the duplication is not bad enough to require that yet.

> Should the `ReferenceResolver` resolve the columns in the check constraint expression?

Good point, `ColumnResolver` should handle `visitAlterTableStatement` and call `_resolveTableReference` there. At the moment we don't analyze any schema statements in drift itself, so it's likely a lot more in there is broken. We don't have to fix that in this PR (unless you want to).

ViveNoctem:
Made the requested changes.

> Good point, `ColumnResolver` should handle `visitAlterTableStatement` and call `_resolveTableReference` there.

Will try to fix this in a separate PR, if *"it's likely a lot more in there is broken"*.

simolus3 (approved):

simolus3:
Thank you for your contribution 👍

PR #3805, opened 2026-05-27
Fix: TableInfo equality is wrong if used across multiple databases.

eliaxelang007:
In the old equality code:

```dart
  @override
  bool operator ==(Object other) {
    // tables are singleton instances except for aliases
    if (other is TableInfo) {
      return other.runtimeType == runtimeType &&
          other.aliasedName == aliasedName;
    }
    return false;
  }

  @override
  int get hashCode => Object.hash(aliasedName, actualTableName);
```

If you:
* Closed your old database
* Still held a reference to a TableInfo attached to that closed database 
        * (In my case, that reference was held by a Riverpod provider family)
* And then did equality between that old TableInfo and a new TableInfo, 
* Even if both TableInfos are attached to completely different databases, if both TableInfos had the same table name, the equality would wrongly say they're equal.

This fixes that by checking if the Table infos have the same [attachedDatabase], too!

### DIFF
--- a/drift/CHANGELOG.md
+++ b/drift/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## 2.34.0 (unreleased)
 
 - Use `BEGIN IMMEDIATE` to start transactions.
+- Make `TableInfo`'s `==` and `hashCode` take into account the database its attached to.
 
 ## 2.33.0
 
--- a/drift/lib/src/runtime/query_builder/schema/table_info.dart
+++ b/drift/lib/src/runtime/query_builder/schema/table_info.dart
@@ -99,16 +99,16 @@ mixin TableInfo<TableDsl extends Table, D> on Table
 
   @override
   bool operator ==(Object other) {
-    // tables are singleton instances except for aliases
+    // Tables are singletons within the databases they're attached to (except for aliases).
     if (other is TableInfo) {
       return other.runtimeType == runtimeType &&
-          other.aliasedName == aliasedName;
+          other.aliasedName == aliasedName && other.attachedDatabase == attachedDatabase;
     }
     return false;
   }
 
   @override
-  int get hashCode => Object.hash(aliasedName, actualTableName);
+  int get hashCode => Object.hash(aliasedName, actualTableName, attachedDatabase);
 }
 
 /// Additional interface for tables in a drift file that have been created with
--- a/drift/lib/src/runtime/query_builder/schema/table_info.dart
+++ b/drift/lib/src/runtime/query_builder/schema/table_info.dart
@@ -102,13 +102,15 @@ mixin TableInfo<TableDsl extends Table, D> on Table
     // Tables are singletons within the databases they're attached to (except for aliases).
     if (other is TableInfo) {
       return other.runtimeType == runtimeType &&
-          other.aliasedName == aliasedName && other.attachedDatabase == attachedDatabase;
+          other.aliasedName == aliasedName &&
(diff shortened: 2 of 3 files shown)


simolus3 (commented):
Thanks for the report and fix! Out of interest, when does this actually cause issues for you? Did this cause you to to use statement methods on a "closed" table?

eliaxelang007:
Hi!

My use case is a bit weird, but hopefully the examples will make it make sense.

This is a simplified version of what my tables look like:

```dart
abstract interface class GroupTable extends Table {
  Column<int> get id;
  Column<String> get name;
}

abstract interface class Group {
  int get id;
  String get name;
}

class GroupCompanion<T extends Group> extends UpdateCompanion<T> {
  /* This is a modified version of companion from the generated code that works with any Group and GroupTable. */
}
```

```dart
@DataClassName("GroupOfThing", implementing: [Group])
class GroupOfThings extends GroupTable {
  @override
  late final id = integer().autoIncrement()();

  @override
  late final name = text()();
}
```

Given that, my Riverpod providers look like this:

```dart
  static final groupOfThingsTableProvider = Provider<$GroupOfThingsTable>((ref) {
    final database /* This is a Drift Database! */ = ref.watch(projectProvider);

    return database.groupOfThings;
  });

  static final groupsProvider =
    StreamProvider.family<List<Group>, TableInfo<GroupTable, Group>>((
      ref,
      table,
    ) {
      return table.select().watch();
    });
```

The provider families take in TableInfos and not the concrete types because I have multiple tables with the same column names and types that need the same kind of UI to interact with.

The provider family acts like a cache map where the TableInfo is the key and the result of the query is the value.

Now, in my app, I use SQLite DB files like project files, so the tables inside them all have the same names.
When I close one "project" and then open another, because the TableInfo equality is "wrong", the provider family's cache map gives out the old cached query results.

So yes, this did cause a few operations on some tables from closed databases!

Hope that clarified it!

P. S.

If anybody comes across this needing a temporary fix, here's mine:

```dart
class EquatableTableInfo<TableDsl extends Table, TableRow> {
  final TableInfo<TableDsl, TableRow> wrapped;

  EquatableTableInfo(this.wrapped);

  @override
  bool operator ==(Object other) {
    if (identical(this, other)) return true;
    if (other is! EquatableTableInfo) return false;

    final otherWrapped = other.wrapped;

    // SAFETY: In TableInfo equality,
    // Drift assumes that you'll only have one database for the entire lifetime of the program.
    // In this app, multiple different databases can exist, but only one can exist at a time.
    // That's why we're also doing equality on the attached databases themselves (even if attachedDatabase is only supposed to be used internally).
    return wrapped == otherWrapped &&
        wrapped.attachedDatabase == // ignore: invalid_use_of_internal_member
            otherWrapped
                .attachedDatabase; // ignore: invalid_use_of_internal_member
  }

  @override
  // Excuse the use of the internal [attachedDatabase]! See the equality operator override for more details.
  int get hashCode => Object.hash(wrapped, wrapped.attachedDatabase); // ignore: invalid_use_of_internal_member
}
```

simolus3 (commented):
Thanks for the context, this makes sense to me 👍 

Could you add a changelog entry for this and a test? For tests, you could open two instances of a drift database and assert that `db1.categories != db2.categories` for example.

eliaxelang007:
Alright! I've added a commit. Do the tests and changelog entry look good to you?

simolus3 (approved):

PR #3807, opened 2026-05-31
added RETURNING only for drift3

### LINKED ISSUE #3806: drift: DML RETURNING partial result
Currently drift only supports `RETURNING *`. I would love to be able to select which columns are being returned when using insert,delete, or update.
https://sqlite.org/lang_returning.html

Some kind of equivalent to selectOnly for insert, delete, and update statements.
I don't know if it's worth adding a separate `insertOnly` with `addColumns` and `go()` to get a `TypedResult`, just like `selectOnly`.
Or adding a method like `goAndReturnOnly(Iterable<Expression> expressions)` where you can directly pass the columns and get the `TypedResult`.

If this is something worth adding to drift, I would like to try.

ViveNoctem:
closes #3806.

Added the possibility to pass `List<Expression>` to `ReturningClause`, to be able to return partial rows in insert, delete, and update statements.
Added `interpretExpressionResults` to get the result as a `List<DriftRow>` instead of the complete `Row`.

Moved some methods from `BaseSelectStatement` to `ResultSetStructure`, to be able to reuse them for the returning clause.

Adding it to the current Drift version would be possible as well, but the new `ResultSetStructure` makes it a bit cleaner than adding a mixin or something to try and reuse the implementation from `SelectStatement`.

### DIFF
--- a/drift/lib/src/drift3_preview/src/query_builder/clauses/returning.dart
+++ b/drift/lib/src/drift3_preview/src/query_builder/clauses/returning.dart
@@ -7,23 +7,10 @@ import '../schema/table.dart';
 
 /// A `RETURNING *` clause that can appear as part of an `INSERT`, `UPDATE` or
 /// `DELETE` statement.
-final class ReturningClause<
-  Row extends Object,
-  RS extends GeneratedTable<Row, RS>
->
-    implements SqlComponent {
+final class ReturningClause implements SqlComponent {
   /// The generated [ResultSetStructure] representing columns for this
   /// `RETURNING` clause.
   final ResultSetStructure structure = ResultSetStructure();
-
-  final ResultSet<Row, RS> _resultSet;
-
-  /// Creates a `RETURNING` clause for the given [ResultSet].
-  ReturningClause(this._resultSet) {
-    // Note: We currently only generate `RETURNING *` clauses returning columns
-    // for a single table.
-    structure.addSelectStarFromSingleTable(_resultSet);
-  }
 
   @override
   void compileWith(StatementCompiler compiler) {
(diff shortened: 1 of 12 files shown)


drift/lib/src/drift3_preview/src/query_builder/clauses/returning.dart
```diff
@@ -17,12 +18,18 @@ final class ReturningClause<
...
+  final List<Expression>? _expressions;
 
   /// Creates a `RETURNING` clause for the given [ResultSet].
-  ReturningClause(this._resultSet) {
-    // Note: We currently only generate `RETURNING *` clauses returning columns
-    // for a single table.
-    structure.addSelectStarFromSingleTable(_resultSet);
+  ///
+  /// If [_expressions] is provided [_resultSet] will not be added to the [ResultSetStructure]
+  ReturningClause(this._resultSet, this._expressions) {
```
simolus3: It's kind of odd that one has to pass a `_resultSet` when only the referenced expressions are added.

Having a reference to `_resultSet` in a field here was kind of lazy, we could:

1. Make the constructor take no arguments, and let the caller be responsible for adding the result set to `structure`.
2. Remove the `Row` and `RS` types from `ReturningClause`, instead adding them to the `interpretResults` method which would also gain a `ResultSet` parameter.

Basically, there shouldn't be anything in the `ReturningClause` class itself that references a result set anymore.

simolus3 (commented):
Thanks for your contribution! The API changes look good to me, I have one note about the implementation.

drift/lib/src/drift3_preview/src/query_builder/results.dart
```diff
@@ -60,20 +65,52 @@ final class ResultSetStructure {
...
+  /// The column will be evaluated for each row of the expression. To read the
+  /// result of the expression in a row, use [DriftRow.read].
+  void addColumns(Iterable<Expression> columns) {
+    for (final column in columns) {
+      expressions[column] ??= _nextPosition;
+    }
+  }
+
+  /// Adds all columns from [ResultSet] to this [ResultSetStructure].
+  void addResultSet(ResultSet resultSet) {
```
ViveNoctem, line 87: @simolus3 i removed `addSelectStarFromSingleTable` because  i just realized it is exactly the same as `addResultSet`, which i copied from `SelectStatement`, except for the `assert(expressions.isEmpty);`.

drift/lib/src/drift3_preview/src/query_builder/clauses/returning.dart
ViveNoctem: Implemented the changes.

simolus3 (approved):
Nice, thanks for your contribution 👍

PR #3808, opened 2026-06-02
doc(sqlite): wrong version in doc 3.52 -> 3.53

homeserve-lsaudon:


sqlparser/lib/src/engine/options.dart
```diff
@@ -117,9 +117,12 @@ class SqliteVersion implements Comparable<SqliteVersion> {
   /// Newer features in `sqlite3` may not be recognized by this library.
   static const SqliteVersion current = v3_53;
 
+  /// Version `3.53.0`
+  static const SqliteVersion v3_53 = SqliteVersion.v3(53);
+
   /// Version `3.52.0` added the `json_array_insert` and `jsonb_array_insert`
   /// functions and enhanced `ALTER TABLE`.
-  static const SqliteVersion v3_53 = SqliteVersion.v3(53);
+  static const SqliteVersion v3_52 = SqliteVersion.v3(52);
```
simolus3: While version 3.52 is technically the one adding these features, that release has been retracted and I don't think we should reference it anymore. So I think only having 3.53 as an option is correct:

```suggestion
  /// Version `3.53.0` added the `json_array_insert` and `jsonb_array_insert`
  /// functions and enhanced `ALTER TABLE`.
  static const SqliteVersion v3_53 = SqliteVersion.v3(53);
```

simolus3 (approved):

PR #3809, opened 2026-06-04
feat(drift_dev): add use_sql_table_name_for_accessors build option

shenlong-tanwen:
Names the table getters on the database class after the table's SQL name (from its `tableName` definition) instead of the Dart class name. With it on, a table with `tableName => 'todos'` is accessed as `database.todos` instead of `database.todoItems`.

For backward compatibility, a new build option `use_sql_table_name_for_accessors` defaulting to `false` is added


drift_dev/lib/src/analysis/results/table.dart
```diff
@@ -33,6 +33,10 @@ class DriftTable extends DriftElementWithResultSet {
   /// The default name to use for the [entityInfoName].
   String baseDartName;
 
+  /// The fixed [dbGetterName] to use, overriding the default derived from
+  /// [baseDartName].
+  String? fixedDbGetterName;
```
simolus3: This is more general than it needs to be. I think I would prefer an approach where we don't need to store this here, and perhaps instead refactor `dbGetterName` into a method taking an options instance. Then the implementation could be `String dbGetterName(DriftOptions options) => DriftSchemaElement.dbFieldName(options.useSqlTableNameForAccessors : id.name ? baseDartName)`.

From a quick look, it looks like this getter is only used in places where we're generating code, and the writer carries a reference to options. So I hope this isn't too painful to change, but let me know if I can help :)

simolus3 (approved):
I like this option, thanks for your contribution 👍 This also matches what we do for views, triggers and indexes, so tables are the only outlier here. I don't remember a good reason for that, so adding this option now and potentially even making it the default in the next major version sounds good to me.

drift_dev/lib/src/analysis/results/table.dart
shenlong-tanwen: I gave this a shot but even a simple grep for `dbGetterName` yields several hits and some of those are in relation and migrator. I'm not sure about the cleanest way to pass the options through to them

simolus3: You were right, patching the manager code was annoying :D I still think it's worth it though, I'll revisit this tomorrow.

shenlong-tanwen: Hey, thanks a lot for the changes. They look really good. The CI was failing because of a styling issue, and I took the liberty of fixing it.

PR #3810, opened 2026-06-09
Support analyzer 13

dbebawy:
## What this does

Migrates `drift_dev` (and widens `drift`'s constraint) to support **`analyzer` 13**.

`analyzer` 13.0.0 introduced breaking AST changes that `drift_dev` depended on:

- The `NamedExpression` AST node was **removed**, replaced by `NamedArgument`. Its name is now a `Token` (`.name.lexeme`) rather than a `Label` (`.name.label.name`), and its value is exposed as `argumentExpression` instead of `expression`.
- `ArgumentList.arguments` now returns `NodeList<Argument>` instead of `NodeList<Expression>`. `Expression` now `implements Argument`, so positional arguments need an explicit cast back to `Expression`.

### Changes

- `drift_dev/lib/src/analysis/resolver/dart/helper.dart`, `.../column.dart`, `.../cli/commands/migrate.dart`: migrate off `NamedExpression` → `NamedArgument` and add casts for the `ArgumentList.arguments` element-type change.
- Constraints:
  - `drift_dev`: `analyzer: '>=10.0.0 <13.0.0'` → `'>=13.0.0 <14.0.0'`
  - `drift`: `analyzer: '>=8.0.0 <13.0.0'` → `'>=8.0.0 <14.0.0'` (runtime package only references analyzer in a test)
- CHANGELOG entries for both packages.

Since the AST changes are breaking, supporting both analyzer 12 and 13 from one source tree isn't possible, so `drift_dev`'s lower bound moves to `13.0.0`.

## Why

Downstream packages now require analyzer 13 and can't co-exist with drift_dev's `<13.0.0` cap. For example `mockito` >= 5.6.5 requires `analyzer ^13.0.0`, so any project depending on both `drift_dev` and a recent `mockito` fails version solving.

## Verification

- `dart analyze` is clean across `drift_dev/lib`, `drift_dev/test`, `drift/lib`, `sqlparser/lib`.
- `dart test` for `drift_dev` passes, including all 316 tests under `test/analysis/` (which exercise the migrated Dart column/view resolver). The only failures are the 4 `test/cli/migrate_test.dart` cases that rely on resolving the legacy `moor`/`moor_flutter` packages — these **fail identically on `develop` with analyzer 12** in the same sandbox, i.e. they are pre-existing/environmental and unrelated to this change.

## Note on CI / workspace resolution

The full melos workspace pulls `flutter_test` (via `drift_sqflite`/`drift_flutter`). With an older bundled Flutter SDK, `flutter_test` pins `test_api 0.7.11`, which forces `package:test` ≤ 1.31.0 and transitively caps `analyzer < 13`. `package:test` **1.31.1** widened to `analyzer < 14` but requires `test_api 0.7.12`, so resolving analyzer 13 across the whole workspace needs a Flutter SDK whose `flutter_test` allows `test_api 0.7.12`. I verified `drift`/`drift_dev`/`sqlparser` against analyzer 13.1.0 by resolving those three (pure-Dart) packages on their own. Please confirm this lines up with the Flutter stable version CI targets before merging.

Happy to adjust the constraint range or split the changes if you'd prefer to keep wider analyzer support.


dbebawy:
Thanks for the review. Addressed in 7d5fea8.

**On the `as Expression` casts (could throw):** valid point for the positional-extraction sites. Since analyzer 13's `Argument` is `sealed` (`Expression | NamedArgument`) and `NamedArgument` is no longer an `Expression`, a stray named argument — only possible in invalid source, since drift's API takes positional arguments here — would have thrown a `CastError` during code generation, whereas under analyzer ≤12 it was a `NamedExpression` (an `Expression`) that degraded into a normal analysis error. I've replaced the casts with a `positionalArgumentExpression` helper that returns `null` for a named argument (no cast, can't throw). Each call site already has a "could not resolve" path for a null/non-literal value, so the behaviour is now a proper drift analysis error instead of a crash.

**On the repeated pattern:** the same helper centralizes all five sites, so there's now a single place handling the `Argument → Expression?` conversion.

One note: the flagged site in the `generatedAs` loop (the `else` branch of `if (expr is NamedArgument)`) couldn't actually throw — because `Argument` is sealed to exactly `Expression | NamedArgument`, the `else` is provably an `Expression`; the cast was only needed because flow analysis doesn't promote on a non-exhaustive negative `is`. It now goes through the helper too for consistency.

simolus3:
Thanks for the contribution. But as you note, analyzer version 13 is not compatible with the latest stable Flutter version. And because we're affected by breaking changes, `drift_dev` can't support both 12 and 13.

So we're forced to make a decision between supporting the latest analyzer and being able to support the latest Flutter version. Given that most drift users use it with Flutter, I don't want to upgrade to the latest analyzer version just yet. I assume you're on a non-stable Flutter branch or aren't using Flutter at all? Do you need any new mockito features or is this mostly about supporting up-to-date transitive dependencies?

dbebawy:
Thanks, that's a fair call and matches what I ran into — no objection to holding.

To answer your questions:

- We're actually on **Flutter stable** (3.44.1) as well, not a side branch. The bump came from Renovate keeping transitive dependencies current rather than any feature need — mockito 5.7.0's only real change is providing dummy values for JS-interop extension types, which we don't mock. So we'll just keep mockito pinned at 5.6.4 until drift moves; there's no urgency on our end.

- For context on the incompatibility: the conflict is in drift's *own* workspace — on stable, `flutter_test` pins `test_api 0.7.11`, which caps `package:test` at ≤1.31.0 and transitively `analyzer <13`. An app that only uses `flutter_test` (no direct `package:test`) can resolve `drift_dev` + `analyzer 13` on stable today. But I take your point that it's largely moot: requiring `analyzer ^13` would still strand anyone on an older Flutter, and the breaking AST changes rule out a single source tree spanning 12 and 13 — so that trade-off is the real one.

Happy to leave this open as a ready-to-go reference for whenever stable's bundled analyzer catches up — the migration and the named-argument hardening are done and tested — or close it if you'd rather track it another way. Your call, and thanks for the quick response.

simolus3:
I was not aware of the `package:test` limitation, I thought Flutter apps might have the same problem. Let's add a dependency override for `test_api: ^0.7.11` to the workspace root, that should allow analyzer 13.0 to resolve. If that passes the CI, I'm fine with the override since it's only used as a dev dependency.

dbebawy:
Done — added in 641e488. Root `dependency_overrides: test_api: ^0.7.11` lets the override pick `test_api` 0.7.12 (required by `package:test` 1.31.1, the first release allowing `analyzer <14`) while bypassing `flutter_test`'s exact `0.7.11` pin.

With it, the **full** workspace resolves on stable Flutter (3.44.1):

```
analyzer  13.0.0
test      1.31.1
test_api  0.7.12
```

Verified locally:
- `drift_dev` analysis suite — 316/316 pass.
- `drift_flutter` tests (the ones using `flutter_test`, incl. widget tests) — pass with `test_api` 0.7.12, so flutter_test tolerates the override.

Curious to see if CI agrees. Thanks for reconsidering!

drift/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 - Use `BEGIN IMMEDIATE` to start transactions.
 - Make `TableInfo`'s `==` and `hashCode` take into account the database its attached to.
+- Widen the `analyzer` constraint to allow version 13.
```
simolus3: Remove this changelog item, analyzer is a dev-dependency here so this doesn't affect users.

drift_dev/pubspec.yaml
```diff
@@ -40,7 +40,7 @@ dependencies:
   sqlparser: ^0.44.0
 
   # Dart analysis
-  analyzer: '>=10.0.0 <13.0.0'
+  analyzer: '>=13.0.0 <14.0.0'
```
simolus3:
```suggestion
  analyzer: ^13.0.0
```

drift_dev/lib/src/analysis/resolver/dart/helper.dart
```diff
@@ -214,13 +214,24 @@ int? readIntLiteral(Expression expression) {
...
 
+/// The [Expression] of a positional [argument] from an [ArgumentList].
+///
+/// Since analyzer 13, `ArgumentList.arguments` is a `NodeList<Argument>` where
+/// the sealed [Argument] is either an [Expression] (positional argument) or a
+/// [NamedArgument]. This returns `null` for a named argument instead of casting,
+/// which would throw. Callers expect a positional expression here, so a `null`
+/// result (only reachable for invalid source) is reported through their regular
+/// "could not resolve" error paths rather than crashing code generation.
+Expression? positionalArgumentExpression(Argument argument) =>
```
simolus3: I actually think removing this helper and casting on use-sites would be cleaner. We only do this for calls to methods we know, and these methods have no named parameters. So there's no need to be defensive here, crashing instead of silently ignoring an unexpected argument is probably better.

drift_dev/CHANGELOG.md
```diff
@@ -5,6 +5,11 @@
 - Add the `use_sql_table_name_for_accessors` build option, naming the table
   getters on the database class after the table's SQL name instead of the Dart
   class name.
+- Support analyzer version 13. Migrate off the removed `NamedExpression` AST
+  node (replaced by `NamedArgument`) and adapt to the `ArgumentList.arguments`
+  element type change, reporting an analysis error instead of crashing when a
+  named argument appears where a positional one is expected. This drops support
+  for analyzer versions below 13.0.0.
```
simolus3: There's no need to mention implementation details in changelogs:

```suggestion
- Require analyzer version 13.
```

simolus3 (commented):
I can probably help with some of the CI failures.

dbebawy:
Thanks for the review — all four addressed and force-pushed (now a clean 2-commit branch):

- **`positionalArgumentExpression` helper removed**, casting at the use-sites instead. Agreed: these are calls to known drift builder methods with no named parameters, so a hard cast that crashes on an unexpected argument is better than silently dropping it.
- `drift_dev` analyzer constraint → `^13.0.0`.
- `drift_dev` changelog → just `- Require analyzer version 13.`
- Removed the `drift` changelog entry (analyzer is a dev-dependency there).

Two CI jobs are still red, both for reasons outside this PR's code — flagging since they're your call:

1. **`test_flutter (3.41.9)`** — version-solving failure, not a test failure: that Flutter's `flutter_test` pins `meta 1.17.0`, but `analyzer >=10.0.2` needs `meta ^1.18.0`, so analyzer 13 can't resolve on it at all. Requiring analyzer 13 effectively drops support for Flutter versions that pin `meta < 1.18`. I didn't want to paper over it with another override, since real users on that Flutter would hit the same wall.
2. **`Documentation`** — `drift_website` depends on `jaspr_content_snippets`, which still caps `analyzer <13`, so the docs build can't resolve `drift_dev` from path until that package updates.

Happy to follow your lead on both (e.g. dropping the older Flutter from the matrix, and/or excluding `drift_dev` from the docs resolution) if you'd like me to push those changes too.

PR #3815, opened 2026-06-21
Added UPDATE FROM to drift3

### LINKED ISSUE #3690: Allow using FROM in UPDATE statements
Hi! SQLite support `UPDATE FROM` statements beginning with version 3.33.0. I've managed to implement this statement using Dart extensions and a little bit of hacks, but I'm afraid I don't have required SQL knowledge to prepare a good enough PR. You can find PoC implementation in the [following gist](https://gist.github.com/ky1vstar/bed19ea82aa4f7b106c87adb28c303e5). It would be nice to have it implemented properly in Drift. Thanks

ViveNoctem:
closes #3690 

Added the possibility to reference columns from other tables/joins in `UpateStatements`
Added `writeExpressions` in `UpdateStatements` to make it possible to set columns based on other tables.

If you have a better idea for `writeExpression` than a `Map<String, Expression` let me now.

### DIFF
--- a/drift/lib/src/drift3_preview/CHANGELOG.md
+++ b/drift/lib/src/drift3_preview/CHANGELOG.md
@@ -11,3 +11,5 @@
 - Drift manager is now a standalone package, `drift_manager`.
 - Add `TransactionOptions`, which can be used to request read-only transactions.
 - `Selectable.watch` now returns a `QueryStream`, which can be used to track which tables it watches.
+- DML statements now support `RETURNING` partial rows.   
+- `FROM` can now be added to `UpdateStatement`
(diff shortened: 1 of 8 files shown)


simolus3 (commented):
Thanks for your contribution. The idea looks good to me, some notes:

- Since this is now shared between update and select statements, I wonder if it could make sense to extract this into a `FromClause` class / `addFromClause` method on the compiler to re-use the implementation?
- Let's also add this to the changelog in the drift3 preview package.

ViveNoctem:
Thanks for the review.

Added the separate `FromClause` class and moved the `FromClauseElement` class to the same file.
Also added the returning only to the changelog

simolus3 (approved):
LGTM, thank you!

PR #3818, opened 2026-06-23
drift_dev: Added support for dart update placeholder in drift files

### LINKED ISSUE #3516: Drift files: Support Dart placeholders for updates
In drift files, companions can be referenced for insert statements using

```SQL
createUser: INSERT INTO users $newUser;
```

which will then make drift generate a `createUser(Insertable<User> newUser)` method to supply different columns.

This syntax does not currently work for updates, but something like `UPDATE users SET $changes` might be nice to have.

ViveNoctem:
closes #3516 

Similar to insert statements, it is now possible to reference insertables in update statements.

```sql
updateUser: UPDATE users SET $changes
```
will generate
```dart
updateUser(Insertable<User> changes);
```
Initially, I had the update placeholder as a separate placeholder type, but since the implementation was practically the same, I merged them.

Should `$writeUpdateInsertable` also be merged into `writeInsertable` with a `isInsert` argument?
Should `DartPlaceholder`  be refactored into a `sealed class`?

### DIFF
--- a/drift_dev/lib/src/analysis/resolver/queries/query_analyzer.dart
+++ b/drift_dev/lib/src/analysis/resolver/queries/query_analyzer.dart
@@ -1,4 +1,5 @@
 import 'package:analyzer/dart/ast/ast.dart' as dart;
+import 'package:collection/collection.dart';
 import 'package:drift/drift.dart' as drift;
 import 'package:recase/recase.dart';
 import 'package:sqlparser/sqlparser.dart' hide ResultColumn;
@@ -758,9 +759,11 @@ class QueryAnalyzer {
   ) {
     final name = placeholder.name;
 
-    final type = placeholder.when(
-      isExpression: (e) {
-        final foundType = context.typeOf(e);
+    final DartPlaceholderType type;
+
+    switch (placeholder) {
+      case DartExpressionPlaceholder():
+        final foundType = context.typeOf(placeholder);
         ColumnType? columnType;
         if (foundType.type != null) {
           columnType = typeMapping.sqlTypeToDrift(foundType.type);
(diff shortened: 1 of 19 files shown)


drift_dev/lib/src/analysis/resolver/queries/query_analyzer.dart
```diff
@@ -778,11 +779,19 @@ class QueryAnalyzer {
...
-        final insert = placeholder.parents.whereType<InsertStatement>().first;
-        final table = insert.table.resultSet;
+        final parentStatement = placeholder.parents.firstWhereOrNull(
+          (s) => s is InsertStatement || s is UpdateStatement,
+        );
+
+        final (table, isInsert) = switch (parentStatement) {
+          InsertStatement() => (parentStatement.table.resultSet, true),
+          UpdateStatement() => (parentStatement.table.resultSet, false),
+          _ => (null, true),
```
simolus3: nit: Can we explicitly match on `null` and add a `_ => throw AssertionError()` default block to make it clearer what kind of values we expect here?

sqlparser/lib/src/ast/drift/inline_dart.dart
```diff
@@ -73,6 +73,9 @@ class DartOrderByPlaceholder extends DartPlaceholder implements OrderByBase {
 }
 
 class DartInsertablePlaceholder extends DartPlaceholder
-    implements InsertSource {
+    implements InsertSource, SetComponent {
   DartInsertablePlaceholder({required String name}) : super._(name);
+
+  @override
+  List<Reference> get columns => const [];
```
simolus3, line 59: Maybe add a comment here explaining that the columns for `UPDATE ... SET $dart` are always generated dynamically at runtime.

simolus3 (commented):
Thanks for all your contributions, this looks great 👍  I have two minor comments, the only other thing missing is [this section in the documentation](https://github.com/simolus3/drift/blob/develop/docs/content/sql_api/drift_files.md?plain=1#L289) which should be updated to mention the new feature.

drift_dev/lib/src/analysis/resolver/queries/query_analyzer.dart
ViveNoctem: The table should never be null. It is required by `$writeInsertable` and `_writeDartPlaceholder` would generate `this.null` as the table argument.
Should i make the `table` in `InsertableDartPlaceholderType` not nullable and throw an `AssertionError` if it can't be determined?

sqlparser/lib/src/ast/drift/inline_dart.dart
ViveNoctem: added

ViveNoctem:
Just to be sure:
Have you read my questions about merging `$writeInsertable` and `$writeUpdateInsertable`, and converting `DartPlaceholder` into a sealed class?

simolus3:
Sorry, I missed those questions.

> Should `$writeUpdateInsertable` also be merged into `writeInsertable` with a `isInsert` argument?

I think it makes sense to have them as separate entrypoints, since the output format is kind of different. I think the implementations are also simple enough to not need a private helper method.

> Should `DartPlaceholder`  be refactored into a `sealed class`?

That `when` method sure is ugly, I agree that a `sealed class` would be much better 👍

drift_dev/lib/src/analysis/resolver/queries/query_analyzer.dart
simolus3: Looking at `QueryAnalyzer._extractPlaceholder`, null mostly seems to be a defense against statements like `INSERT INTO actually_a_view $placeholder`. Here the table would be null and we'd emit a warning as part of an earlier analysis step, I think warnings + generating broken code is better than throwing outright. So keeping the null case sounds fine?

I should have definitely added a comment for that though :D

ViveNoctem: I'm not sure which earlier analysis step you're referring to.
When using a view with a placeholder, I only get the broken generated code and no warnings.
Therefore, I added an `AnalysisError` if the table couldn't be determined.

Besides that `INSERT INTO actually_a_view` is an actual, but probably very niche, use case, if triggers with `INSTEAD OF` exist.

simolus3: > Besides that `INSERT INTO actually_a_view` is an actual, but probably very niche, use case, if triggers with `INSTEAD OF` exist.

Good point, but the existing logic can't handle that either. So the current changes look good to me.

simolus3 (approved):

PR #3820, opened 2026-06-26
build(docker): improve testing with docker

### LINKED ISSUE #3811: [DX] Running `docker_test.sh` does not work as expected
First of all thank you for this great library!

I wanted to contribute an improvement, because I ran into what I think is a bug, but I couldn't get the local setup to run as excepted.
My idea was that running `./tool/docker_test.sh` should work without much setup, but there are multiple issues I encountered trying to make the Dockerfile work.

Do you actively make use of this Dockerfile and if so, what am I missing?

jonasbadstuebner:
Utilizes docker-compose to spin up a local dev environment

Adds documentation on how to use this setup

I am not sure if the `melos.yaml` was still used, I made changes to it and wondered why they seem to have no effect - I think the `melos` section in the `pubspec.yaml` is sufficient for configuring melos.
I also changed the config of it, because melos wouldn't run properly on my end, when it tried to run tests made for Flutter with `dart test`.

Please let me know if I should change anything about this PR.

Closes #3811


jonasbadstuebner:
Having this work helped me create two issues with test cases already, so I'm highly interested in having this merged upstream.

jonasbadstuebner:
Just found out the wasm test is not running as expected. I'll look into it. But if you could already have a look at the general direction this would be going, I'd appreciate it :bow: 

EDIT: I just realized this one is an integration test, which don't work.
I think I'd have to dive a little deeper in how the tests are set up to figure out how to make them work inside docker.

pankaj-basnet:
Hi everyone! I’m a self-taught developer and have been really enjoying using **drift** in the  [wger-project/flutter](https://github.com/wger-project/flutter/pull/1183) recently , and I also just discovered  Melos !

I would love to contribute directly to the `drift` repository, if given an opportunity . Given my few months of experience in open source projects, would you recommend any specific areas or 'good first issues' where I could dive deeper and be helpful? or any other open source project, where help is needed?

 Thanks for the awesome work on this project, @simolus3 .it's a fantastic library!"

CONTRIBUTING.md
```diff
@@ -73,12 +73,56 @@ used to construct common queries.
 
 ### Debugging the builder
 
-To debug the builder, run `pub run build_runner generate-build-script` in the `drift`
+To debug the builder, run `dart run build_runner build` in the `drift`
```
simolus3, line 76: This paragraph is now longer up-to-date. Can we link to [this section](https://pub.dev/packages/build_runner#debugging-builds) instead?

CONTRIBUTING.md
```diff
@@ -73,12 +73,56 @@ used to construct common queries.
...
+### Running tests inside docker
+For an interactive unit-testing environment, spin up the required containers
+and open a shell by running:
+```sh
+./tool/docker_test.sh
+```
+Once inside the shell, you can run commands as described above to test your
+code changes.
+
+:warning: Note: This environment only supports unit tests. Integration tests will not work.
```
simolus3, line 107: This renders as a nice block:

```suggestion
> [!WARNING]
> This environment only supports unit tests. Integration tests will not work.
```

CONTRIBUTING.md
```diff
@@ -73,12 +73,56 @@ used to construct common queries.
...
+
+**Real-time file changes in Docker**
+By default, the Docker container won't reflect local file changes. To avoid
+rebuilding the container after every edit, mount your working directories
+directly when opening the shell:
+```sh
+docker-compose run --rm \
+  -v "./sqlparser/test:/app/sqlparser/test" \
+  -v "./sqlparser/lib/src/analysis:/app/sqlparser/lib/src/analysis" \
```
simolus3, line 116: I'm not sure we should recommend mounting individual subdirectories, that sounds error prone? Would mounting a package directory work?

I suppose the reason we can't mount the entire project is a conflict with `.dart_tool` dependency paths, right?

extras/integration_tests/web_wasm/test/drift_wasm_test.dart
```diff
@@ -1,5 +1,6 @@
 // ignore: implementation_imports
 import 'dart:async';
```
simolus3, line 2: Given the dependencies that this needs (two browsers...), I'm also fine with this not working in containers and removing the Chrome/FF services.

Dockerfile
```diff
@@ -1,32 +1,24 @@
...
+RUN whoami
+
+COPY --chown=flutter:flutter . /app/
+WORKDIR /app
+RUN dart pub get && melos bootstrap
+RUN ./tool/upgrade_all.sh
+
+ENTRYPOINT ["/app/tool/docker/entrypoint.sh"]
+
+CMD export LD_LIBRARY_PATH=/tmp/sqlite/out ; ./tool/test_all.sh
```
simolus3, line 23: We shouldn't need the `LD_LIBRARY_PATH` anymore, as `package:sqlite3` now uses build hooks to bundle SQLite.

tool/upgrade_all.sh
```diff
@@ -1,2 +1,2 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
simolus3, line 1: We should be able to remove this command / script too, since pub workspaces are resolved as a coherent unit we don't need to run `pub upgrade` in every package.

tool/docker/haproxy.cfg
```diff
@@ -0,0 +1,43 @@
+## This file is used because in the docker run, localhost doesn't work (easily)
+## but dev and testing without docker should stay possible.
```
simolus3, line 2: Maybe it's just me, but it still took me a while to figure out why this is necessary (podman pods allowing multiple images in the same network namespace has spoiled me lol).

```suggestion
## With the docker-compose setup, different services run in separate containers and we can't
## easily connect to them via localhost.
## This configures a proxy to connect to containers via localhost, allowing connection strings in
## tests to work in containers.
```

Dockerfile
```diff
@@ -1,32 +1,24 @@
-FROM dart
+FROM gmeligio/flutter-android:3.44.1
```
simolus3, line 1: This is a fairly large image, and we only need Flutter for unit tests (so definitely no full Android SDK). I'm not sure if there's a bare-bones Flutter image, but I see that e.g. `gmeligio/flutter-web` is much smaller already.

simolus3 (commented):
Thanks for the contribution. I didn't run this myself yet, but I have some inline comments.

Also as a broader item, I wonder if it could perhaps make sense to use dev containers for this instead? I haven't used them in a while, but it sounds like the better editor integration might be worth it, do you have experience with that?

jonasbadstuebner:
Thank you for looking into this - those are all very valid points and I’m going to address them next week.

I don’t have experience with dev containers, but from what I understand they are only a wrapper around a setup very similar to what I did, more or less abstracting away the docker-compose commands.
I don’t use VSCode or any IDE and have to look into if this is easily testable for me then.

simolus3:
>  I don’t use VSCode or any IDE and have to look into if this is easily testable for me then.

No worries then, I think having a working container setup is a nice improvement on its own.

PR #3826, opened 2026-07-03
Fix read pool starting dequeued selects in the wrong zone

### LINKED ISSUE #3824: Read pool: CancellationException leaks to unrelated queries because _QueryExecutorPool._run() starts dequeued selects in the previous caller's Zone
## Summary

When a database uses a read pool (`NativeDatabase.createInBackground(..., readPool: N)` or `MultiExecutor.withReadPool`), an unrelated, plain awaited select (`.get()`, `.getSingle()`, ...) can fail with `CancellationException` ("Operation was cancelled") if it was queued in `_QueryExecutorPool` behind a stream-query fetch whose subscription got cancelled while the fetch was running.

This is, I believe, the unresolved remainder of #2323, which was closed for lack of a repro. Below is a deterministic repro plus a root-cause trace. Also related: #2180, #2818, #3585.

## Environment

- Reproduced on drift 2.26.1, 2.31.0, and 2.34.0 (latest at the time of writing). The relevant logic in `connection_pool.dart` and `client_impl.dart` is unchanged from 2.26.1 through 2.34.0 (only reformatted).
- Dart 3.6.1 and 3.10.8, macOS arm64 (also observed in a production Flutter POS app on multiple platforms)
- `NativeDatabase.createInBackground(file, readPool: 3)` with WAL

## Root cause

1. Stream query fetches run inside a cancellation zone: `runCancellable<Rows>(_fetcher.fetchData, token: operation)` in `stream_queries.dart`. The token is cancelled when the last listener unsubscribes (`_onCancelOrPause` -> `markAsClosed` -> `_cancelRunningQueries`).

2. With a read pool, every out-of-transaction select goes through `_QueryExecutorPool.runSelect`. When all readers are busy, the select is queued as a `_PendingSelect` and started later by `_run()`.

3. The bug is in `_run()` (`connection_pool.dart`):

```dart
completer.completer.complete(Future.sync(() async {
  try {
    return await executor.runSelect(completer.statement, completer.args);
  } finally {
    _running.remove(completer);
    _idleExecutors.add(executor);
    _run(); // <- runs in the zone of the select that JUST FINISHED
  }
}));
```

   The recursive `_run()` in the `finally` executes in the `Zone` of the select that just completed, not in the zone of the queued select it is about to start. Drift's cancellation tokens live in zone values, so the dequeued select now executes under a foreign cancellation token.

4. On the remote (isolate) client, `_BaseExecutor._runRequest` (`client_impl.dart`) does:

```dart
checkIfCancelled();          // throws if the (foreign!) token is already cancelled
...
doOnCancellation(() { ... }); // otherwise ties this request to the foreign token
```

   So the innocent dequeued select either throws `CancellationException` immediately, or is registered for cancellation on a token it has nothing to do with and gets cancelled mid-flight later. Either way, the exception completes the future of unrelated user code. The same check also exists for non-isolate sequential executors in `engines.dart` (`_synchronized` -> `checkIfCancelled`).

Note that this would be incorrect even if cancelling pool-queued selects were acceptable policy: the dequeued select is not attached to its own cancellation token, but to the token of whichever select happened to free up the pool slot. The pairing between token and query is timing-dependent and arbitrary. A queued fetch belonging to a still-active stream A can be cancelled because unrelated stream B was unsubscribed, and a plain `.get()` that was never cancellable in the first place can fail with a cancellation it cannot opt out of (before 2.28, `CancellationException` was `@internal`, so callers could not even catch it by name).

## Deterministic repro

`pubspec.yaml`: depends on `drift: ^2.34.0`, `sqlite3: ^3.3.4` (also reproduces as-is on `drift: 2.26.1` with `sqlite3: ^2.4.0`).

```dart
import 'dart:async';
import 'dart:io';

import 'package:drift/native.dart';

class ReproDb extends GeneratedDatabase {
  ReproDb(super.e);

  @override
  Iterable<TableInfo<Table, dynamic>> get allTables => const [];

  @override
  int get schemaVersion => 1;
}

// A select that takes a while on the reader isolate.
String slowQuery(int n) =>
    'WITH RECURSIVE cnt(x) AS (SELECT 1 UNION ALL SELECT x+1 FROM cnt WHERE x < $n) '
    'SELECT count(x) AS c FROM cnt';

Future<void> main(List<String> args) async {
  // Pass "control" to skip the mid-flight cancellation; the innocent
  // select then completes normally, isolating the cancel as the trigger.
  final control = args.contains('control');
  final file = File('${Directory.systemTemp.path}/drift_repro_$pid.sqlite');
  if (file.existsSync()) file.deleteSync();

  final db = ReproDb(NativeDatabase.createInBackground(
    file,
    readPool: 2,
    setup: (database) {
      database.execute('PRAGMA journal_mode = WAL;');
      database.execute('PRAGMA synchronous = NORMAL;');
    },
  ));

  // Open the database and all pooled connections before the race.
  await db.customSelect('SELECT 1').get();

  // Occupy both pooled readers with in-flight stream fetches.
  // Different SQL so the two streams don't share a query key.
  final long = db.customSelect(slowQuery(30000000)).watch().listen((_) {});
  final short = db.customSelect(slowQuery(8000000)).watch().listen((_) {});

  // Let both fetches get dispatched to the reader connections.
  await Future<void>.delayed(const Duration(milliseconds: 200));

  // Queue an innocent, unrelated select behind the saturated pool.
  final innocent = db.customSelect('SELECT 42 AS answer').get();

  // Cancel the short stream while its fetch is still running.
  if (!control) {
    await short.cancel();
  }

  try {
    final rows = await innocent;
    print('NO BUG: innocent select returned ${rows.first.data}');
  } catch (e) {
    print('BUG REPRODUCED: innocent select threw: "$e" (${e.runtimeType})');
    exitCode = 1;
  }

  await long.cancel();
  await db.close();
  if (file.existsSync()) file.deleteSync();
}
```

Output (reproduces on every run):

```
--- control (no cancel) ---
NO BUG: innocent select returned {answer: 42}
--- bug, run 1 ---
BUG REPRODUCED: innocent select threw: "Operation was cancelled" (CancellationException)
--- bug, run 2 ---
BUG REPRODUCED: innocent select threw: "Operation was cancelled" (CancellationException)
--- bug, run 3 ---
BUG REPRODUCED: innocent select threw: "Operation was cancelled" (CancellationException)
```

Stack trace of the failing `await innocent` (drift 2.34.0). Read bottom-up: the unrelated select is executing as a continuation of the cancelled stream's fetch, started from the pool's `finally` dequeue:

```
#0      checkIfCancelled (package:drift/src/runtime/cancellation_zone.dart:97:5)
#1      _BaseExecutor._runRequest (package:drift/src/remote/client_impl.dart:99:5)
#2      _BaseExecutor.runSelect (package:drift/src/remote/client_impl.dart:153:26)
#3      DatabaseConnection.runSelect (package:drift/src/runtime/api/connection.dart:136:17)
#4      _QueryExecutorPool._run.<anonymous closure> (package:drift/src/runtime/executor/connection_pool.dart:82:33)
#5      new Future.sync (dart:async/future.dart:325:27)
#6      _QueryExecutorPool._run (package:drift/src/runtime/executor/connection_pool.dart:80:14)
#7      _QueryExecutorPool._run.<anonymous closure> (package:drift/src/runtime/executor/connection_pool.dart:86:11)
<asynchronous suspension>
#8      NonNullableCancellationExtension.resultOrNullIfCancelled (package:drift/src/runtime/cancellation_zone.dart:65:24)
<asynchronous suspension>
#9      QueryStream.fetchAndEmitData (package:drift/src/runtime/executor/stream_queries.dart:343:20)
<asynchronous suspension>
```

Timeline of the failing run:

1. Two `.watch()` fetches occupy both pooled readers (each fetch runs in its own cancellation zone).
2. `SELECT 42` is queued in `_QueryExecutorPool._queue`.
3. `short.cancel()` cancels the second stream's token; the statement still completes on the reader.
4. When it completes, the `finally` calls `_run()` inside the cancelled zone, which dequeues `SELECT 42` and calls `executor.runSelect` there.
5. `checkIfCancelled()` in `_runRequest` sees the cancelled foreign token and throws; `await innocent` fails with "Operation was cancelled".

## Real-world impact

In our POS app this hits the payment flow: committing the payment transaction triggers table update notifications, all active `.watch()` streams re-fetch at once and saturate the pool, and UI rebuilds cancel some of those subscriptions mid-fetch. The awaited re-read of the just-committed order then throws "Operation was cancelled" even though the payment committed, so the cashier sees an error for a successful payment. Any awaited out-of-transaction select can be affected; this is not specific to one query.

Unlike the situation described in #3585, these exceptions are not harmless here: they complete futures that unrelated application code is awaiting.

## Suggested fix (verified)

Capture the enqueuer's zone in `_PendingSelect` and start dequeued work in that zone. I applied this patch to a local copy of drift 2.34.0 and the repro above passes on every run (while the control behavior is unchanged), which also confirms the diagnosis:

```dart
class _PendingSelect {
  _PendingSelect(this.statement, this.args)
      : completer = Completer<List<Map<String, Object?>>>(),
        zone = Zone.current;

  final Zone zone;
  ...
}

// in _run():
completer.completer.complete(completer.zone.run(() => Future.sync(() async {
  ...
})));
```

This both stops foreign tokens from leaking into unrelated selects and preserves the intended behavior for queued selects that belong to a cancellable stream fetch themselves (their own token keeps working).

zkang93:
> 🤖 Authored by Claude Fable 5 with [Claude Code](https://claude.com/claude-code); reviewed by Ang Zheng Kai ([@zkang93](https://github.com/zkang93)).

## What this fixes

`_QueryExecutorPool._run()` starts a queued select from the `finally` block of whichever select freed up the executor. That `finally` continuation runs in the *previous* select's zone, so the dequeued select executed under a foreign zone. Since drift's cancellation tokens are zone-scoped, this had two consequences:

- A plain awaited select (`.get()`, `.getSingle()`, ...) queued behind a saturated read pool could fail with `CancellationException` when an unrelated stream query subscription was cancelled mid-fetch, completing futures that application code is awaiting (see #3824 for a deterministic repro and full trace).
- Conversely, a queued select issued from a zone that *was* cancelled escaped its own cancellation, because its token was no longer in scope when the select started.

## The fix

`_PendingSelect` now captures `Zone.current` at enqueue time, and `_run()` starts dequeued work via `zone.run(...)`. This restores the invariant that dart:async callbacks maintain everywhere else — work runs in the zone it was registered from — and makes the queued path behave exactly as if an executor had been free at enqueue time. Queued selects that belong to a cancellable stream fetch keep responding to their own token; unrelated selects are no longer affected by foreign tokens.

## Tests

Two regression tests in `connection_pool_test.dart`, using mock executors that check for cancellation like real executors do:

- a select queued outside any cancellation zone completes normally even when the select that freed the reader had its zone cancelled (failed with `CancellationException` before this fix, with the same stack trace as the issue);
- a queued select issued from a cancelled zone is still cancelled once it starts (silently ran to completion before this fix).

Also verified against the deterministic isolate repro from #3824 (fails on every run before, passes on every run after, control unchanged) and the full drift VM test suite (899 passed).

Closes #3824

### DIFF
--- a/drift/CHANGELOG.md
+++ b/drift/CHANGELOG.md
@@ -1,3 +1,10 @@
+## 2.34.1
+
+- Fix reads queued in a `MultiExecutor` connection pool being resumed in the
+  wrong zone once an executor became available. Because Drift's cancellation
+  tokens are zone-scoped, this could cause queries to be cancelled or observed
+  in the wrong zone (#3824).
+
 ## 2.34.0
 
 - Use `BEGIN IMMEDIATE` to start transactions.
--- a/drift/lib/src/runtime/executor/connection_pool.dart
+++ b/drift/lib/src/runtime/executor/connection_pool.dart
@@ -27,11 +27,20 @@ sealed class MultiExecutor extends QueryExecutor {
 
 class _PendingSelect {
   _PendingSelect(this.statement, this.args)
-    : completer = Completer<List<Map<String, Object?>>>();
+    : completer = Completer<List<Map<String, Object?>>>(),
+      zone = Zone.current;
 
   final String statement;
   final List<Object?> args;
   final Completer<List<Map<String, Object?>>> completer;
+
+  /// The zone from which the select was issued.
+  ///
+  /// When this select is dequeued after waiting for an executor, it must be
+  /// started in this zone: Drift's cancellation tokens are zone-scoped, and
+  /// the dequeueing [_QueryExecutorPool._run] call runs in the zone of
+  /// whichever earlier select happened to free up the executor.
+  final Zone zone;
 }
 
 class _QueryExecutorPool {
@@ -77,15 +86,20 @@ class _QueryExecutorPool {
     _running.add(completer);
 
     completer.completer.complete(
-      Future.sync(() async {
-        try {
-          return await executor.runSelect(completer.statement, completer.args);
-        } finally {
-          _running.remove(completer);
-          _idleExecutors.add(executor);
-          _run();
-        }
-      }),
+      completer.zone.run(
+        () => Future.sync(() async {
+          try {
+            return await executor.runSelect(
+              completer.statement,
+              completer.args,
+            );
+          } finally {
+            _running.remove(completer);
+            _idleExecutors.add(executor);
+            _run();
+          }
+        }),
+      ),
     );
   }
 }
--- a/drift/test/engines/connection_pool_test.dart
+++ b/drift/test/engines/connection_pool_test.dart
@@ -1,4 +1,7 @@
+import 'dart:async';
+
 import 'package:drift/drift.dart';
+import 'package:drift/src/runtime/cancellation_zone.dart';
 import 'package:mockito/mockito.dart';
 import 'package:test/test.dart';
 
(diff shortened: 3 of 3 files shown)


simolus3 (commented):
Thanks for the contribution, I agree with the fix 👍 Could you also add a changelog entry to the drift package (a pending `2.34.1` entry)?

zkang93:
I have added the version 2.34.1 entry to the drift package CHANGELOG.md.

simolus3 (approved):

PR #3835, opened 2026-07-13
fix: don't report a boolean default TRUE/1 as a schema difference (#3738)

### LINKED ISSUE #3738: Boolean column with a default value always fails in SchemaVerifier.migrateAndValidate (TRUE vs 1)
Consider a boolean column with a default value:

```dart
class UserTable extends Table {
  BoolColumn get isFirstLogin =>
      boolean().withDefault(const Constant<bool>(true))();
}
```

SchemaVerifier.migrateAndValidate fails on schema validation:

```
Schema does not match
user_table:
 columns:
  is_first_login:
   Not equal: `NOT NULL DEFAULT TRUE CHECK (is_first_login IN (0, 1))` (expected) and `NOT NULL DEFAULT (1) CHECK ("is_first_login" IN (0, 1))` (actual)
```

This is the simplest migration/schema test:
```dart
void main() {
  driftRuntimeOptions.dontWarnAboutMultipleDatabases = true;
  final SchemaVerifier verifier = SchemaVerifier(GeneratedHelper());

  group('simple database migrations', () {
    // These simple tests verify all possible schema updates with a simple (no
    // data) migration. This is a quick way to ensure that written database
    // migrations properly alter the schema.
    const List<int> versions = GeneratedHelper.versions;
    for (final (int i, int fromVersion) in versions.indexed) {
      group('from $fromVersion', () {
        for (final int toVersion in versions.skip(i + 1)) {
          test('to $toVersion', () async {
            final mc.InitializedSchema<Database> schema = await verifier
                .schemaAt(fromVersion);
            final SqlDatabase database = SqlDatabase(schema.newConnection());
            await verifier.migrateAndValidate(
              database,
              toVersion,
              options: const mc.ValidationOptions(validateDropped: true),
            );
            await database.close();
          });
        }
      });
    }
  });
}
```

This project has both SQLite and Postgres dialects enabled in build.yaml. Perhaps it contributes to the issue?

Yusufihsangorgel:
A boolean column with a default value fails `SchemaVerifier.migrateAndValidate`, even though the schema is semantically identical:

```
Not equal: `NOT NULL DEFAULT TRUE CHECK (is_first_login IN (0, 1))` (expected)
       and `NOT NULL DEFAULT (1) CHECK ("is_first_login" IN (0, 1))` (actual)
```

Booleans are stored as integers in SQLite, so a `DEFAULT TRUE` emitted by one schema snapshot and a `DEFAULT 1` emitted by another describe the same column. But `find_differences` compared the default expressions textually and flagged a spurious difference, which breaks migration tests for any table with a boolean default (#3738).

### Fix

When comparing column constraints, a boolean default written as `TRUE`/`FALSE` is now treated as equal to the same value written in its integer form `1`/`0`. Genuinely different defaults are still reported as differences.

### Tests

Added cases to `find_differences_test.dart`: a boolean default written as an integer, the exact #3738 scenario (boolean default + `IN (0, 1)` check), and a guard that a *genuinely* different default is still detected as a change. Reverting the fix makes the first two fail; with the fix the whole `find_differences` suite passes.

Closes #3738.

### DIFF
--- a/drift_dev/lib/src/services/schema/find_differences.dart
+++ b/drift_dev/lib/src/services/schema/find_differences.dart
@@ -187,19 +187,73 @@ class FindSchemaDifferences {
       );
     }
 
-    if (options.validateColumnConstraints) {
-      try {
-        enforceEqualIterable(ref.constraints, act.constraints);
-      } catch (e) {
-        final firstSpan = ref.constraints.spanOrNull?.text ?? '';
-        final secondSpan = act.constraints.spanOrNull?.text ?? '';
-        return FoundDifference(
-          'Not equal: `$firstSpan` (expected) and `$secondSpan` (actual)',
-        );
-      }
+    if (options.validateColumnConstraints &&
+        !_columnConstraintsMatch(ref.constraints, act.constraints)) {
+      final firstSpan = ref.constraints.spanOrNull?.text ?? '';
+      final secondSpan = act.constraints.spanOrNull?.text ?? '';
+      return FoundDifference(
+        'Not equal: `$firstSpan` (expected) and `$secondSpan` (actual)',
+      );
     }
 
     return const Success();
+  }
+
+  /// Compares two column constraint lists like [enforceEqualIterable], but
+  /// tolerates a boolean default (`TRUE`/`FALSE`) being written as its
+  /// equivalent integer (`1`/`0`), optionally wrapped in parentheses.
+  ///
+  /// Booleans are stored as integers in SQLite, so a `DEFAULT TRUE` emitted by
+  /// an older drift version and a `DEFAULT (1)` emitted by a newer one describe
+  /// the same column and must not be reported as a schema difference.
+  /// See https://github.com/simolus3/drift/issues/3738.
+  bool _columnConstraintsMatch(
+    List<ColumnConstraint> reference,
+    List<ColumnConstraint> actual,
+  ) {
+    if (reference.length != actual.length) return false;
+
+    for (var i = 0; i < reference.length; i++) {
+      final ref = reference[i];
+      final act = actual[i];
+
+      // Only treat a boolean/integer default mismatch as equal when the two
+      // constraints are otherwise identical, including an optional
+      // `CONSTRAINT <name>`. Differing names fall through to enforceEqual.
+      if (ref is Default && act is Default && ref.name == act.name) {
+        final refValue = _booleanOrNumericLiteral(ref.expression);
+        final actValue = _booleanOrNumericLiteral(act.expression);
+
+        if (refValue != null && actValue != null) {
+          if (refValue != actValue) return false;
+          continue;
+        }
+      }
+
+      try {
+        enforceEqual(ref, act);
+      } catch (_) {
+        return false;
+      }
+    }
+
+    return true;
+  }
+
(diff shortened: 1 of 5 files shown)


drift_dev/lib/src/services/schema/find_differences.dart
```diff
@@ -187,19 +187,70 @@ class FindSchemaDifferences {
...
+    for (var i = 0; i < reference.length; i++) {
+      final ref = reference[i];
+      final act = actual[i];
+
+      if (ref is Default && act is Default) {
+        final refValue = _booleanOrNumericLiteral(ref.expression);
+        final actValue = _booleanOrNumericLiteral(act.expression);
+
+        if (refValue != null && actValue != null) {
+          if (refValue != actValue) return false;
```
simolus3, line 228: These are rarely used, but constraints can also have a `name` attached to them and we should check that those are equal.

simolus3 (commented):
Thanks! I'm still not entirely sure why we'd sometimes generate it as a boolean literal or a expression, but I agree it makes sense to treat `TRUE` and `1` as equal for default constraints.

There's a second schema verifier implementation for an upcoming drift release in `future/drift_sqlite/lib/src/schema_verifier/find_differences.dart`, could you please apply the changes there as well?

drift_dev/lib/src/services/schema/find_differences.dart
Yusufihsangorgel: Good catch. The `TRUE`/`1` tolerance now only applies when both `Default` constraints have the same `name`; when the names differ it falls through to `enforceEqual`, which already asserts name equality in `visitColumnConstraint`, so a differently-named (or named-vs-unnamed) default is still reported. Added a regression test — `with a boolean default under a different constraint name` — covering equal default values under different `CONSTRAINT` names.

Yusufihsangorgel:
Thanks for the quick review! Pushed an update:

- Constraint `name` equality is now checked before applying the `TRUE`/`1` tolerance (details in the inline thread), plus a regression test for it.
- Applied the same change to the second verifier at `future/drift_sqlite/lib/src/schema_verifier/find_differences.dart`.

On why both forms show up: from what I've seen it's a cross-version rendering difference rather than one version emitting both — the reference schema (a serialized snapshot) renders the boolean default as `TRUE`, while the current generator emits `DEFAULT (1)`, so it surfaces in the `SchemaVerifier.migrateAndValidate` path that compares a stored snapshot against freshly generated SQL. I chose to tolerate it in the verifier rather than normalize the rendering, since absorbing that kind of cross-version cosmetic difference is what the verifier is for — but happy to change the approach if you'd rather unify it at the generation side.

simolus3 (approved):
Thanks!

simolus3:
Released in `drift_dev` version `2.34.4`.

ViveNoctem:
@simolus3
> Thanks! I'm still not entirely sure why we'd sometimes generate it as a boolean literal or a expression, but I agree it makes sense to treat `TRUE` and `1` as equal for default constraints.

I'm a bit late, but a few weeks ago, I looked into this with `git bisect`.
This commit seems to have changed the behavior.
https://github.com/simolus3/drift/commit/45716ab3bee13956354a6549d54544e5dbae79e9
But I didn't have the time to look further into it back then.

PR #3838, opened 2026-07-21
Escape reserved Dart keywords in SQL column names

mohanedy:
## Problem

A SQL column whose name is a reserved Dart keyword (e.g. `class`) generates invalid Dart. The column name is turned into a Dart identifier via `ReCase(name).camelCase` with no keyword check, so the generator emits code like:

```dart
late final GeneratedColumn<int> class = GeneratedColumn<int>('class', ...);
```

which fails to parse (`Can't have modifier 'late' here`, etc.).

This is most visible through `make-migrations`. The serialized schema snapshot embeds a `fixed_sql` block (the real `CREATE TABLE` statements). When reading it back, `SchemaReader` prefers `fixed_sql` and reconstructs tables directly from SQL (`extractDriftElementsFromSql`), which bypasses the serialized `getter_name`. So even a table defined in Dart as `IntColumn get productClass => integer().named('class')()` — which normally generates fine — crashes when its schema snapshot is turned into a migration test database, because the reconstructed column falls back to the raw SQL name `class`.

Stack (drift_dev 2.33.0, still present on `develop` 2.34.4):

```
Could not format because the source could not be parsed:
line ...: Can't have modifier 'late' here.
  #4 GenerateUtils.generateSchemaCode (.../schema/generate_utils.dart)
  #5 _MigrationTestEmitter.writeTestDatabases (.../make_migrations.dart)
```

## Fix

- `dartNameForSqlColumn` now appends `$` when the derived name is a reserved Dart keyword (`class` → `class$`), using the analyzer's authoritative `Keyword` list. Built-in identifiers (e.g. `mixin`) are valid identifiers and left untouched.
- The drift table resolver (`table.dart`) now routes column names through `dartNameForSqlColumn` instead of calling `ReCase(...).camelCase` directly, so both `.drift` files and SQL-reconstructed schemas produce valid identifiers. Views already used the shared helper.

The SQL column name is unchanged — only the generated Dart getter/field/parameter identifier is escaped.

## Tests

Added a resolver regression test (`test/analysis/resolver/drift/table_test.dart`) covering a `class` column (escaped to `class$`), a built-in identifier (`mixin`, unchanged), and a normal column. Full `test/analysis` and `test/writer` suites pass locally (393 tests).

Also verified end-to-end against a real project: `make-migrations` previously crashed on a `class` column and now completes, and the generated `schema_vN.dart` analyzes clean.

### DIFF
--- a/drift_dev/CHANGELOG.md
+++ b/drift_dev/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 2.34.5
+
+- Escape SQL column names that are reserved Dart keywords (e.g. `class`) when
+  deriving the Dart getter name, fixing invalid generated code for tables read
+  from SQL (such as the schema snapshots emitted by `make-migrations`).
+
 ## 2.34.4
 
 - Schema verifier: Treat `TRUE`/`FALSE` as equal to `1`/`0`, respectively.
--- a/drift_dev/lib/src/analysis/resolver/drift/table.dart
+++ b/drift_dev/lib/src/analysis/resolver/drift/table.dart
@@ -11,6 +11,7 @@ import '../../driver/state.dart';
 import '../../results/results.dart';
 import '../intermediate_state.dart';
 import '../resolver.dart';
+import '../shared/column_name.dart';
 import '../shared/dart_types.dart';
 import '../shared/data_class.dart';
 import 'element_resolver.dart';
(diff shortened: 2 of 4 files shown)


simolus3 (approved):
Thank you for your contribution! This looks good to me.

mohanedy:
The `CI / Integration Tests` pipeline is failing due to an unrelated warning, so please @simolus3 let me know if I can help with fixing that as well.

ahmeddhus:
I've been waiting for this fix. Great work! 🚀

simolus3:
I've fixed that warning, thanks for the ping! I'll release these changes later today.

PR #3839, opened 2026-07-21
Fix quoted link targets and two typos in the docs

meliharik:
Seven markdown links across four doc pages wrap their target in literal quotes, so the rendered href includes the quote characters and 404s:

| file | link |
|---|---|
| `type_converters.md:33` | `[compiled custom queries]("/queries/custom")` — also an absolute path that doesn't exist; now points at `sql_api/custom_queries.md` like `faq.md:161` does |
| `type_converters.md:247` | `[expressions]('dart_api/expressions.md')` |
| `type_converters.md:255` | `[drift file]('sql_api/drift_files.md')` |
| `type_converters.md:256` | `[builder option]('generation_options/index.md')` |
| `dart_api/schema_inspection.md:14` | `[getting started]('../setup.md')` |
| `migrations/tests.md:132` | `[DevTools extension]('../Tools/devtools.md')` — also wrong case; the directory is `tools/` |
| `tools/index.md:72` | `[the guide]("../Migrations/tests.md")` — also wrong case; the directory is `migrations/` |

All targets exist under the correct lowercase paths (verified against `git ls-files`, since a case-insensitive filesystem hides the case bugs locally).

Two small prose fixes in the same section of `type_converters.md`:

- line 249: "will **compare not apply** the type converter" → "will not apply the type converter"
- line 252: `` `equasValue` `` → `` `equalsValue` `` (matching the API link right above it)

Docs only.


simolus3 (approved):
Thanks!

PR #3846, opened 2026-08-24
Document sharing one DriftIsolate via IsolateNameServer #2095

### LINKED ISSUE #2095: Publish working example projects that shows how to use the DriftIsolates?
Maybe it is just me but honestly I have no idea how to implement this: https://drift.simonbinder.eu/docs/advanced-features/isolates/

I'd like fully working examples where I can see in which files the given method is, how it is used throughout the app (is it stored somewhere or open a new appdatabase whenever needed), what is in the main etc...

Do I just randomly call into this db from main thread's method middle part and it will run next code on a new thread? Or I have to call the method in a new isolate to make it work?

I'd like to run some inserts in an isolate since it is blocking the UI but I can't decode what's in the doc. A complete project with examples would help a lot...

AzazelSensei:
People hitting this usually spawn a DriftIsolate in the UI and again in a background service. Those are two databases, so `watch()` in the foreground never sees the background writes.

This adds that warning to the independent-isolates section, plus the lookup-or-spawn snippet using IsolateNameServer that you already confirmed on the issue. Flutter apps in the same engine can keep using `shareAcrossIsolates: true` instead.

Fixes #2095

### DIFF
--- a/docs/content/isolates.md
+++ b/docs/content/isolates.md
@@ -248,12 +248,23 @@ no clear communication scheme between isolates.
 Still, you may want to share a live drift database between your UI engine and potential background engines,
 even without them directly knowing about each other.
 
+If each isolate calls `DriftIsolate.spawn` on its own, you get two independent executor isolates.
+Writes from a background service then will not update `watch()` streams in the UI isolate.
+The same happens if both sides open a plain `NativeDatabase` on the same file.
+
 An [IsolateNameServer](https://api.flutter.dev/flutter/dart-ui/IsolateNameServer-class.html) from `dart:ui` can
 be used to transparently share a drift isolate between such workers.
 You can store the [`connectPort`](https://drift.simonbinder.eu/api/isolate/driftisolate/connectport) of a `DriftIsolate`
 under a specific name to look it up later.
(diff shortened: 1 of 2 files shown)


simolus3 (commented):
Thanks for improving documentation on this!

Reading the snippet though, I wonder if we should instead add a link to `drift_flutter` sources as an inspiration instead of adding a standalone snippet. `IsolateNameServer` is a brittle API, and one needs to defend against cases like a hot restart where some isolates are restarted but the dead port is still registered in the isolate server (`drift_flutter` uses timeouts and pings for that).

But more importantly, `IsolateNameServer` only exists for Flutter and `drift_flutter` already does the right thing. So I suppose what the docs should suggest is a stronger focus on using `drift_flutter` instead of trying to wire up isolates manually.

AzazelSensei:
Dropped the IsolateNameServer snippet. The page now points at drift_flutter with shareAcrossIsolates, and at native.dart if someone still needs a custom handshake.

simolus3:
I have merged your PR with some minor changes in 906593ee4f47d31ea63750ac3b97b20fbe06c23b.

PR #3852, opened 2026-09-04
Remove web/channel_legacy.dart

biggs0125:
The Dart SDK will soon drop support for dart:html. To give the community time to migrate we would like to start deleting usages of this library from heavily used packages early.


simolus3 (approved):
Thanks! It's worth noting that the only affected file is behind a conditional export, so old package versions continue to work without `dart:html` too (as long as the deprecated extension isn't used).

biggs0125:
Thanks for cleaning this up and getting this landed!

mkustermann:
@simolus3 Would it be possible to publish this version to pub.dev so users of `package:drift` can make their apps free of legacy interop?

simolus3:
I have a few other changes I want to get in, but I'll prepare a release today 👍

simolus3:
Released in `drift` version `2.35.0`.

Issue #3367, opened 2024-12-03
drift_flutter force to use ".sqlite" suffix 


lucasdidur:
Because of 

```dart
return File(p.join(dbFolder.path, '$name.sqlite'));
```

Old databases can't not be changed or reused if the old file does not contains .sqlite.  So old database and data is lost (my case).

Solution:

Add a named parameter extension or remove the '.sqlite' part.

simolus3:
You can use `driftDatabase(native: DriftNativeOptions(databasePath: () async => ...))` to migrate to `drift_flutter` while otherwise keeping your existing database path.

I will update the documentation to mention how to migrate from the old setup to `drift_flutter`, but I'm not sure where to put it yet.

onur-kesim:
The documentation part of this is now in — #3862 added a `drift_flutter` section to the migration guide:

> `driftDatabase()` does not use the same file without the `databasePath` option: It stores a `<name>.sqlite` file under `getApplicationDocumentsDirectory()`, which is a different file than the one opened above. To keep using your existing database, pass a `databasePath` callback that returns its exact path.

It sits in `docs/content/guides/migrating_to_drift.md`, next to the `sqflite` and plain-`drift` snippets, with a worked example.

That covers the documentation side. It does not cover the original request — @lucasdidur asked for an API change (a named parameter, or dropping the `.sqlite` suffix), and that's a separate decision. If the documented workaround is the intended answer, this can probably be closed; if the API question is still open, retitling might fit better.

simolus3:
Thanks again for the documentation PR! Given that `databasePath` is documented more prominently now, I don't think this needs an API change.

PR #3862, opened 2026-09-12
docs: keep the existing database file when migrating to drift

onur-kesim:
Refs #3367.

In "Opening a drift database", the guide says the drift snippets open *the same* database as the code above them, but the previous code opens `demo.db` while both drift snippets opened `db.sqlite`. Following the guide literally opens a new, empty database instead of the existing one. Both are now `demo.db`.

The second part is what #3367 asks for: `driftDatabase()` stores `<name>.sqlite` under `getApplicationDocumentsDirectory()` by default, so someone switching to `drift_flutter` silently ends up on a different file. The new snippet passes `databasePath` to keep the existing one, and the text notes that `databasePath` and `databaseDirectory` are mutually exclusive.

Documentation only. The two Dart snippets analyze cleanly against the `docs` package (drift, drift_flutter, path, path_provider). The `drift_sqflite` snippet only changed its file name; `drift_sqflite` is not a dependency of the docs package, so it cannot be analyzed there.

On placement: in #3367 you wrote that you were not sure where to put this. The migration guide seemed like the natural home because it already maps old setups to drift, but I am happy to move it into the setup guide instead.


simolus3:
Thanks for the contribution! I have merged this in d9f88e5b27531cfcfcd198e0e440963198175801, the changes will be visible with the next drift release.

PR #3865, opened 2026-09-15
Web: Flush IndexedDB after transactions and RETURNING statements

### LINKED ISSUE #3864: [Web] IndexedDB: writes inside a transaction are not flushed on commit and are lost on reload
**Describe the bug**

On web with an IndexedDB storage implementation (`sharedIndexedDb` / `unsafeIndexedDb`), writes made inside a `transaction()` are committed to the in-memory VFS but never flushed to IndexedDB. They are only persisted if some later statement runs outside a transaction; if the page is reloaded before that, the committed transaction is silently lost.

**Cause**

Since 2.34.2 ("Flush IndexedDB writes after statements"), `_WasmDelegate` flushes after a statement only when it is not in a transaction (`drift/lib/wasm.dart`):

```dart
Future _runWithArgs(String statement, List<Object?> args) async {
  runWithArgsSync(statement, args);

  if (!isInTransaction) {
    await _flush();
  }
}
```

But the `COMMIT` itself goes through this path while the flag is still set. `_StatementBasedTransactionExecutor.send()` (`drift/lib/src/runtime/executor/helpers/engines.dart`) runs the commit first and clears the flag afterwards:

```dart
Future<void> send() async {
  if (!_ensureOpenCalled) return;

  await runCustom(_commitCommand, const []);   // isInTransaction is still true -> no flush
  _release();                                  // sets isInTransaction = false
}
```

So a top-level transaction never flushes on commit. The same code is still on `develop` as of 2.35.0.

**Reproduction**

```dart
// Web build, IndexedDB storage chosen (e.g. served without COOP/COEP headers).
await db.transaction(() async {
  await db.into(db.items).insert(ItemsCompanion.insert(name: 'lost'));
});
// Reload the page without issuing any other write.
// After reload, the row is gone.
```

Doing a plain write (outside a transaction) before reloading persists the earlier transaction too, since the flush writes all dirty blocks.

We hit this in a release web build (Chrome/Edge, served from GitHub Pages, so no COOP/COEP → IndexedDB). Our first-run setup inserts the initial account inside a transaction; after a refresh the account table was empty. Inspecting the `blocks` store in IndexedDB before the refresh confirmed the row was never written there. Native builds are unaffected.

**Workaround**

Running any statement outside the transaction after it commits triggers the flush, e.g. overriding `transaction` on the database class:

```dart
@override
Future<T> transaction<T>(Future<T> Function() action, {bool requireNew = false}) async {
  final result = await super.transaction(action, requireNew: requireNew);
  await customStatement('SELECT 1');
  return result;
}
```

**Possible fix**

Flush after the commit in `_WasmDelegate` once the transaction has been released, e.g. have the transaction delegate flush when the depth-0 transaction completes, or check `database.autocommit` instead of `isInTransaction` after running a statement.

**Versions**

- drift 2.34.3 (also checked 2.35.0 / `develop`)
- drift_flutter 0.3.1
- sqlite3 3.5.2, `sqlite3.wasm` + `drift_worker.js` from the drift-2.34.3 release

sawongam:
Fixes #3864

When using IndexedDB storage, `_WasmDelegate` only flushes if `isInTransaction` is false. The problem is that the COMMIT goes through `runCustom` while that flag is still true, and the flag only gets reset in `_release()` after the commit finished. Since the file system is opened with `writeAutomatically: false`, nothing written in a transaction ends up in IndexedDB until some unrelated write happens later. If the page is reloaded before that, the transaction is gone.

I ran into this in our app where the first-run setup creates the admin account in a transaction. On the deployed web build, refreshing right after setup always sent us back to setup because the account table was empty.

This PR checks `database.autocommit` instead, which is already true once the COMMIT (or ROLLBACK) statement has run, and is false for savepoints inside an outer transaction.

While looking at this I noticed `runSelect` never flushes either, so writes with a RETURNING clause (`insertReturning` etc.) had the same problem even outside of a transaction. I added an override for that too. `flush()` returns right away when there's no pending work, so plain selects shouldn't be affected.

I added a web integration test that inserts in a transaction and with `insertReturning`, reloads the page without any delay and checks both rows are still there. I ran the Firefox / dart2js tests locally, the new test fails on `develop` and passes with this change. I haven't been able to run the Chrome ones locally.


drift/lib/wasm.dart
```diff
@@ -359,12 +359,27 @@ class _WasmDelegate extends Sqlite3Delegate<CommonDatabase> {
...
+  }
 
-    if (!isInTransaction) {
-      await _flush();
-    }
+  @override
+  Future<QueryResult> runSelect(String statement, List<Object?> args) async {
+    final result = await super.runSelect(statement, args);
+    // Selects can write too (e.g. with a RETURNING clause). Flushing without
+    // pending writes is cheap.
```
simolus3, line 380: Good call. I have thought about checking `CommonPreparedStatement.isReadOnly` for this, but since this doesn't do anything apart from checking the work list in the VFS, it should be fine.

simolus3 (approved):
Thanks for the contribution!

PR #3866, opened 2026-09-20
Release transactions held by web clients that are gone

mattcosta5651:
## The problem

On the web all tabs share one drift server in a worker, and a client transaction is a series of round trips. The server keeps the open transaction at the head of `_executorBacklog`, so every other client waits for its turn.

A tab closed mid-transaction never sends its commit, and drift cannot currently notice. `MessagePort` has no close event for a context going away — `WebPortToChannel.channel` says as much in its doc comment — and `ServerImplementation.serve` only removes a closed channel from `_activeChannels`; the executors that channel opened are never released. The abandoned transaction stays at the head of the backlog and every later transaction in every surviving tab waits behind it until all clients are gone.

I found this in a Flutter web app whose end-to-end test opens a second tab and closes it. That tab was closed 17ms into a cache transaction, and the first tab then sat in a `wanderagent_cache` transaction for 422 seconds that had not run a single statement, with nothing else open in its own connection. A two-tab reproduction in Chromium hangs on 2.34.2/2.35.0 and completes with this change.

## The change

**Liveness.** Clients take a uniquely named Web Lock and hold it for their whole lifetime, sending the name in `ServeDriftDatabase.clientLock` behind a new protocol version `v5`. The worker requests the same lock; the browser grants it only once the client's context is destroyed, which is the signal that the client is gone, and the worker then ends that client's channel. This builds on the `locks` getter and the `AcquireLock` extension already in `wasm_setup/shared.dart`. Where Web Locks are unavailable, or either side predates v5, behaviour is unchanged.

**Release.** `ServerImplementation` records which channel opened each transaction or exclusive executor. When a channel closes, it rolls back and releases what that channel still held — newest first, so a nested executor ends before its parent. An executor granted to a channel that closed while waiting for its turn is abandoned the same way, and `_loadExecutor` now throws for a released executor rather than waiting for a turn that can never come.

## Tests

Two tests in `remote_test.dart`, in the style of the existing `nested transactions` case, using a client whose channel stops delivering messages without a rollback or close — what a closed tab does. Both time out against current `develop` and pass with this change:

- a closed client's open transaction is rolled back, and another client can then write;
- a transaction granted to a client that closed while queued behind another transaction is abandoned too.

`dart analyze` is clean and the rest of `remote_test.dart` still passes. I have not been able to run the browser suites on this machine, so the worker side has been exercised through an app build rather than drift's own web tests — worth a look from someone who can run those.

Happy to adjust anything here, including whether this deserves its own protocol version or should ride along without one.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


drift/lib/src/web/wasm_setup/shared.dart
```diff
@@ -544,3 +546,40 @@ extension AcquireLock on LockManager {
...
+
+/// Takes a lock that is held until this JavaScript context is gone and returns
+/// its name, or null where Web Locks are unavailable.
+Future<String?> holdClientLock() async {
+  final manager = locks;
+  if (manager == null) return null;
+
+  final name =
+      'drift-client-${DateTime.now().microsecondsSinceEpoch}-'
+      '${Random().nextInt(0x7fffffff)}';
```
simolus3: Can we use `crypto.randomUUID()` instead of the current time + random?

E.g.

```dart
@JS('crypto')
external Crypto get _crypto;

// ...

final name = 'drift-client-${_crypto.randomUUID()}';
```

drift/lib/src/web/wasm_setup.dart
```diff
@@ -259,6 +259,9 @@ final class _ProbeResult implements WasmProbeResult {
           enableMigrations: enableMigrations,
           protocolVersion: sharedWorker!.version,
           newSerialization: sharedWorker.version >= ProtocolVersion.v3,
+          clientLock: sharedWorker.version >= ProtocolVersion.v5
```
simolus3: Instead of introducing a new protocol version, we can probably include the lock name unconditionally (and old workers would just ignore it). The field is nullable either way, so both old/new combinations continue to work.

drift/lib/src/remote/server_impl.dart
```diff
@@ -23,6 +23,13 @@ class ServerImplementation implements DriftServer {
   final bool closeExecutorWhenShutdown;
 
   final Map<int, QueryExecutor> _managedExecutors = {};
+
+  /// The channel that opened each transaction or exclusive executor.
+  ///
+  /// A client can disappear without closing its transactions - a browser
+  /// tab being closed sends no rollback and no close message. Knowing who
+  /// opened an executor lets us end it when that client is gone.
+  final Map<int, DriftCommunication> _executorOwners = {};
```
simolus3: Instead of using a separate map for this, can we store a `(QueryExecutor, DriftCommunication)` pair in `_managedExecutors` (or perhaps introduce a private class for this)?

drift/lib/src/remote/server_impl.dart
```diff
@@ -197,23 +207,31 @@ class ServerImplementation implements DriftServer {
...
   Future<QueryExecutor> _loadExecutor(int? transactionId) async {
     await _waitForTurn(transactionId);
-    return transactionId != null
-        ? _managedExecutors[transactionId]!
-        : connection;
+    if (transactionId == null) return connection;
+
+    final executor = _managedExecutors[transactionId];
+    if (executor == null) {
+      throw StateError(
```
simolus3: This is effectively unreachable, right? A closed client wouldn't be able to send transaction messages, so unless two clients collude to exchange the internal transaction id this seems unreachable. So the lower-quality error message of `!` seems acceptable to me.

drift/lib/src/remote/server_impl.dart
```diff
@@ -284,8 +313,48 @@ class ServerImplementation implements DriftServer {
...
+  /// A client that disappears mid-transaction, like a browser tab that is
+  /// closed, otherwise leaves its executor at the head of the backlog. Every
+  /// other client of this server then waits behind it indefinitely.
+  Future<void> _abandonExecutorsOf(DriftCommunication comm) async {
+    final owned = [
+      for (final MapEntry(:key, :value) in _executorOwners.entries)
+        if (value == comm) key,
+    ];
+    // A nested executor is created after its parent and has to end first.
+    owned.sort((a, b) => b.compareTo(a));
```
simolus3: Couldn't we also await `_abandonExecutor` for all unordered executors concurrently? The `_waitForTurn` ensures we can't close an executor that has an active child.

simolus3 (commented):
I agree with the fix of using navigator locks to detect this. I'm happy to run tests with a real worker (we also do that in CI), but I have a few minor comments first.

drift/lib/src/web/wasm_setup/shared.dart
mattcosta5651: Done — the name is now `drift-client-${_crypto.randomUUID()}` via a top-level `@JS('crypto')` getter, and `dart:math` is no longer imported here.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

drift/lib/src/web/wasm_setup.dart
mattcosta5651: Done — v5 is gone and `clientLock` is sent unconditionally; the field stays nullable, so an older worker ignores it and a newer worker just finds it absent. `ProtocolVersion.current` is back to v4.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

drift/lib/src/remote/server_impl.dart
mattcosta5651: Done — `_managedExecutors` now holds a private `_ManagedExecutor(executor, owner)` and the side map is gone. `owner` is null for executors the server puts up itself, such as the one a `beforeOpen` callback runs on.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

drift/lib/src/remote/server_impl.dart
mattcosta5651: Agreed, reverted to `!` — and with it the `_waitForTurn` change that existed only to reach that error.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

drift/lib/src/remote/server_impl.dart
mattcosta5651: Done — they're abandoned concurrently with `Future.wait`; each still waits for its own turn, so a parent only proceeds once its children have ended.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

mattcosta5651:
All five points are addressed in the latest push, thanks — `crypto.randomUUID()`, no protocol bump, one `_ManagedExecutor` map, the plain `!` back in `_loadExecutor`, and concurrent abandons.

One thing I'd like your call on, from the `!` discussion. Reverting that also removed the `_waitForTurn` guard, and those were doing two different jobs. With the plain predicate, a statement that arrives for a transaction which has already committed, rolled back or been abandoned can never see its id at the head of the backlog, so it waits in `_waitForTurn` forever rather than reaching `_loadExecutor` to fail.

You're right that a closed client can't send statements — but an *unawaited* statement from a live client can, which is the case the `ArgumentError` about "don't await all operations made inside a transaction" already warns about. This isn't caused by the patch: it's how the code behaves today, and my earlier version had only incidentally changed it.

I've left the simple form here since you asked for it. If you'd like the fail-fast guard as well, it's two lines and I'm happy to push it:

```dart
return !_managedExecutors.containsKey(transactionId) ||
    (_executorBacklog.isNotEmpty && _executorBacklog.first == transactionId);
```

Also worth flagging for your worker-based run: the branch now drains statements already in flight against an executor before rolling it back. A handler resolves its executor and then yields, so a rollback starting in between could let that statement run outside its transaction — the assertion that would catch it is stripped in release builds. There's a test for it in `remote_test.dart`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

simolus3:
Thanks! I have merged this in 9d2330b31fe0788d9911a7d8c09fdf976b770fd2, also added an integration test.

